Issue #15, opened 2018-01-31
Support Mac OSX


smtakeda:

smtakeda:
So far the build and tests are done on Sierra.

PR #16, opened 2018-01-31
Enabled Mac OSX build

smtakeda:
Enabling mac build


smtakeda:
Build test is on Sierra. May need to do on El Capitan later

smtakeda:
Fixed #15

PR #18, opened 2018-02-06
Interface for put get command

howryu:
First commit. 

Just come up with some classes/interfaces. Arguments passed between classes are not finalized. 

Feel free to add comments. Did not necessarily merge this pr.


cpp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
+
+
+set(SOURCE_FILES
+        EncryptionProvider.cpp
+        FileTransferAgent.cpp
+        SnowflakeS3Client.cpp
+        StorageClientFactory.cpp
```
smtakeda: Include headers as well so that Clion can recognize them.

cpp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+        )
+
+add_library(snowflakeclientcpp STATIC ${SOURCE_FILES})
```
smtakeda: Is it possible to merge both C and C++ into one library? e.g., libsnowflakeclient.a

cpp/CMakeLists.txt
```diff
@@ -0,0 +1,13 @@
...
+
+target_link_libraries(snowflakeclient)
```
smtakeda: This looks like nop. The first argument is the target library followed by other dependencies.

cpp/EncryptionMaterial.hpp
```diff
@@ -0,0 +1,49 @@
+//
+// Created by hyu on 2/5/18.
```
smtakeda: Can you please remove your name from the code?
We don't need to keep somebody's name in the code. Instead, those change history should be in the git history.

smtakeda: Also add copyright here.

cpp/EncryptionMaterial.hpp
```diff
@@ -0,0 +1,49 @@
...
+//
+
+#ifndef SNOWFLAKECLIENT_REMOTESTOREFILEENCRYPTIONMATERIAL_HPP
+#define SNOWFLAKECLIENT_REMOTESTOREFILEENCRYPTIONMATERIAL_HPP
+
+#include "string"
+
+namespace Snowflake
```
smtakeda: please use snowflakeclient as the namespace.

smtakeda: Ah, nevermind you have sub name `Client` below.

cpp/EncryptionProvider.cpp
```diff
@@ -0,0 +1,15 @@
+//
+// Created by hyu on 2/5/18.
```
smtakeda: Please replace your name with the copyright.

cpp/EncryptionProvider.hpp
```diff
@@ -0,0 +1,27 @@
+//
+// Created by hyu on 2/5/18.
```
smtakeda: Please replace your name with the copyright.

cpp/EncryptionProvider.cpp
```diff
@@ -0,0 +1,15 @@
...
+//
+
+#include "EncryptionProvider.hpp"
+
+char* Snowflake::Client::EncryptionProvider::encrypt()
+{
+  return "";
```
smtakeda: This is causing compilation error.

Actually it is warning but we want to fix all warnings for safetly.

cpp/FileTransferAgent.hpp
```diff
@@ -0,0 +1,42 @@
...
+      /**
+       * Called by external component to execute put/get command
+       * @param command put/get command
+       * @return a fixed view result set representing upload/download result
+       */
+      FileTransferResult* execute(std::string * command);
+
+    private:
+
+      void upload();
```
smtakeda: I believe these upload and download would need to be public as ODBC wants to call?

cpp/StageInfo.hpp
```diff
@@ -0,0 +1,39 @@
...
+#define SNOWFLAKECLIENT_STAGEINFO_HPP
+
+namespace Snowflake
+{
+  namespace Client
+  {
+    enum StageType
+    {
+      S3,
+      AZURE
```
smtakeda: do we want to support local for tests?

smtakeda:
Overall it is good start. You may want to start write examples or tests to show the usage of new classes.

cpp/CMakeLists.txt
howryu: Sure, I will make it one library.

cpp/FileTransferAgent.hpp
howryu: Not necessarily. The only API external component should call is FileTransferAgent.execute(std::string * sql)

howryu:
Ok here is a summary of how I want to define the protocol for you guys to review. @notkriswagner
 
The api that external component should call is FileTransferAgent.execute(std::string * sql), which should return a FileTransferResult class. External component should add result set wrapper on top this class to include all information needed to be returned to user. 

To construct FileTransferAgent, external component needs to pass in a class which implements IStatement.hpp, which is used to submit query to server to get (put/get command metadata like encryption material, stage information).

smtakeda:
@howryu, can you create an example that actually calls the API so that we can understand better from the code. Also add the example to example directory to make it runnable as well.

howryu:
Ah will refresh my branch, but examples/simple_put_get.cpp should give you an example on how to use the API

CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ add_definitions(-DLOG_USE_COLOR)
 set(CMAKE_VERBOSE_MAKEFILE ON)
 set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -pthread -fPIC -Werror")
 
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} --std=c++14 -fpermissive -Wall -Wextra -lpthread")
```
smtakeda: Be aware that C++14 may not fully supported by Visual Studio 2015: 
https://msdn.microsoft.com/en-us/library/hh409293.aspx

Make sure the code can be compiled on Windows.

CMakeLists.txt
```diff
@@ -10,6 +10,9 @@ add_definitions(-DLOG_USE_COLOR)
...
+set(CMAKE_CXX_COMPILER g++52)
```
smtakeda: Can you move this option to the cmake command line? Is g++52 a requirement? If not, adjust it to AppVeyor environment as well.

examples/simple_put_get.cpp
```diff
@@ -0,0 +1,48 @@
...
+    Snowflake::Client::StatementPutGet(sfstmt);
+
+  Snowflake::Client::FileTransferAgent agent(stmtPutGet);
+
+  Snowflake::Client::FileTransferResult * result = agent.execute(&sql);
+
+  cleanup:
+  /* delete statement */
+  snowflake_stmt_term(sfstmt);
+  delete stmtPutGet;
```
smtakeda: `stmtPutGet` should be deleted before `snowflake_stmt_term(sfstmt)`.

examples/simple_put_get.cpp
```diff
@@ -0,0 +1,48 @@
...
+  SF_STMT *sfstmt = snowflake_stmt(sf);
+
+  std::string sql = "put file:///home/hyu/test1.csv @~";
+
+  Snowflake::Client::IStatementPutGet *stmtPutGet = new
+    Snowflake::Client::StatementPutGet(sfstmt);
+
+  Snowflake::Client::FileTransferAgent agent(stmtPutGet);
+
+  Snowflake::Client::FileTransferResult * result = agent.execute(&sql);
```
smtakeda: How do you handle errors?

examples/simple_put_get.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+  cleanup:
```
smtakeda: The flow like `goto cleanup` structure is for C. `try`, `catch` is better for C++.

include/snowflake/client.h
```diff
@@ -287,6 +298,44 @@ typedef struct sf_snowflake_column_desc {
...
 } SF_COLUMN_DESC;
 
+/**
+ * Encryption material
+ */
+typedef struct sf_encryption_material {
+    char *query_stage_master_key;
+    char query_id[SF_UUID4_LEN];
+    int64 smk_id;
+} SF_ENC_MAT;
```
smtakeda: Btw, do you need to expose SF_ENC_MAT, SF_STAGE_CRED, SF_STAGE_INFO to users? If not, please move them to client_int.h or other headers inside `lib`. The basic idea here is `client.h` should include only public constants and functions.

include/snowflake/client.h
```diff
@@ -570,10 +620,11 @@ snowflake_stmt_get_attr(SF_STMT *sfstmt, SF_STMT_ATTRIBUTE type, void *value);
 /**
  * Executes a statement.
  * @param sfstmt SNOWFLAKE_STMT context.
+ * @param is_put_get_command true this is a put/get command
  *
  * @return 0 if success, otherwise an errno is returned.
  */
-SF_STATUS STDCALL snowflake_execute(SF_STMT *sfstmt);
+SF_STATUS STDCALL snowflake_execute(SF_STMT *sfstmt, sf_bool is_put_get_command);
```
smtakeda: This is basically not great addition, because the application has to specify the query type.

Can you add a new snowflake_execute_ex(SF_STMT *sfstmt, sf_bool is_put_get_command)? Then `snowflake_execute` should detect the query type.

cpp/PutGetParseResponse.cpp
```diff
@@ -0,0 +1,42 @@
+//
+// Created by hyu on 2/7/18.
```
smtakeda: Maybe you may wan to set Clion's template so that it uses the copyright header.

CMakeLists.txt
howryu: I will remove it.

include/snowflake/client.h
```diff
@@ -18,6 +18,7 @@ extern "C" {
 #include <pthread.h>
 #include "snowflake/basic_types.h"
 #include "version.h"
+#include "../../lib/arraylist.h"
```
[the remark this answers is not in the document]
howryu: Ok

cpp/StageInfo.hpp
```diff
@@ -5,9 +5,9 @@
 #ifndef SNOWFLAKECLIENT_STAGEINFO_HPP
 #define SNOWFLAKECLIENT_STAGEINFO_HPP
 
-#include "string"
+#include <string>
 #include <unordered_map>
-#include <snowflake/client.h>
+#include <client_int.h>
```
smtakeda: change this to "client_int.h" instead of <client_int.h> to refer the local file specifically. I believe this will cause an error on Windows.

cpp/FileTransferAgent.cpp
```diff
@@ -3,7 +3,7 @@
 //
 
 #include "FileTransferAgent.hpp"
-#include "iostream";
+#include "iostream"
```
smtakeda:
```
#include <iostream>
```

examples/selectvar.c
```diff
@@ -7,6 +7,7 @@
 #include <memory.h>
 #include <snowflake/client.h>
 #include <example_setup.h>
+#include "string.h"
```
smtakeda: Is this external library? then use
```
#include <string>
```

lib/client.c
```diff
@@ -17,6 +17,7 @@
 #include "results.h"
 #include "error.h"
 #include "chunk_downloader.h"
+#include "regex.h"
```
smtakeda: Is this part of standard library? then 
```
#include <regex.h>
```

smtakeda: The build script didn't catch the error. Fixing build script now.
https://ci.appveyor.com/project/smtakeda/libsnowflakeclient/build/70
```
 C:\projects\libsnowflakeclient\lib\client.c(20): fatal error C1083: Cannot open include file: 'regex.h': No such file or directory [C:\projects\libsnowflakeclient\cmake-build\snowflakeclient.vcxproj]
```

howryu:
@smtakeda Seems appveyor failed to decrypt file if this is a pr. continuous-integration/appveyor/pr

examples/selectvar.c
howryu: This is a C header for strcpy. I don't understand why master branch did not have problem. I did add anything into this test.

Any, it should be 
```
#include <string.h>
```

smtakeda:
Yes, that's known limitation. Your .NET tests are failing. See the bottom of this page.
https://www.appveyor.com/docs/how-to/secure-files/

smtakeda:
@howryu just reminder. https://ci.appveyor.com/project/smtakeda/libsnowflakeclient/build/74 is failing though the status is green. The build script should have caught it.
I've fixed in win32 branch, which will be merged after all tests pass.

howryu:
I will close this one for now until everything is done.

PR #19, opened 2018-02-06
Added win64

smtakeda:
- Adding win64 support by adding platform dependent layer.
- Changed the struct SF_ERROR to SF_ERROR_STRUCT due to conflict with an enum on win64.
- WIP


smtakeda:
Just lerarned PR build on Appveyor won't pick up the credential, so all tests will fail.

smtakeda:
Abandoning  this PR.

PR #24, opened 2018-02-13
Migrated all examples to cmocka tests

smtakeda:
- Added cmocka to deps and deps-build
- Moved all examples that intended test to tests using cmocka
- Enabled ctest so that you can run tests by ctest instead of running individual executables.
- Fixed and removed timestamp conversion issue on OSX and Windows. Currently we don't support prior to year 1970 on Windows.


smtakeda:
Note example directory still exists, but if you want to add test cases, all should go to tests and not example.

PR #26, opened 2018-02-14
Aws build

notkriswagner:
Note: I had to use cmake 3.0+ and GCC/G++ 4.9+ to get the AWS library to build properly. I also had to update tools on travis to get a more updated version of GCC/G++


tests/CMakeLists.txt
```diff
@@ -75,3 +79,23 @@ FOREACH (T ${TESTS})
             --suppressions=${VALGRIND_SUPPRESSION}
             ./${T})
 ENDFOREACH ()
+
+if (LINUX)
```
smtakeda: Do you have a mac? Can you try build and run tests on Mac after this change?

tests/test_include_aws.cpp
```diff
@@ -4,11 +4,20 @@
```
smtakeda: Great to have a test case from the beginning!

scripts/build_curl_openssl.sh
```diff
@@ -91,3 +92,31 @@ PKG_CONFIG_PATH="$OPENSSL_BUILD_DIR/lib/pkgconfig" LIBS="-ldl" ./configure ${cur
 echo "Building and Installing Curl"
 make 2>&1 > /dev/null
 make install 2>&1 > /dev/null
+
+# build aws-cpp-sdk
```
smtakeda: As we talked in the meeting, can you create a separate script for AWS SDK build so that the build_curl_openssl.sh calls? It would be great if you change build_curl_openssl.sh to build_dependencies.sh as well.

smtakeda:
LGTM other than the minor changes.

smtakeda:
merged the change for now as the tests passed.
Please address the changes in the comments.

smtakeda:
FYI, I'll split build scripts into openssl, curl and aws.

PR #27, opened 2018-02-17
Fixed depenency component builds

smtakeda:
- Filled missing files for OpenSSL build
- Split build_curl_open.ssh into build_openssl.sh, build_curl.sh and build_awssdk.sh
- Added build_dependencies.sh that call all sub build scripts.
- Fixed AWSSDK build using locally generated OpenSSL and Curl static libraries


scripts/build_libsnowflakeclient.sh
```diff
@@ -21,9 +21,15 @@ done
...
+source $DIR/_init.sh
+
 cd $DIR/..
 rm -rf cmake-build
 mkdir cmake-build
 cd cmake-build
-cmake ..
+cmake_opts=(
+    "-DCMAKE_C_COMPILER=$GCC"
+    "-DCMAKE_CXX_COMPILER=$GXX"
```
smtakeda: Using gcc4.9+

scripts/build_awssdk.sh
```diff
@@ -0,0 +1,47 @@
...
+    "-DCMAKE_CXX_COMPILER=$GXX"
+    "-DBUILD_ONLY=s3"
+    "-DCMAKE_INSTALL_PREFIX=$AWS_BUILD_DIR"
+    "-DBUILD_SHARED_LIBS=OFF"
+    "-DCMAKE_PREFIX_PATH=\"$LIBCURL_BUILD_DIR/;$OPENSSL_BUILD_DIR/\""
+    "-DENABLE_TESTING=OFF"
+    "-DOPENSSL_ROOT_DIR=$DEPENDENCY_DIR/openssl"
+    "-DOPENSSL_USE_STATIC_LIBS=true"
+    "-DCURL_INCLUDE_DIR=$DEPENDENCY_DIR/curl/include"
+    "-DCURL_LIBRARY=$DEPENDENCY_DIR/curl/lib"
```
smtakeda: @notkriswagner those additional parameters are to build AWSSDK with the locally built OpenSSL and Curl. CMAKE_PREFIX_PATH seems not working.

smtakeda: Without this change, AWSSDK attempts to use the system's curl and openssl, and if openssl is 1.0.x, the link will fail because the function name was changed from 1.1.x.

notkriswagner:
Okay sounds good to me. The CMAKE_PREFIX_PATH option is very finicky and I'm glad to get rid of it. The documentation says that for Linux, multiple paths should be separated by a ":", but by experimentation, I found that is not true and that when you specify multiple paths for CMAKE_PREFIX_PATH, you must surround the paths in quotes and use a ";". At least thats what I was using on the command line. Either way I'm happy to get rid of the CMAKE_PREFIX_PATH option

smtakeda:
Actually I have not removed it yet in main. Can you remove it in the next commit?

PR #29, opened 2018-02-26
Added basic API for C++ Wrapper

notkriswagner:
The C++ wrapper is under the Snowflake::Client namespace while all of the C API structs and functions are under the Snowflake::CAPI namespace. This will allow a user to use either the C API or C++ wrapper in their code. Only a connection and statement class are defined. Structs that only hold data (input/output bindings, error struct) are not defined as classes since there are no internal operations for those structs.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -36,11 +36,12 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 if (UNIX)
     # Linux and OSX
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -fPIC -Werror")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x -std=gnu++0x")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -g -fPIC -Werror -std=c++0x -std=gnu++0x")
 endif ()
 if (LINUX)
     # Linux. MacOS doesn't require pthread option
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -pthread")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -pthread")
 endif ()
 
 if (LINUX)
@@ -109,7 +110,8 @@ include_directories(
         deps-build/${PLATFORM}/openssl/include
         deps-build/${PLATFORM}/zlib/include
         deps-build/${PLATFORM}/aws/include
-        include)
+        include
+        cpp/include)
 
 message("libcurl is located at " ${CURL_LIB})
 message("libssl is located at " ${SSL_LIB})
@@ -142,5 +144,6 @@ if (WIN32)
     target_link_libraries(snowflakeclient)
 endif ()
 
+add_subdirectory(cpp)
 add_subdirectory(examples)
 add_subdirectory(tests)
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -1,0 +1,19 @@
+# Copyright (c) 2017 Snowflake Computing, Inc. All rights reserved.
+#
+# CMakeList for Snowflake Client++
+#
+
+set(SOURCE_FILES_CXX
+        include/snowflake++/SnowflakeInclude.hpp
+        include/snowflake++/SnowflakeConnection.hpp
+        include/snowflake++/SnowflakeStatement.hpp
+        lib/SnowflakeConnection.cpp
+        lib/SnowflakeStatement.cpp)
+
+set(TESTLIB_OPTS_CXX snowflakeclient -Wl,--whole-archive curl ssl crypto pthread aws-cpp-sdk-core aws-cpp-sdk-s3 -Wl,--no-whole-archive)
+
+add_library(snowflakeclient++ STATIC ${SOURCE_FILES_CXX})
+
+set_target_properties(snowflakeclient++ PROPERTIES LINKER_LANGUAGE CXX)
+
+target_link_libraries(snowflakeclient++ ${TESTLIB_OPTS_CXX})
--- a/cpp/include/snowflake++/SnowflakeConnection.hpp
+++ b/cpp/include/snowflake++/SnowflakeConnection.hpp
@@ -1,0 +1,51 @@
+/*
+ * Copyright (c) 2017-2018 Snowflake Computing, Inc. All rights reserved.
+ */
+
+#ifndef SNOWFLAKECLIENT_SNOWFLAKECONNECTION_HPP
+#define SNOWFLAKECLIENT_SNOWFLAKECONNECTION_HPP
+
+#include "SnowflakeInclude.hpp"
+
+namespace Snowflake {
+    namespace Client {
+        class SnowflakeConnection {
+        public:
+
+            /* Construct a blank Snowflake Connection */
+            SnowflakeConnection(void);
+
+            /*
+             * Construct with a connection pointer. Copies connection info from
+             * passed in connection
+             */
+            SnowflakeConnection(Snowflake::CAPI::SF_CONNECT &connection_);
+
+            ~SnowflakeConnection(void);
+
+            Snowflake::CAPI::SF_STATUS connect();
+
+            Snowflake::CAPI::SF_STATUS setAttribute(Snowflake::CAPI::SF_ATTRIBUTE type_,
+                                                    const void *value_);
+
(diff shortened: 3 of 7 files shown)


cpp/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
+# Copyright (c) 2017 Snowflake Computing, Inc. All rights reserved.
+#
+# CMakeList for Snowflake Client++
+#
+
+set(SOURCE_FILES_CXX
+        include/snowflake++/SnowflakeInclude.hpp
```
smtakeda: I believe it is better to avoid `+` in a path name. Do you have any reference?
can it be include/snowflakecpp?  or even shared with C's header file directory?

cpp/include/snowflake++/SnowflakeConnection.hpp
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#ifndef SNOWFLAKECLIENT_SNOWFLAKECONNECTION_HPP
+#define SNOWFLAKECLIENT_SNOWFLAKECONNECTION_HPP
+
+#include "SnowflakeInclude.hpp"
+
+namespace Snowflake {
+    namespace Client {
+        class SnowflakeConnection {
```
smtakeda: Maybe you just need `class Connection`

cpp/include/snowflake++/SnowflakeStatement.hpp
```diff
@@ -0,0 +1,75 @@
...
+#ifndef SNOWFLAKECLIENT_SNOWFLAKESTATEMENT_HPP
+#define SNOWFLAKECLIENT_SNOWFLAKESTATEMENT_HPP
+
+#include <SnowflakeInclude.hpp>
+#include <string>
+#include "SnowflakeConnection.hpp"
+
+namespace Snowflake {
+    namespace Client {
+        class SnowflakeStatement {
```
smtakeda: Maybe you just need `class Statement`

cpp/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+        include/snowflake++/SnowflakeConnection.hpp
+        include/snowflake++/SnowflakeStatement.hpp
```
smtakeda: Also the header file name should be compact, e.g., Util.hpp, Connection.hpp and Statement.hpp as the directory name already have `snowflake` name space.

smtakeda:
Can you start adding examples or test cases? It will clarify the usage of API.

cpp/CMakeLists.txt
notkriswagner: I like include/snowflakecpp idea

notkriswagner:
Sure, I'll add a test case in the next commit

cpp/CMakeLists.txt
```diff
@@ -0,0 +1,19 @@
...
+        lib/SnowflakeConnection.cpp
+        lib/SnowflakeStatement.cpp)
+
+set(TESTLIB_OPTS_CXX snowflakeclient -Wl,--whole-archive curl ssl crypto pthread aws-cpp-sdk-core aws-cpp-sdk-s3 -Wl,--no-whole-archive)
+
+add_library(snowflakeclient++ STATIC ${SOURCE_FILES_CXX})
```
notkriswagner: @smtakeda should the library name also change to snowflakeclientcpp?

cpp/CMakeLists.txt
smtakeda: Can we put both hpp and h together in a single directory? I'm not 100% sure both C and C++ can share the same library, though, in theory, linker is supposed to pick and choose corresponding object files.

cpp/CMakeLists.txt
smtakeda: Ideally one library includes both c and C++ API. I believe a C application could link properly as the linker picks right objects from *.a file.

smtakeda:
@notkriswagner please fix tests.

smtakeda:
@notkriswagner I still see the compilation errors. can you run the build locally and ensure no error?

notkriswagner:
@smtakeda I did run it locally and everything worked. I'll double check though and investigate further

smtakeda:
```
4>C:\projects\libsnowflakeclient\include\snowflake/client.h(68): error C2371: 'SF_TYPE': redefinition; different basic types
```

notkriswagner:
I ended up changing the enum name to 'SF_DB_TYPE' from 'SF_DB_TYPE'. @smtakeda This is the solution that I came up with, but I'm open to other ideas. Its not ideal to change a type name in the C API, but I'm not sure what else we can do since it conflicts with a type name for the Windows SDK.

smtakeda:
@notkriswagner I am not aware of `SF_TYPE` issue, though, I came across a similar conflict with other function name but resolved by changing the name. Let's keep your change for now. All enum may be changed later as well.

Btw, the last step for PR is resolve a conflict with the main branch. Can you rebase and push?

CMakeLists.txt
```diff
@@ -152,7 +153,8 @@ include_directories(
         deps-build/${PLATFORM}/zlib/include
         deps-build/${PLATFORM}/aws/include
         include
-        lib)
+        lib
+        cpp/include)
```
smtakeda: Is this an internal header directory?

CMakeLists.txt
```diff
@@ -189,5 +191,6 @@ if (WIN32)
     target_link_libraries(snowflakeclient)
 endif ()
 
+add_subdirectory(cpp)
```
smtakeda: I think this was already in main. Please rebase.

tests/CMakeLists.txt
```diff
@@ -63,7 +68,7 @@ message("valgrind suppression file is located at " ${VALGRIND_SUPPRESSION})
 
 if (LINUX)
     set(TESTLIB_OPTS_C snowflakeclient ${CMOCKA_LIB} -Wl,--whole-archive curl ssl crypto -Wl,--no-whole-archive pthread)
-    set(TESTLIB_OPTS_CXX snowflakeclient ${CMOCKA_LIB} -Wl,--whole-archive curl ssl crypto pthread aws-cpp-sdk-core aws-cpp-sdk-s3 -Wl,--no-whole-archive)
+    set(TESTLIB_OPTS_CXX snowflakeclientcpp snowflakeclient ${CMOCKA_LIB} -Wl,--whole-archive curl ssl crypto pthread aws-cpp-sdk-core aws-cpp-sdk-s3 -Wl,--no-whole-archive)
```
smtakeda: Basically we want to put both CPP and C code together into a single library file.
The header file locations could be separated like `snowflakecpp`, but we will revisit it later.

tests/CMakeLists.txt
```diff
@@ -92,20 +97,20 @@ ENDFOREACH ()
...
         # Removed to suppress false alarm: "--run-libc-freeres=no"
-        add_test(valgrind_${T}
+        add_test(valgrind_${T}_cpp
                 valgrind
                 --tool=memcheck
                 --leak-check=full
                 --error-exitcode=1
                 --suppressions=${VALGRIND_SUPPRESSION}
-                ./${T})
+                ./${T}_cpp)
```
smtakeda: please follow @howryu 's update for the tests.

tests/test_select1.cpp
```diff
@@ -0,0 +1,51 @@
...
+        int64 out = 0;
+        SF_BIND_OUTPUT c1 = {
+          idx : 1,
+          c_type : SF_C_TYPE_INT64,
+          max_length : 0,
+          value : (void *) &out,
+          len : sizeof(out),
+          is_null : 0
+        };
+        stmt->bindResult(c1);
```
smtakeda: We may want to have a better way for binding. The idea is no C type or even the buffer size is required, but can defer by the time of fetch.

Anyway let's postpone the design to the later stage.

smtakeda: I don't know others but think about JDBC interface. It is not necessarily that JDBC is the best, though. More or less object oriented based interface is preferred to leverage C++ feature.

CMakeLists.txt
notkriswagner: This is a header directory that the user needs to include when using the C++ library.

CMakeLists.txt
notkriswagner: I think so as well. I wasn't able to rebase my branch with master, I had to merge it in. When this gets merged into the master branch, it can be done as part of a squash commit.

CMakeLists.txt
smtakeda: If external headers, please move them to `include`.

smtakeda:
@notkriswagner is this still pending? I thought you've already merged some C++ interface?

PR #35, opened 2018-04-03
Fix logging

smtakeda:
- Added CAPI version INFO to the log. This is consistent with the recent ODBC change.
- Fixed memory leak in sf_array_list when binding results. If the result buffer is not bound and the returned column is more than 8, fetch attempts to read uninitialized buffer.
- Added a function converting a debug level string to enum.

### DIFF
--- a/lib/arraylist.c
+++ b/lib/arraylist.c
@@ -38,8 +38,8 @@ void STDCALL sf_array_list_grow(ARRAY_LIST *al, size_t min_size) {
 }
 
 void STDCALL sf_array_list_set(ARRAY_LIST *al, void *item, size_t index) {
-    if (al->size < index) {
-        sf_array_list_grow(al, index);
+    if (al->size <= index) {
+        sf_array_list_grow(al, index+1);
     }
     // If element we are writing to is NULL and item is not NULL, we want to increment 'used'.
     // Otherwise we are writing to a spot that already contains an element
@@ -54,5 +54,8 @@ void STDCALL sf_array_list_set(ARRAY_LIST *al, void *item, size_t index) {
 }
 
 void* STDCALL sf_array_list_get(ARRAY_LIST *al, size_t index) {
+    if (al->size <= index) {
+        sf_array_list_grow(al, index+1);
+    }
     return al->data[index];
 }
--- a/lib/client.c
+++ b/lib/client.c
@@ -627,6 +627,13 @@ SF_STATUS STDCALL snowflake_connect(SF_CONNECT *sf) {
     // Reset error context
     clear_snowflake_error(&sf->error);
 
+    char os_version[128];
+    sf_os_version(os_version);
+
+    log_info("Snowflake C/C++ API: %s, OS: %s, OS Version: %s",
+             SF_API_VERSION,
+             sf_os_name(),
+             os_version);
     cJSON *body = NULL;
     cJSON *data = NULL;
     cJSON *resp = NULL;
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -133,35 +133,14 @@ cJSON *STDCALL create_auth_json_body(SF_CONNECT *sf,
     cJSON *data;
     cJSON *client_env;
     cJSON *session_parameters;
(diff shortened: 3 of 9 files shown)


lib/arraylist.c
```diff
@@ -38,8 +38,8 @@ void STDCALL sf_array_list_grow(ARRAY_LIST *al, size_t min_size) {
 }
 
 void STDCALL sf_array_list_set(ARRAY_LIST *al, void *item, size_t index) {
-    if (al->size < index) {
-        sf_array_list_grow(al, index);
+    if (al->size <= index) {
+        sf_array_list_grow(al, index+1);
```
smtakeda: For example, if the array size is 8 and index is 8, it should extend the buffer size.

A confusion here is sf_array_list_grow takes the array size, while sf_array_list_set takes the index.

lib/arraylist.c
```diff
@@ -54,5 +54,8 @@ void STDCALL sf_array_list_set(ARRAY_LIST *al, void *item, size_t index) {
 }
 
 void* STDCALL sf_array_list_get(ARRAY_LIST *al, size_t index) {
+    if (al->size <= index) {
+        sf_array_list_grow(al, index+1);
+    }
```
smtakeda: Same as the previous comment. Without this, uninitialized memory will be read by the following line.

lib/client.c
```diff
@@ -627,6 +627,13 @@ SF_STATUS STDCALL snowflake_connect(SF_CONNECT *sf) {
     // Reset error context
     clear_snowflake_error(&sf->error);
 
+    char os_version[128];
+    sf_os_version(os_version);
+
+    log_info("Snowflake C/C++ API: %s, OS: %s, OS Version: %s",
+             SF_API_VERSION,
+             sf_os_name(),
+             os_version);
```
smtakeda: This will output the CAPI info in the log.

lib/connection.c
```diff
@@ -133,35 +133,14 @@ cJSON *STDCALL create_auth_json_body(SF_CONNECT *sf,
...
-#  elif defined(__APPLE__)
-        cJSON_AddStringToObject(client_env, "OS_VERSION", "MacOS");
-#  else
-        cJSON_AddStringToObject(client_env, "OS_VERSION", "Unknown");
-#  endif
-    }
-#endif
+    cJSON_AddStringToObject(client_env, "OS", sf_os_name());
+    sf_os_version(os_version);
+    cJSON_AddStringToObject(client_env, "OS_VERSION", os_version);
```
smtakeda: Moved OS specific code to platform.c

lib/connection.c
```diff
@@ -715,7 +694,7 @@ ARRAY_LIST *json_get_object_keys(const cJSON *item) {
 size_t
 json_resp_cb(char *data, size_t size, size_t nmemb, RAW_JSON_BUFFER *raw_json) {
     size_t data_size = size * nmemb;
-    log_debug("Curl response size: %zu\n", data_size);
+    log_debug("Curl response size: %zu", data_size);
```
smtakeda: Removed the return code.

Folks, please don't add the return code to the end.

lib/logger.c
```diff
@@ -26,110 +26,123 @@
...
+        va_end(args);
+        fprintf(L.fp, "\n");
+        fflush(L.fp);
+    }
+
+    /* Release lock */
+    unlock();
+}
+
+SF_LOG_LEVEL log_from_str_to_level(const char *level_in_str) {
```
smtakeda: The new function converting a string to enum.

lib/platform.c
```diff
@@ -302,3 +312,76 @@ sf_bool STDCALL _is_put_get_command(char* sql_text)
...
+#if _WIN64
+        "x86_64"
+#else
+        "x86"
+#endif // _WIN64
+    );
+#endif // Unknown Platform
+}
+
+int STDCALL sf_strncasecmp(const char *s1, const char *s2, size_t n) {
```
smtakeda: This is required to support Windows.

Issue #7, opened 2018-01-30
Review logging


smtakeda:
Goals

- Ensure the log includes required info for debug.
- Ensure the log can be enabled and disabled as the customer needs.
- Ensure the log location can be specified.
- Ensure the log doesn't include sensitive information, i.e., password, private key

smtakeda:
Questions:
- How DEBUG global flag is used aside from multiple log level?
- Can we set the log directory or even file for log after snowflake_global_init?

howryu:
Another question to think about is how to provide interface to hook logging facilities to upper layer, i.e. ODBC

bdforbes:
Hi, I'm wondering if this issue is the root cause of something I've encountered in the Snowflake ODBC driver (version 2.20.05.00). I've set up a 64-bit ODBC Data Source connection in Windows, specifying a Tracing level of 0. However, when I use the ODBC connection from R, I still see a huge amount of logging going to the console (I think it's stderr). It's almost entirely INFO level tracing about the connection, which I though should be excluded with Tracing set to 0.

![image](https://user-images.githubusercontent.com/15697539/101130948-887f5400-3658-11eb-95e9-86dd7d1b1a9f.png)

Is this a known issue?

sfc-gh-igarish:
To clean up and re-prioritize more pressing bugs and feature requests we are closing all issues older than 6 months as of March 1, 2023. If there are any issues or feature requests that you would like us to address, please create them according to the new templates we have created. For urgent issues, opening a support case with this link [Snowflake Community](https://community.snowflake.com/s/article/How-To-Submit-a-Support-Case-in-Snowflake-Lodge) is the fastest way to get a response.

PR #36, opened 2018-04-03
fixed NULL input for log level. Added -fPIC to build CPP files

smtakeda:


CMakeLists.txt
```diff
@@ -36,7 +36,7 @@ set(CMAKE_VERBOSE_MAKEFILE ON)
 if (UNIX)
     # Linux and OSX
     set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -g -fPIC -Werror")
-    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x -std=gnu++0x")
+    set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -std=c++0x -std=gnu++0x -fPIC")
```
smtakeda: @howryu @notkriswagner do you have any specific reason why you didn't add -fPIC? Without this, the PHP PDO build will fail.

howryu: Go ahead and add it. No reasons for me.

howryu:
LGTM. Not related to this pr, but I have a general question. 

Should we add ability to capture logs from curl, aws sdk in the same log file as snowflakel logs? If yes, we need to provide some interface. Also maybe addd namespace for each log entry.

smtakeda:
yes, we need curl log as well. Can you file a issue including more specific needs?

smtakeda:
@howryu https://github.com/snowflakedb/libsnowflakeclient/issues/7 is the place to add comments about logging.

PR #39, opened 2018-04-04
Compression handling

howryu:


cpp/FileCompressionType.cpp
```diff
@@ -0,0 +1,199 @@
...
+const FileCompressionType FileCompressionType::NONE =
+  FileCompressionType("", "", true);
+
+bool Snowflake::Client::FileCompressionType::matchMagicNumber(char *header) const
+{
+  if (m_magicNumbers)
+  {
+    for (size_t i = 0; i < m_magicNumbers->size(); i++)
+    {
+      if (!memcmp(header, m_magicNumbers->at(i), m_magicBytes))
```
smtakeda: You may wan to check if data in header is less than m_magicBytes and return false.

tests/test_simple_put.cpp
```diff
@@ -128,17 +137,68 @@ static int teardown(void **unused)
...
 int main(void) {
-  initialize_test(SF_BOOLEAN_FALSE);
   const struct CMUnitTest tests[] = {
-    cmocka_unit_test_setup_teardown(test_simple_put, NULL, teardown),
+    cmocka_unit_test_teardown(test_simple_put_auto_compress, teardown),
+    cmocka_unit_test_teardown(test_simple_put_auto_detect_gzip, teardown),
+    cmocka_unit_test_teardown(test_simple_put_no_compress, teardown),
+    cmocka_unit_test_teardown(test_simple_put_gzip, teardown),
+    cmocka_unit_test_teardown(test_simple_put_auto_detect_bz2, teardown),
+    cmocka_unit_test_teardown(test_simple_put_auto_detect_zst, teardown),
```
smtakeda: please add the 0 and 1 byte length file test.

howryu:
One of the travis check failed because of `sudo apt get update` failed.

smtakeda:
You can restart the job. if not, let me know.

howryu:
@smtakeda It's hard to say whether `header` will contains null in middle or not since this is just first 4 bytes of any file. Not sure if it is possible that a file's first 4 byte will contain null or not. However, I already preallocate enough space (5 bytes including `\0`). And `memcmp` takes an argument indicating how many bytes to compare, in this case is `m_magicBytes`. Idea is that m_magicBytes will never be bigger than MAX_MAGIC_BYTES macro. So I think we are good here.

PR #43, opened 2018-04-08
Concurrent Upload

howryu:
This change includes
- Support specifying wildcard pattern in PUT command source locations.
- Implement a naive pool thread object to do parallel upload
- Refactor the code by moving all metadata related operation into a class called FileMetadataInitiailizer, so that code can be tested separately.


cpp/EncryptionProvider.hpp
```diff
@@ -19,18 +19,6 @@ namespace Client
 class EncryptionProvider
 {
 public:
-  /**
```
howryu: This method is moved to FileMetadataInitializer

tests/test_unit_thread_pool.cpp
```diff
@@ -0,0 +1,29 @@
...
+#include <thread>
+
+void test_thread_pool(void **unused)
+{
+  Snowflake::Client::Util::ThreadPool tp(4);
+
+  for (int i=0; i<10; i++)
+  {
+    tp.AddJob([=]{
+      std::this_thread::sleep_for(std::chrono::seconds(4));
```
smtakeda: Is there any way to verify all of the threads are completed?

tests/test_unit_file_metadata_init.cpp
```diff
@@ -0,0 +1,117 @@
...
+#define FILES_IN_DIR "file1.csv", "file2.csv", "file3.csv", "file4.csv", "file1.gz"
+
+
+typedef ::Snowflake::Client::FileCompressionType FileCompressionType;
+
+void getDataDirectory(std::string& dataDir)
+{
+  const std::string current_file = __FILE__;
+  std::string testsDir = current_file.substr(0, current_file.find_last_of('/'));
+  dataDir = testsDir + "/data/";
```
smtakeda: ok for now to run tests on Linux. But keep in mind that windows will needs double backslashes.

cpp/FileMetadataInitializer.cpp
```diff
@@ -0,0 +1,143 @@
...
+  {
+    // open dir failed
+    throw;
+  }
+}
+
+void Snowflake::Client::FileMetadataInitializer::initCompressionMetadata(
+  FileMetadata *fileMetadata)
+{
+  if(!strncasecmp(m_sourceCompression, "AUTO_DETECT", 11) ||
```
smtakeda: can it be sizeof(AUTO_DETECT)?

cpp/FileMetadataInitializer.cpp
```diff
@@ -0,0 +1,143 @@
...
+    if (!fileMetadata->sourceCompression)
+    {
+      // no compression found
+      throw;
+    }
+  }
+
+  if (fileMetadata->sourceCompression == &FileCompressionType::NONE)
+  {
+    fileMetadata->targetCompression = &FileCompressionType::GZIP;
```
smtakeda: Is this the similar one I recently fixed in the Python Connector? If no auto compress, then the file compression type would not be Gzip? I'm not sure what this gzip mean.

cpp/FileMetadataInitializer.cpp
howryu: Sure will do it.

cpp/FileMetadataInitializer.cpp
howryu: This GZIP is the data returned to the caller i.e. user will know that target compression is gzip. It will not influence the upload result since s3 connector did not need to know the compression type of file. The bug in Python Connector is because the upload metadata contains a content-encoding metadata indicating the source is gzip compression. Any way, will fix it.

tests/test_unit_file_metadata_init.cpp
howryu: Sure.

tests/test_unit_thread_pool.cpp
howryu: The thread pool I implemented cannot return future. (It requires some additional work to implement future)

I can verify by passing in a variable reference declared in main thread and update the value in worker thread. After threads are done, check those values are updated or not.

PR #44, opened 2018-04-17
Multi part upload

howryu:
In this commit, I enabled supporting multi part upload by s3-client library instead of transfer-manager library because transfer manager requires input to be a file (which means that after encryption, additional disk io is required since larges files need to be written into disk before using transfer manager to upload).

With this implementation, it will buffer the chunk of data to 5mb and upload a single parts, which means reading content from original file and encryption the block on the fly, and then do upload parts.


smtakeda:
@howryu can you rebase and push again?

cpp/SnowflakeS3Client.cpp
```diff
@@ -45,6 +53,8 @@ SnowflakeS3Client::SnowflakeS3Client(StageInfo *stageInfo):
   Aws::InitAPI(options);
   clientConfiguration.region = *stageInfo->getRegion();
   clientConfiguration.caFile = Aws::String(caBundleFile);
+  clientConfiguration.requestTimeoutMs = 40000;
+  clientConfiguration.connectTimeoutMs = 30000;
```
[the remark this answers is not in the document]
smtakeda: It would be nice these values are configurable. Maybe environment variables?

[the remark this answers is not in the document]
howryu: Do we need some general ways to config values/parameters? Or do we want to leave every configurable values in environment variables?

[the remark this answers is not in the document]
smtakeda: Yeah, I take back the previous suggestion.

In C API,
snowflake_global_set_attribute is for the parameter in global scope.
snowflake_set_attribute is for the connection scope.  etc. 

However, C++ API needs something similar. Let's file a ticket for now and work on it after @notkriswagner's framework is in.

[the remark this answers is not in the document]
smtakeda: Those constant values are for C++ only, so we don't want to put them in C layer.

cpp/SnowflakeS3Client.cpp
```diff
@@ -98,6 +121,126 @@ TransferOutcome SnowflakeS3Client::upload(FileMetadata *fileMetadata,
...
+  Aws::S3::Model::UploadPartOutcome outcome = s3Client->UploadPart(uploadPartRequest);
+
+  if (outcome.IsSuccess())
+  {
+    uploadCtx->m_outcome = TransferOutcome::SUCCESS;
+    uploadCtx->m_etag = outcome.GetResult().GetETag();
+  }
+  else
+  {
+    fprintf(stderr, "Upload request: %s\n", outcome.GetError().GetMessage().c_str());
```
smtakeda: let's clean these message up later in the logging API in C++.

cpp/SnowflakeS3Client.cpp
```diff
@@ -98,6 +121,126 @@ TransferOutcome SnowflakeS3Client::upload(FileMetadata *fileMetadata,
...
+
+    m_threadPool->WaitAll();
+
+    Aws::S3::Model::CompletedMultipartUpload completedMultipartUpload;
+    for (unsigned int i=0; i< totalParts; i++)
+    {
+      if (uploadParts[i].m_outcome == TransferOutcome::SUCCESS)
+      {
+        Aws::S3::Model::CompletedPart completedPart;
+        completedPart.WithETag(uploadParts[i].m_etag)
```
smtakeda: is E-Tag used for specific purpose? I'm 
I just want to remind that E-Tag cannot include a hash value for the content for the purpose of content match.

cpp/SnowflakeS3Client.cpp
```diff
@@ -98,6 +121,126 @@ TransferOutcome SnowflakeS3Client::upload(FileMetadata *fileMetadata,
...
+      m_threadPool->AddJob([&splitter, i, this, &uploadParts]()->void
+                           {
+                             Util::ByteArrayStreamBuf * buf = splitter.getNextSplitPart();
+                             uploadParts[i].buf = buf;
+                             this->uploadParts(&uploadParts[i]);
+                             splitter.markDone(buf);
+                           });
+    }
+
+    m_threadPool->WaitAll();
```
smtakeda: could this raise any known exception? such as ExpiredToken?

howryu: Expired token is not supported currently. I probably do it in separate pr. This change is already big enough.

cpp/SnowflakeS3Client.cpp
howryu: This is required by multi part upload API. After each part is uploaded successfully, a etag will be returned to client side. And all parts' etag should be recorded and be put in last complete upload request.

cpp/SnowflakeS3Client.cpp
howryu: Yeah will do that.

jianshenghuang:
Nice work. Do we have tests for upload large files with the new multi part upload with and w/o encryption?

howryu:
@jianshenghuang Yes. We have it in `tests/test_parallel_upload.cpp`. All tests are uploaded with encryption. Do I need to worry no encryption case?

jianshenghuang:
No need to worry about no encryption.

PR #46, opened 2018-04-18
Fixed region support by accepting NULL connection attribute values.

smtakeda:
There was a bug in handling empty parameters for connection.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -512,6 +512,26 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
         alloc_buffer_and_copy(&sf->port, "443");
     }
 
+    log_debug("application name: %s", sf->application_name);
+    log_debug("application version: %s", sf->application_version);
+    log_debug("authenticator: %s", sf->authenticator);
+    log_debug("user: %s", sf->user);
+    log_debug("password: %s", sf->password ? "****" : sf->password);
+    log_debug("host: %s", sf->host);
+    log_debug("port: %s", sf->port);
+    log_debug("account: %s", sf->account);
+    log_debug("region: %s", sf->region);
+    log_debug("database: %s", sf->database);
+    log_debug("schema: %s", sf->schema);
+    log_debug("warehouse: %s", sf->warehouse);
+    log_debug("role: %s", sf->role);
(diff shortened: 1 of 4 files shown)


lib/client.c
```diff
@@ -622,7 +642,7 @@ SF_CONNECT *STDCALL snowflake_init() {
         sf->passcode = NULL;
         sf->passcode_in_password = SF_BOOLEAN_FALSE;
         sf->insecure_mode = SF_BOOLEAN_FALSE;
-        sf->autocommit = SF_BOOLEAN_FALSE;
+        sf->autocommit = SF_BOOLEAN_TRUE;
```
smtakeda: autocommit is enabled by default.

lib/client.c
```diff
@@ -920,16 +940,16 @@ SF_STATUS STDCALL snowflake_set_attribute(
             alloc_buffer_and_copy(&sf->authenticator, value);
             break;
         case SF_CON_INSECURE_MODE:
-            sf->insecure_mode = *((sf_bool *) value);
+            sf->insecure_mode = value ? *((sf_bool *) value) : SF_BOOLEAN_FALSE;
```
smtakeda: NULL can be specified for the attribute.

lib/connection.c
```diff
@@ -790,8 +790,7 @@ sf_bool STDCALL http_perform(CURL *curl,
             }
         }
 
-
-        res = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, &json_resp_cb);
+        res = curl_easy_setopt(curl, CURLOPT_WRITEFUNCTION, (void*)&json_resp_cb);
```
smtakeda: Somehow this is required to build Release binary, or get an error.

scripts/build_libsnowflakeclient.sh
```diff
@@ -30,6 +20,7 @@ cd cmake-build
 cmake_opts=(
     "-DCMAKE_C_COMPILER=$GCC"
     "-DCMAKE_CXX_COMPILER=$GXX"
+    "-DCMAKE_BUILD_TYPE=$target"
```
smtakeda: Now the script can build a debug binary.

PR #52, opened 2018-04-30
C++ Wrapper API

notkriswagner:
Added basic C++ Wrapper API. An example of how the API works is shown in "test_cpp_select1.cpp". Updated SF_TYPE to SF_DB_TYPE to solve naming conflict with Windows.

### DIFF
--- a/cpp/lib/Exceptions.cpp
+++ b/cpp/lib/Exceptions.cpp
@@ -1,0 +1,4 @@
+//
+// Created by vagrant on 2/27/18.
+//
+
--- a/cpp/lib/Exceptions.cpp
+++ b/cpp/lib/Exceptions.cpp
@@ -1,4 +1,4 @@
-//
-// Created by vagrant on 2/27/18.
-//
+/*
+ * Copyright (c) 2017-2018 Snowflake Computing, Inc. All rights reserved.
+ */
 
--- a/include/snowflake/Param.hpp
+++ b/include/snowflake/Param.hpp
@@ -1,0 +1,40 @@
+/*
+ * Copyright (c) 2017-2018 Snowflake Computing, Inc. All rights reserved.
+ */
+
+#ifndef SNOWFLAKECLIENT_PARAM_HPP
+#define SNOWFLAKECLIENT_PARAM_HPP
+
+#include "client.h"
+
+namespace Snowflake {
+    namespace Client {
+        class Param {
+        public:
+            Param();
+
+            ~Param();
+
+            size_t len();
+
+            void set_len(size_t len_);
+
+            SF_DB_TYPE dbDataType();
+
+            void set_dbDataType(SF_DB_TYPE type_);
+
+            SF_C_TYPE cDataType();
+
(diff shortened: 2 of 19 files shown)


cpp/lib/Exceptions.cpp
```diff
@@ -0,0 +1,4 @@
+//
+// Created by vagrant on 2/27/18.
```
smtakeda: please remove this.

include/snowflake/Param.hpp
```diff
@@ -0,0 +1,40 @@
...
+    namespace Client {
+        class Param {
+        public:
+            Param();
+
+            ~Param();
+
+            size_t len();
+
+            void set_len(size_t len_);
```
smtakeda: setLength?

include/snowflake/Param.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+            SF_DB_TYPE dbDataType();
+
+            void set_dbDataType(SF_DB_TYPE type_);
```
smtakeda: setDbDataType?

include/snowflake/Param.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+            SF_C_TYPE cDataType();
+
+            void set_cDataType(SF_C_TYPE type_);
```
smtakeda: setCDataType?

include/snowflake/Param.hpp
```diff
@@ -0,0 +1,40 @@
...
+
+            void* value();
+            
+            void set_value(void *value_, SF_C_TYPE type_);
```
smtakeda: setValue?

include/snowflake/Statement.hpp
```diff
@@ -0,0 +1,77 @@
...
+            // TODO add method parameters to create a SF_BIND_INPUT
+            Param& createParam();
+
+            void destroyParams();
+
+            Column& column(size_t i);
+
+            const char *sfqid();
+
+            const std::string err_msg();
```
smtakeda: errMsg?

tests/test_cpp_select1.cpp
```diff
@@ -0,0 +1,40 @@
...
+        cnx->connect();
+        Snowflake::Client::Statement *stmt = new Snowflake::Client::Statement(*cnx);
+        std::string command("select 1;");
+        stmt->query(command);
+
+        int counter = 0;
+        while ((status = stmt->fetch()) == SF_STATUS_SUCCESS) {
+            assert_int_equal(stmt->column(1).asInt32(), 1);
+            ++counter;
+        }
```
smtakeda: This looks much cleaner now.
So asInt32() method will defer converting the value to here instead of fetching and converting in fetch method, right?

include/snowflake/Param.hpp
notkriswagner: I can change all of the set functions that format. At school the convention is: 

```
// set function
void set_<variable_name>(int value)

// get function
int <variable_name>()
```

PR #53, opened 2018-05-01
Add realloc memory hook for fetching data

smtakeda:
This is an enhancement for PDO for Snowflake to fix:

https://github.com/snowflakedb/pdo_snowflake/issues/148

The problem is if multiple columns use STRING data type, and you fetch data, the size of buffers are allocated to the number of columns multiplied by 16MB, too much memory consumption.
It is due to the C API design, which enforces the applications to provider buffer before fetching data.

The fix is have `snowflake_fetch` function reallocate the memory if the given maximum buffer size is not large enough and return the new pointer. It is achieved by the statement attribute `SF_STMT_USER_REALLOC_FUNC` that is called.

### DIFF
--- a/include/snowflake/client.h
+++ b/include/snowflake/client.h
@@ -215,7 +215,7 @@ typedef enum SF_GLOBAL_ATTRIBUTE {
  * Attributes for Snowflake statement context.
  */
 typedef enum SF_STMT_ATTRIBUTE {
-    INTERNAL
+    SF_STMT_USER_REALLOC_FUNC
 } SF_STMT_ATTRIBUTE;
 
 /**
@@ -322,6 +322,12 @@ typedef struct SF_STMT {
     SF_COLUMN_DESC *desc;
     void *stmt_attrs;
     sf_bool is_dml;
+
+    /**
+     * User realloc function used in snowflake_fetch
+     */
+    void *(*user_realloc_func)(void*, size_t);
+
     SF_CHUNK_DOWNLOADER *chunk_downloader;
     SF_PUT_GET_RESPONSE *put_get_response;
 } SF_STMT;
@@ -577,7 +583,7 @@ snowflake_stmt_set_attr(SF_STMT *sfstmt, SF_STMT_ATTRIBUTE type,
  * @return 0 if success, otherwise an errno is returned.
  */
 SF_STATUS STDCALL
-snowflake_stmt_get_attr(SF_STMT *sfstmt, SF_STMT_ATTRIBUTE type, void *value);
+snowflake_stmt_get_attr(SF_STMT *sfstmt, SF_STMT_ATTRIBUTE type, void **value);
 
 /**
  * Executes a statement.
--- a/lib/client.c
+++ b/lib/client.c
@@ -1407,10 +1407,9 @@ SF_STATUS STDCALL snowflake_fetch(SF_STMT *sfstmt) {
                         break;
                     default:
                         slen = strlen(raw_result->valuestring);
-                        log_debug("slen: %llu, buflen: %llu:%llu, realloc: %p",
-                                     (unsigned long)slen,
+                        log_debug("slen: %llu, maxbuflen: %llu, realloc func: %p",
+                                     slen,
                                      result->max_length,
-                                     result->len,
                                      (void*)sfstmt->user_realloc_func);
                         result->len = slen;
                         if (sfstmt->user_realloc_func != NULL &&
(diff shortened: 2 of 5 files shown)


include/snowflake/client.h
```diff
@@ -322,6 +322,12 @@ typedef struct SF_STMT {
     SF_COLUMN_DESC *desc;
     void *stmt_attrs;
     sf_bool is_dml;
+
+    /**
+     * User realloc function used in snowflake_fetch
+     */
+    void *(*user_realloc_func)(void*, size_t);
+
```
smtakeda: A new member of SF_STMT.

include/snowflake/client.h
```diff
@@ -577,7 +583,7 @@ snowflake_stmt_set_attr(SF_STMT *sfstmt, SF_STMT_ATTRIBUTE type,
  * @return 0 if success, otherwise an errno is returned.
  */
 SF_STATUS STDCALL
-snowflake_stmt_get_attr(SF_STMT *sfstmt, SF_STMT_ATTRIBUTE type, void *value);
+snowflake_stmt_get_attr(SF_STMT *sfstmt, SF_STMT_ATTRIBUTE type, void **value);
```
smtakeda: The return value must be a pointer of the pointer.

lib/client.c
```diff
@@ -1405,10 +1406,22 @@ SF_STATUS STDCALL snowflake_fetch(SF_STMT *sfstmt) {
...
                     default:
-                        /* copy original data as is except Date/Time/Timestamp/Binary type */
+                        slen = strlen(raw_result->valuestring);
+                        log_debug("slen: %llu, maxbuflen: %llu, realloc func: %p",
+                                     slen,
+                                     result->max_length,
+                                     (void*)sfstmt->user_realloc_func);
+                        result->len = slen;
+                        if (sfstmt->user_realloc_func != NULL &&
+                            slen > result->max_length) {
```
smtakeda: The buffer is reallocated ONLY IF:
- SF_STMT_USER_REALLOC_FUNC is set to the realloc function
- the buffer size is large enough to copy the data.

tests/data/.gitignore
```diff
@@ -0,0 +1 @@
+test_parallel_upload
```
smtakeda: Ignore the tmp data directory.

tests/test_adjust_fetch_data.c
```diff
@@ -0,0 +1,66 @@
...
+    // size is not large enough to fit.
+    snowflake_stmt_set_attr(sfstmt, SF_STMT_USER_REALLOC_FUNC, realloc);
+
+    status = snowflake_query(sfstmt, "select randstr(100,random()) from table(generator(rowcount=>2))", 0);
+    if (status != SF_STATUS_SUCCESS) {
+        dump_error(&(sfstmt->error));
+    }
+    assert_int_equal(status, SF_STATUS_SUCCESS);
+
+    size_t init_size = 10;
```
smtakeda: initial buffer size is 10.

tests/test_adjust_fetch_data.c
```diff
@@ -0,0 +1,66 @@
...
+        .value = calloc(1, init_size),
+        .len = init_size,
+        .max_length = init_size
+    };
+    snowflake_bind_result(sfstmt, &c1);
+    assert_int_equal(snowflake_num_rows(sfstmt), 2);
+
+    int counter = 0;
+    while ((status = snowflake_fetch(sfstmt)) == SF_STATUS_SUCCESS) {
+        assert_int_equal(strlen(c1.value), 100);
```
smtakeda: The result buffer requires 100 bytes.

tests/test_adjust_fetch_data.c
```diff
@@ -0,0 +1,66 @@
...
+        ++counter;
+    }
+    if (status != SF_STATUS_EOF) {
+        dump_error(&(sfstmt->error));
+    }
+    assert_int_equal(status, SF_STATUS_EOF);
+    free(c1.value);
```
smtakeda: Free the memory.

tests/test_adjust_fetch_data.c
```diff
@@ -0,0 +1,66 @@
...
+        dump_error(&(sf->error));
+    }
+    assert_int_equal(status, SF_STATUS_SUCCESS);
+
+    /* query */
+    SF_STMT *sfstmt = snowflake_stmt(sf);
+
+    // realloc function is used to allocate a new memory when the given buffer
+    // size is not large enough to fit.
+    snowflake_stmt_set_attr(sfstmt, SF_STMT_USER_REALLOC_FUNC, realloc);
```
smtakeda: This is the realloc for CAPI to call.

NOTE: we cannot assume CPI always use `realloc`, the standard C API function. The application may have its own memory functions, e.g., PHP ecalloc, erealloc, efree, so the function has to be specified by the application.

howryu:
LGTM

PR #54, opened 2018-05-18
Get command

howryu:
Support parallel download and multi part download

TODO:
Handle token renew.

### DIFF
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -361,14 +361,14 @@ RemoteStorageRequestOutcome Snowflake::Client::FileTransferAgent::downloadSingle
   std::basic_fstream<char> dstFile(dstNameFull.c_str(),
                                 std::ios_base::out | std::ios_base::binary);
 
-  Crypto::CipherIOStream * decryptOutputStream = new Crypto::CipherIOStream(
+  Crypto::CipherIOStream decryptOutputStream(
                                dstFile,
                                Crypto::CryptoOperation::DECRYPT,
                                fileMetadata->encryptionMetadata.fileKey,
                                fileMetadata->encryptionMetadata.iv, 128);
 
   RemoteStorageRequestOutcome outcome = client->download(fileMetadata,
-                                                         decryptOutputStream);
+                                                         &decryptOutputStream);
 
   dstFile.close();
 
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -14,6 +14,7 @@
 #include "crypto/Cryptor.hpp"
 #include "util/CompressionUtil.hpp"
 #include "util/ThreadPool.hpp"
+#include "EncryptionProvider.hpp"
 
 using ::std::string;
 using ::std::vector;
@@ -29,6 +30,11 @@ Snowflake::Client::FileTransferAgent::FileTransferAgent(
 
 Snowflake::Client::FileTransferAgent::~FileTransferAgent()
 {
+  reset();
+}
+
+void Snowflake::Client::FileTransferAgent::reset()
+{
   if (m_executionResults != nullptr)
   {
     delete m_executionResults;
@@ -38,11 +44,16 @@ Snowflake::Client::FileTransferAgent::~FileTransferAgent()
   {
     delete m_storageClient;
   }
+
+  m_largeFilesMeta.clear();
+  m_smallFilesMeta.clear();
 }
 
 FileTransferExecutionResult *
 Snowflake::Client::FileTransferAgent::execute(string *command)
 {
+  reset();
+
   // first parse command
   if (!m_stmtPutGet->parsePutGetCommand(command, &response))
   {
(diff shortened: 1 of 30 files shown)


tests/test_parallel_upload_download.cpp
```diff
@@ -115,6 +115,18 @@ void test_parallel_upload_core(int fileNumber)
...
+  std::string getCommand = "get @%test_parallel_upload_download file://" + dataDir + "test_parallel_download";
+  result = agent.execute(&getCommand);
+  assert_int_equal(fileNumber, result->getResultSize());
+
+  while(result->next())
+  {
+    assert_string_equal("SUCCEED", result->getStatus());
+  }
+
+  std::string compCmd = "diff " + dataDir + "test_parallel_upload/ " + dataDir + "test_parallel_download/";
```
smtakeda: diff won't work on windows.

tests/test_parallel_upload_download.cpp
```diff
@@ -115,6 +115,18 @@ void test_parallel_upload_core(int fileNumber)
   }
   assert_int_equal(status, SF_STATUS_EOF);
 
+  std::string getCommand = "get @%test_parallel_upload_download file://" + dataDir + "test_parallel_download";
```
smtakeda: where is dataDir? can multiple tests run in a single machine?

cpp/FileTransferAgent.cpp
```diff
@@ -250,3 +277,101 @@ void Snowflake::Client::FileTransferAgent::compressSourceFile(
...
+    fileMetadata->destFileName;
+
+  std::basic_fstream<char> dstFile(dstNameFull.c_str(),
+                                std::ios_base::out | std::ios_base::binary);
+
+  Crypto::CipherIOStream decryptOutputStream(
+                               dstFile,
+                               Crypto::CryptoOperation::DECRYPT,
+                               fileMetadata->encryptionMetadata.fileKey,
+                               fileMetadata->encryptionMetadata.iv, 128);
```
smtakeda: 128 can be hardcoded here? There is a case of 256 encryption.

cpp/FileTransferAgent.cpp
```diff
@@ -250,3 +277,101 @@ void Snowflake::Client::FileTransferAgent::compressSourceFile(
...
 }
+
+void Snowflake::Client::FileTransferAgent::download(string *command)
+{
+  m_executionResults = new FileTransferExecutionResult(CommandType::DOWNLOAD,
+    m_largeFilesMeta.size() + m_smallFilesMeta.size());
+
+  std::string createDirIfNotExist = "mkdir -p " +
+    std::string(response.GetLocalLocation());
+  const int err = system(createDirIfNotExist.c_str());
```
smtakeda: why you cannot use mkdir function?

cpp/SnowflakeS3Client.cpp
```diff
@@ -299,29 +308,193 @@ void SnowflakeS3Client::addUserMetadata(
...
-    return TransferOutcome::FAILED;
+    return RemoteStorageRequestOutcome::FAILED;
+  }
+}
+
+RemoteStorageRequestOutcome SnowflakeS3Client::download(
+  FileMetadata *fileMetadata,
+  std::basic_iostream<char>* dataStream)
+{
+  if (fileMetadata->srcFileSize > DATA_SIZE_THRESHOLD)
```
smtakeda: so what was your solution to get the size before download?

tests/test_parallel_upload_download.cpp
howryu: I understand that. Maybe need better ways to handle this once moving to windows. Either I need to do line by line checking or calculate the hash of files. Or some native windows system call to diff on two directories.

tests/test_parallel_upload_download.cpp
howryu: dataDir is under <workspace>/libsnowflakeclient/tests/data/

It is fine as long as each test run has its own local git repo.

cpp/FileTransferAgent.cpp
howryu: It is random value. I should not use 128 here, it is confusing. This value is encryption block size, i.e. for one encryption call, how many bytes of data is encrypted. I will change this to a MACRO ENC_DEC_BLOCK_SIZE. I think hard code value is fine. I did not see a point to add a parameter just for that.

cpp/FileTransferAgent.cpp
howryu: AFAIK, mkdir system call did not provide "-p", I want to create directories not failing if there is one already existed.

cpp/SnowflakeS3Client.cpp
howryu: FileMetadataInitializer class will issue a aws s3 rest request to get object metadata of the file size. (This is how s3 transfer library did as well)

tests/test_parallel_upload_download.cpp
smtakeda: ok, then add the file name to .gitignore such that somebody accidentally add to the repo.

cpp/FileTransferAgent.cpp
smtakeda: I mean this is not portable across platforms.

cpp/SnowflakeS3Client.cpp
smtakeda: ok

cpp/FileTransferAgent.cpp
smtakeda: ok, I just wondered it is related to the parameter CLIENT_ENCRYPTION_KEY_SIZE, which can be 128 or 256.
Since 128 is used in multiple places, please define a macro for better doc. It doesn't need to be a parameter.

PR #55, opened 2018-05-28
Get token renew

howryu:
This pr contains multiple changes.
- Token renew logic in get command
- Add Aws Logger implementation to merge aws log into snowflake logger
- In get command, encryption material could be different for different files. So handle this case correctly.
- Add interface class in include/snowflake directory so that only interface is exposed to Snowflake ODBC.


cpp/FileTransferAgent.cpp
```diff
@@ -103,13 +109,24 @@ void Snowflake::Client::FileTransferAgent::initFileMetadata()
...
+            sf_log_debug(CXX_LOG_NS,
+                         "Token expired when getting download metadata");
+            this->renewToken(command);
+            i--;
+          }
+          break;
+        }
       default:
         sf_log_error(CXX_LOG_NS, "Invalid command type");
+        throw;
```
smtakeda: what exception will be thrown by this?

tests/test_simple_put.cpp
```diff
@@ -172,7 +172,7 @@ void test_simple_put_one_byte(void **unused)
 
 static int gr_setup(void **unused)
 {
-  initialize_test(SF_BOOLEAN_FALSE);
+  initialize_test(SF_BOOLEAN_TRUE);
```
smtakeda: what is this parameter?

cpp/logger/SFAwsLogger.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+#include <aws/core/utils/logging/LogSystemInterface.h>
+
+using Aws::Utils::Logging::LogLevel;
+
+namespace Snowflake
+{
+namespace Client
+{
+namespace Logger
```
smtakeda: Do we have multiple logger? If not, we don't need to have this namespace.

cpp/FileTransferAgent.cpp
howryu: We should throw an SnowflakeException and marked it as internal error since invalid command type should never happen. There are other places that exception need to be thrown. I will revisit this exception handling in a separate commit. (This is done before Kris introduced SnowflakeException class, so I did not throw any explicit exception)

tests/test_simple_put.cpp
howryu: Will change to back to false. Basically this is used to control whether log will be printed to console or not. I enabled it for debugging purpose, but forget to change it back.

cpp/logger/SFAwsLogger.hpp
howryu: Ok will remove this one.

PR #58, opened 2018-06-05
Improve logging interface so that external logger can be injected

howryu:
Mainly for ODBC to inject logger


cpp/logger/SFAwsLogger.cpp
```diff
@@ -34,11 +44,20 @@ void Snowflake::Client::SFAwsLogger::LogStream(LogLevel logLevel,
                                                const char *tag,
                                                const Aws::OStringStream &messageStream)
 {
-  if (logLevel != LogLevel::Off)
+  /*if (logLevel != LogLevel::Off)
```
howryu: Will remove this block.

PR #60, opened 2018-06-09
Running macos put/get tests

howryu:
Running put/get tests on MacOS. No code change. Only update the CMakeLists.txt and tests/CMakeLIsts.txt


smtakeda:
looks good.

PR #62, opened 2018-06-16
Run put/get command tests on windows

howryu:


scripts/build_libsnowflakeclient.bat
```diff
@@ -18,6 +18,9 @@ if not "%VisualStudioVersion%"=="14.0" (
     if %ERRORLEVEL% NEQ 0 goto :error
 )
 
+if not exist %scriptdir%\..\deps-build\%arcdir%\aws\ call "%scriptdir%\build_awssdk.bat" %platform% %build_type%
+if %ERRORLEVEL% NEQ 0 goto :error
```
howryu: Basically, appveyor will build aws sdk every time it runs tests. Since aws sdk on windows is too big such that github rejected the commit.

cpp/util/CompressionUtil.cpp
```diff
@@ -11,10 +11,22 @@
 #define WINDOW_BIT 15
 #define GZIP_ENCODING 16
 
+#if defined(MSDOS) || defined(OS2) || defined(WIN32) || defined(__CYGWIN__)
```
smtakeda: We don't support MSDOS, OS2 and __CYGWIN__

scripts/build_libsnowflakeclient.bat
smtakeda: ACK

tests/utils/TestSetup.cpp
```diff
@@ -46,3 +46,10 @@ Snowflake::Client::Connection* TestSetup::connectionWithAutocommitFactory(const
     }
     return cnx;
 }
+
+std::string TestSetup::getDataDir()
+{
+  const std::string current_file = __FILE__;
```
smtakeda: How did you locate the current file? __FILE__ refers to the file at compile time.

tests/utils/TestSetup.hpp
```diff
@@ -9,12 +9,22 @@
 #include <snowflake/Connection.hpp>
 #include <snowflake/Statement.hpp>
 
+#ifdef _WIN32 
+#define PATH_SEP '\\'
+#else
+#define PATH_SEP '/'
```
smtakeda: platform.h has this.

lib/platform.c
```diff
@@ -453,4 +509,41 @@ void STDCALL sf_log_timestamp(char *tsbuf, size_t tsbufsize) {
...
+  return system(rmCmd);
+#endif
+}
+
+void STDCALL sf_get_tmp_dir(char * tmpDir)
+{
+#ifdef _WIN32
+  GetTempPath(100, tmpDir);
+#else
+  strncpy(tmpDir, "/tmp/", sizeof("/tmp/"));
```
smtakeda: Can you get TMP environment variable followed by TEMP?

cpp/util/CompressionUtil.cpp
howryu: Will remove it.

tests/utils/TestSetup.cpp
howryu: Well, as long as tests are compiled and run on the same machine. It should be fine. Actually, if you run tests locally, `__FILE__` is the best way to get current file/directory. Otherwise, you will need to specify environment variable.

smtakeda: ok, fair enough. Unlike the internal Jenkins build, the current Travis job includes both build and tests. It work unless we split the job.

smtakeda:
@howryu except the test failures on Appveyor, the change looks good.

PR #65, opened 2018-06-19
Exception

howryu:
Add exception handling. 

Right now, I am not sure how @notkriswagner want to design exception class, so I just decided to come up with a class of my own. I can merge it back to exception framework (if there is one) later.

### DIFF
--- a/include/snowflake/SnowflakeTransferException.hpp
+++ b/include/snowflake/SnowflakeTransferException.hpp
@@ -14,7 +14,13 @@ namespace Client
 
 enum TransferError
 {
-  INTERNAL_ERROR
+  INTERNAL_ERROR,
+  COMPRESSION_ERROR,
+  MKDIR_ERROR,
+  UNSUPPORTED_FEATURE,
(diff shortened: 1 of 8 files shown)


smtakeda:
For ODBC driver, the current implementation is fine. Later when we expose the C++ API, we may fix or improve it.

include/snowflake/SnowflakeTransferException.hpp
```diff
@@ -0,0 +1,46 @@
...
+  INTERNAL_ERROR,
+  COMPRESSION_ERROR,
+  MKDIR_ERROR,
+  UNSUPPORTED_FEATURE,
+  COLUMN_INDEX_OUT_OF_RANGE,
+  DIR_OPEN_ERROR,
+  COMPRESSION_NOT_SUPPORTED
+};
+
+class SnowflakeTransferException : public std::exception
```
notkriswagner: This is actually very similar to the base exception class that I have for the C++ wrapper work, which just contains `getCode()` and `what()` functions and private variables to hold the code and msg. I'll pull this change and work it into my branch.

PR #72, opened 2018-06-28
Osx 10.11

howryu:


deps-build/darwin/aws/lib/cmake/AWSSDK/AWSSDKConfigVersion.cmake
```diff
@@ -19,13 +19,13 @@ else()
 endif()
 
 # if the installed or the using project don't have CMAKE_SIZEOF_VOID_P set, ignore it:
-if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "" OR "8" STREQUAL "")
+if("${CMAKE_SIZEOF_VOID_P}" STREQUAL "" OR "4" STREQUAL "")
```
smtakeda: can you elaborate how this would work for MacOSX?

deps/curl-7.58.0/docs/curl.1
```diff
@@ -22,8 +22,7 @@
 .\"
 .\" DO NOT EDIT. Generated by the curl project gen.pl man page generator.
 .\"
-.TH curl 1 "November 16, 2016" "Curl 7.58.0" "Curl Manual"
-
+.TH curl 1 "16 Dec 2016" "Curl 7.52.0" "Curl Manual"
```
smtakeda: This is an interesting update. Did you have 7.52.0. somewhere on your machine?

deps-build/darwin/aws/lib/cmake/AWSSDK/AWSSDKConfigVersion.cmake
howryu: This is generated by aws sdk automatically. Not sure if how this worked.

deps/curl-7.58.0/docs/curl.1
howryu: I am not sure. My machine has libcurl 7.54.0. I have no idea where this comes from.

Issue #76, opened 2018-08-08
SF_STMT->total_fieldcount returns -1 sometimes after query is executed


notkriswagner:
The total_fieldcount field in SF_STMT struct sometimes returns -1 after a DDL query is executed. For DDL queries, this field should be set to 0.

notkriswagner:
Looks like this is caused when an empty JSON response is returned from the server. Investigating why we get back an empty JSON response.

notkriswagner:
The empty JSON response occurs when we incorrectly label a query as being a PUT/GET query when it is not. An example of this is the query `CREATE OR REPLACE TABLE obfuscated_table_name (daily_budget number(38,6))`. The regex pattern matches to the "get" part of the column name "daily_budget" and incorrectly labels this as a PUT/GET query.

The regex pattern at fault is `^([ ]*\\/*.*\\/*[ ]*)*(put|get)[ ]+`. When I put this through a regex tester (and from looking at the snowflake-connector-python), I see that changing the first regex group to be a non-capturing group (`^(?>[ ]*\\/*.*\\/*[ ]*)*(put|get)[ ]+`) solves the problem and the false positive matches go away.

The `regex.h` C library does not have support for non-capturing groups in regex patterns. There are C libraries that do have support for non-capturing groups, such as [PCRE2](https://www.pcre.org/), but that would also mean adding another library to the C API

smtakeda:
PCRE2 seems overkills just for PUT and GET command detection. @notkriswagner isn't the problem the existing regex pattern is not specific enough for it? You could capture the groups and can ignore them?

smtakeda:
`REG_NOSUB` looks like to report success or failure.

smtakeda:
It is very strange `^` BOL is ignored...
```
regcomp(&put_get_regex, "^([ ]*\\/*.*\\/*[ ]*)*(put|get)[ ]+",
            REG_ICASE | REG_EXTENDED);
```
PUT and GET must start from the beginning excluding comments.

notkriswagner:
In C with `regex.h` you can't ignore groups that you capture (see: https://stackoverflow.com/questions/40816011/posix-regular-expression-non-capturing-group).

The problem is the part that is supposed to ignore the comments. I'm trying to figure out how to make it more restrictive without using a non-capturing group.

smtakeda:
@notkriswagner can you put the pointer to the code change? Just push your branch to here for review. 

[...]

PR #78, opened 2018-08-13
Fixed PUT/GET Detection

notkriswagner:
Changed the regex string that does PUT/GET detection to reduce the number of false positives (queries that are labeled as PUT/GET queries but are not). Also fixed an issue where a query error would not get propagated to the user. Fixed some whitespace issues in connection.c

### DIFF
--- a/lib/platform.c
+++ b/lib/platform.c
@@ -365,7 +365,7 @@ sf_bool STDCALL _is_put_get_command(char *sql_text) {
     regex_t put_get_regex;
     // On MacOS seems \s to match white space character did not work. Change to '[ ]' for now
     //TODO maybe regex compilation should be moved to static variable so that no recompilation needed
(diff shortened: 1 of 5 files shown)


notkriswagner:
PR for issue #76

lib/platform.c
```diff
@@ -365,7 +365,7 @@ sf_bool STDCALL _is_put_get_command(char *sql_text) {
     regex_t put_get_regex;
     // On MacOS seems \s to match white space character did not work. Change to '[ ]' for now
     //TODO maybe regex compilation should be moved to static variable so that no recompilation needed
-    regcomp(&put_get_regex, "^([ ]*\\/*.*\\/*[ ]*)*(put|get)[ ]+",
+    regcomp(&put_get_regex, "^([ ]*\\/\\*.*\\*\\/[ ]*)*([ ]*)*(put|get)[ ]+",
```
smtakeda: Thanks for a patch. I think ultimately we should not parse the query in the clients but the server. At this moment, that's fine.

PR #79, opened 2018-08-15
force S3Client to use virtual addressing style

zad:
pasted CI tests and ready to merge

### DIFF
--- a/cpp/SnowflakeS3Client.cpp
+++ b/cpp/SnowflakeS3Client.cpp
@@ -76,8 +76,10 @@ SnowflakeS3Client::SnowflakeS3Client(StageInfo *stageInfo, unsigned int parallel
     Aws::String(stageInfo->credentials.at(AWS_SECRET_KEY)),
     Aws::String(stageInfo->credentials.at(AWS_TOKEN)));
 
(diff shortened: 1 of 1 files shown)


cpp/SnowflakeS3Client.cpp
```diff
@@ -76,8 +76,8 @@ SnowflakeS3Client::SnowflakeS3Client(StageInfo *stageInfo, unsigned int parallel
     Aws::String(stageInfo->credentials.at(AWS_SECRET_KEY)),
     Aws::String(stageInfo->credentials.at(AWS_TOKEN)));
 
-  s3Client = new Aws::S3::S3Client(credentials, clientConfiguration);
-
+  s3Client = new Aws::S3::S3Client(credentials, clientConfiguration,
+          Aws::Client::AWSAuthV4Signer::PayloadSigningPolicy::Never, true);
```
smtakeda: can you please add a comment for `true`? By looking at this code, it is not clear what this means until we check the reference doc.

zad: That's a good point!

smtakeda:
@zad, I'll do the merge after the tests are completed and let you know the tag.

PR #82, opened 2018-09-05
Add support for Base 64 Url coding/decoding and testcases

ChTimTsubasa:
For Base64Url coding, we want to replace + with - and / with _.

This PR served as a library implementation for JWT token generation. Test cases are addressed.


cpp/util/Base64.cpp
```diff
@@ -40,9 +51,10 @@ size_t Base64::decodedLength(const void *const vsrc,
   }
 }
 
-size_t Base64::encode(const void *const vsrc,
-                      const size_t srcLength,
-                      void *const vdst) noexcept
+size_t Base64::encodeHelper(const void *const vsrc,
+                            const size_t srcLength,
+                            void *const vdst,
+                            const char *dictionary) noexcept
```
smtakeda: Was `encode` already used? or never? @howryu ?

smtakeda: I'm just wondering if the caller should also change if already used.

ChTimTsubasa: I believe the encode is used. One place is (https://github.com/snowflakedb/libsnowflakeclient/blob/master/cpp/EncryptionProvider.cpp#L43). But it does not matter since the encode()'s functionality and interface do not change.

smtakeda: ok, got it:
https://github.com/snowflakedb/libsnowflakeclient/pull/82/files#diff-953ac7085f8cf5f3c00e77fd87e3d61fR73

howryu:
LGTM as well.

smtakeda:
I'll bump up the version and add a tag.

PR #85, opened 2018-09-07
Fetch performance tests

notkriswagner:
Added set of performance tests to the C API to test fetch. Also added script to populate test data in a specific deployment.

PR #87, opened 2018-09-11
Added first set of performance tests 

notkriswagner:
Added first set of performance tests along with the necessary test infrastructure to support the performance tests as well as populate test data into the test deployment. Fixed error due to passing in NULL arraylist. Implemented total_row_index in C API as well to support performance tests.


notkriswagner:
This PR is a cleaned up version of PR #85. Assuming all tests pass, this branch will be committed and #85 will be deleted.

PR #88, opened 2018-09-14
Expose JWT and Base64 interface

ChTimTsubasa:


notkriswagner:
LGTM. Are you going to eventually create a C interface for this code?

include/snowflake/IBase64.hpp
```diff
@@ -0,0 +1,47 @@
...
+  const char *what() const throw()
+  {
+    return message_.c_str();
+  }
+
+private:
+  std::string message_;
+};
+
+  class IBase64
```
smtakeda: I'm not 100% sure, but would it make sense to have an interface even for Base64 class, which seems to me pretty solid and stable? In my understanding, the interface makes sense if there are multiple implementations and potentially the application may switch between them without changing the use of interface.  I'm just wondering if an interface would add unnecessary complexity in the code... I'm missing some use case. Note I'm not super good at C++.

smtakeda:
Overall looks good. Added minor comments.

include/snowflake/IBase64.hpp
ChTimTsubasa: There are two reasons to have this interface:
1. It's going to be used by ODBC driver, and this is the only place that a component in libsnowflakeclient can be exposed to the ODBC driver.

2. Our original base64 code's interface is too c style for a c++ code, which makes it hard to read and use in a high level application like ODBC driver. I would argue this interface is way much easier to use.

ChTimTsubasa:
@notkriswagner Probably not if the need for JWT in PHP is not urgent in the short term. But I would say it should not be too hard to implement a c wrapper.

cpp/jwt/Header.cpp
```diff
@@ -25,15 +26,15 @@ CJSONHeader::CJSONHeader()
   json_root_ = {snowflake_cJSON_CreateObject(), CJSONOperation::cJSONDeleter};
   if (json_root_ == nullptr)
   {
-    throw JwtMemoryAllocationFailure();
+    throw std::bad_alloc();
```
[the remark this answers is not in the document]
ChTimTsubasa: It is going to be catched in the ODBC driver. I was using JwtMemoryAllocationFailure at first but then I realize it make no sense to have a custom one if there is already an standard exception for memory allocation failure.

include/snowflake/IBase64.hpp
ChTimTsubasa: For your interface concern. Note that this is more of a wrapper rather than a interface in a strict sense, as you can see all the member function are static. I can change the name to Base64Wrapper is IBase64 leads to confusion.

ChTimTsubasa:
I would add some comments and push it later.

notkriswagner:
@howryu I'm going to merge this in a few hours. Speak now or forever hold your peace.

PR #91, opened 2018-09-18
New fetch api

notkriswagner:
Implemented new fetch API and removed old fetch code. SF_BIND_OUTPUT struct is no longer needed and has been removed. Types have been updated to make it so that our int32 and uint32 types are actually 32 bits.


include/snowflake/basic_types.h
```diff
@@ -9,15 +9,18 @@
...
 
 /**
  * Supported data types
  */
 typedef char int8;
 typedef unsigned char uint8;
-typedef unsigned long int uint32;
-typedef long int int32;
+typedef unsigned int uint32;
+typedef int int32;
```
notkriswagner: In the future we'll probably want to check type sizes on a particular platform to ensure that our types are the correct size.

tests/test_adjust_fetch_data.c
```diff
@@ -29,28 +29,22 @@ void test_select_long_data_with_small_initial_buffer(void **unused) {
...
+    
+    char *value = NULL;
     assert_int_equal(snowflake_num_rows(sfstmt), 2);
 
     int counter = 0;
     while ((status = snowflake_fetch(sfstmt)) == SF_STATUS_SUCCESS) {
-        assert_int_equal(strlen(c1.value), 100);
+        snowflake_column_as_str(sfstmt, 1, &value, NULL);
+        assert_int_equal(strlen(value), 100);
+        free(value);
```
smtakeda: @notkriswagner , this is ok for C/C++ API internally, but for PHP PDO and any other framework that has the memory manager, you'll need to allow the application to specify a struct including callbacks of `malloc`, `calloc`, `realloc` and `free` to match mem alloc and free. Depending on the application, calling `free` function here may not match `calloc`/`malloc`/`realloc` in the C/C++ API.

smtakeda: Or will snowflake_stmt_term free the memory for the user?

smtakeda: Let's discuss tomorrow.
I think it is better for snowflake_stmt_term to free the memory automatically.

tests/test_number.c
```diff
@@ -115,64 +116,34 @@ void test_number(void **unused) {
...
-        assert_string_equal(v.c2out, c2.value);
-        assert_string_equal(v.c3out, c3.value);
-        assert_string_equal(v.c4out, c4.value);
+        snowflake_column_as_int64(sfstmt, 1, &c1);
+        TEST_CASE_TO_STRING v = test_cases[c1 - 1];
+        snowflake_column_as_float64(sfstmt, 2, &float_val);
+        snowflake_column_as_str(sfstmt, 2, &str, NULL);
+        assert(float_val = v.c2in);
+        assert_string_equal(v.c2out, str);
+        free(str);
```
smtakeda: If the application is responsible for freeing memory, it is very likely that it will forget and leak memory..

tests/test_adjust_fetch_data.c
notkriswagner: I think it might be better to use memory hooks, but let's talk tomorrow

include/snowflake/client.h
```diff
@@ -693,6 +676,148 @@ const char *STDCALL snowflake_c_type_to_string(SF_C_TYPE type);
...
+/**
+ * Returns the raw column data in the form of a const char pointer that the user can then use
+ * to read the string data or copy to another buffer. A NULL column will return a NULL pointer
+ *
+ * @param sfstmt SF_STMT context
+ * @param idx Column index
+ * @param value_ptr Raw column data is stored in this pointer
+ * @return 0 if success, otherwise an errno is returned
+ */
+SF_STATUS STDCALL snowflake_column_as_const_str(SF_STMT *sfstmt, int idx, const char **value_ptr);
```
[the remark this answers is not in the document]
notkriswagner: Yes its always null terminated

lib/memory.h
```diff
@@ -18,6 +18,15 @@ extern "C" {
...
 #define SF_FREE(p) ((void) (sf_free(p, __FILE__, __LINE__), (p) = NULL))
 
+typedef struct SF_INTERNAL_MEM_HOOKS {
+    void *(*alloc)(size_t size);
+    void (*dealloc)(void *ptr);
+    void *(*realloc)(void *ptr, size_t size);
+    void *(*calloc)(size_t nitems, size_t size);
+} SF_INTERNAL_MEM_HOOKS;
+
+static SF_INTERNAL_MEM_HOOKS global_hooks = {malloc, free, realloc};
```
smtakeda: calloc?

smtakeda: @notkriswagner calloc is missing.

notkriswagner: Yup, I added calloc into the struct afterward, forgot to add it to the initialization.

PR #94, opened 2018-09-24
Timestamp type

notkriswagner:
Implemented SF_TIMESTAMP type as well as associated timestamp creation and get functions. Updated existing code to use new SF_TIMESTAMP type. Added a function to transform a column into a timestamp type. Added test case to test all get functions.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -2427,22 +2427,6 @@ SF_STATUS STDCALL snowflake_column_as_str(SF_STMT *sfstmt, int idx, char **value
                 goto cleanup;
             }
 
-//            if (!_extract_timestamp_dynamic(
-//              &value,
-//              &value_len,
-//              sfstmt->desc[idx - 1].type,
-//              column->valuestring,
-//              sfstmt->connection->timezone,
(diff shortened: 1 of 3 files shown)


notkriswagner:
This PR only deals with receiving data from the C API using the SF_TIMESTAMP struct. A future PR will deal with input binding of the timestamp type.

lib/client.c
```diff
@@ -2362,23 +2396,37 @@ SF_STATUS STDCALL snowflake_column_as_str(SF_STMT *sfstmt, int idx, char **value
...
         case SF_DB_TYPE_TIMESTAMP_LTZ:
-        case SF_DB_TYPE_TIMESTAMP_TZ:
-            if (!_extract_timestamp_dynamic(
-              &value,
-              &value_len,
-              sfstmt->desc[idx - 1].type,
-              column->valuestring,
-              sfstmt->connection->timezone,
-              sfstmt->desc[idx - 1].scale)) {
+        case SF_DB_TYPE_TIMESTAMP_TZ: ;
```
smtakeda: is this semicolon required?

PR #98, opened 2018-10-03
Compress clang compile error

ChTimTsubasa:
This PR fixed the clang compile error when compiling ODBC driver on Mac.


notkriswagner:
Appveyor job failures look unrelated to PR so I restarted the job and I expect it to succeed this time.

notkriswagner:
Otherwise LGTM

ChTimTsubasa:
There is a bug in the base64UrlDecode. 

Basically the Base64UrlDecoding is not looking at the correct reverse index table when checking with the final several bytes. 

This is fixed and test case that address the bug is created.

@notkriswagner Can you take another review on the new push? Thank you so much!

PR #99, opened 2018-10-04
Request GUID

notkriswagner:
Implemented a request GUID that is unique per request to Snowflake

### DIFF
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -13,6 +13,7 @@
 
 #define curl_easier_escape(curl, string) curl_easy_escape(curl, string, 0)
 #define QUERYCODE_LEN 7
+#define REQUEST_GUID_KEY_SIZE 13
 
 /*
  * Debug functions from curl example. Should update at somepoint, and possibly remove from header since these are private functions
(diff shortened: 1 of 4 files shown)


lib/connection.c
```diff
@@ -769,11 +796,24 @@ sf_bool STDCALL http_perform(CURL *curl,
 
     //TODO set error buffer
 
+    // Find request GUID in the supplied URL
+    char *request_guid_ptr = strstr(url, "request_guid=");
+    // Set pointer to the beginning of the UUID string if request GUID exists
+    if (request_guid_ptr) {
+        request_guid_ptr = request_guid_ptr + 13;
```
smtakeda: can this 13 be a constant or add a comment here?

lib/connection.c
```diff
@@ -769,11 +796,24 @@ sf_bool STDCALL http_perform(CURL *curl,
...
+    }
+
     do {
         // Reset buffer since this may not be our first rodeo
         SF_FREE(buffer.buffer);
         buffer.size = 0;
 
+        // Generate new request guid, if request guid exists in url
+        if (request_guid_ptr && uuid4_generate_non_terminated(request_guid_ptr)) {
```
smtakeda: This would be more clear that uuid4 generation failed.
```
if (request_guid_ptr && uuid4_generate_non_terminated(request_guid_ptr) != UUID4_ESUCCESS)
```

lib/connection.c
notkriswagner: Sure

PR #102, opened 2018-10-09
Proxy

notkriswagner:
Added code to enable proxy support for PUT/GET commands. The code reads the value from the env vars (all_proxy, http_proxy, https_proxy) and sets the appropriate configuration fields in AWS API.

### DIFF
--- a/tests/test_unit_proxy.cpp
+++ b/tests/test_unit_proxy.cpp
@@ -2,50 +2,51 @@
  * Copyright (c) 2018 Snowflake Computing, Inc. All rights reserved.
  */
 
+#include <cassert>
 #include "util/Proxy.hpp"
 #include "utils/test_setup.h"
 #include "utils/TestSetup.hpp"
 
 typedef ::Snowflake::Client::Util::Proxy Proxy;
 
-void test_proxy_parts_equality(const char *proxy_str, const char *user, const char *pwd, const char *machine, unsigned port, const char *scheme) {
+void test_proxy_parts_equality(const char *proxy_str, const char *user, const char *pwd, const char *machine, unsigned port, Proxy::Protocol scheme) {
     Proxy proxy(proxy_str);
 
     assert_string_equal(proxy.getUser().c_str(), user);
     assert_string_equal(proxy.getPwd().c_str(), pwd);
     assert_string_equal(proxy.getMachine().c_str(), machine);
     assert_int_equal(proxy.getPort(), port);
-    assert_string_equal(proxy.getScheme().c_str(), scheme);
+    assert(proxy.getScheme() == scheme);
 }
 
 void test_proxy_machine_only(void **unused)
 {
-    test_proxy_parts_equality("localhost", "", "", "localhost", 0, "");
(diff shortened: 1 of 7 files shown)


howryu:
Can you add some tests? Also maybe considering adding some util classes to parse proxy information?

tests/test_unit_proxy.cpp
```diff
@@ -0,0 +1,62 @@
...
+    assert_string_equal(proxy.getUser().c_str(), user);
+    assert_string_equal(proxy.getPwd().c_str(), pwd);
+    assert_string_equal(proxy.getMachine().c_str(), machine);
+    assert_int_equal(proxy.getPort(), port);
+    assert_string_equal(proxy.getScheme().c_str(), scheme);
+}
+
+void test_proxy_machine_only(void **unused)
+{
+    test_proxy_parts_equality("localhost", "", "", "localhost", 0, "");
```
smtakeda: would this be port 80?

tests/test_unit_proxy.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+void test_proxy_machine_and_port(void **unused)
+{
+    test_proxy_parts_equality("localhost:8081", "", "", "localhost", 8081, "");
+}
+
+void test_proxy_machine_and_scheme(void **unused)
+{
+    test_proxy_parts_equality("https://localhost", "", "", "localhost", 0, "https");
```
smtakeda: and this would be port 443?

tests/test_unit_proxy.cpp
```diff
@@ -0,0 +1,62 @@
...
+}
+
+void test_proxy_all(void **unused)
+{
+    test_proxy_parts_equality("https://someuser:somepwd@somewhere.com:5050", "someuser", "somepwd", "somewhere.com", 5050, "https");
+}
+
+void test_proxy_empty(void **unused)
+{
+    test_proxy_parts_equality("", "", "", "", 0, "");
```
smtakeda: port 0 is often used to get an unused port from the ephemeral port range.  In case of proxy, we use HTTP/HTTPS, so if no port is specified, it is better returning 80/443.

tests/test_unit_proxy.cpp
notkriswagner: I set it to 0 (which means all ports) if the port is not specified and don't set any value in the AWS SDK for proxy port. I don't want to assume a particular port to connect to a proxy server incase 80 is not the default.

tests/test_unit_proxy.cpp
notkriswagner: As I explained above, I use it to signify "no port specified", I could change the API to return 80/443 if you think that would be better

smtakeda: Except this test case where an empty input, I think the default port should be used.

notkriswagner: Okay, will update

notkriswagner: If the protocol isn't specified should I just assume HTTP?

PR #108, opened 2018-10-25
Issue 107

notkriswagner:
Add cstdlib to Proxy.hpp to fix build issue on OS X. Removed `brew update python` from travis script since python3 is installed by default.


notkriswagner:
@smtakeda and @h-chaturvedi review when you have a chance

PR #112, opened 2018-11-05
OCSP Skip

notkriswagner:
Added an ocsp skip based on the connection struct's insecure mode flag.


notkriswagner:
@smtakeda and @h-chaturvedi please review when you get a chance

PR #113, opened 2018-11-07
Fixed DATE type to SF_TIMESTAMP struct Conversion bug

notkriswagner:
Fixed DATE type to SF_TIMESTAMP struct conversion bug. Added a test case to check the native timestamp type column conversion from all Snowflake time/date types.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -95,11 +95,6 @@ static const int64 pow10_int64[] = {
 };
 
 /**
- * Maximum one-directional range of offset-based timezones (24 hours)
- */
-#define TIMEZONE_OFFSET_RANGE  (int64)(24 * 60);
-
-/**
  * Find string size, create buffer, copy over, and return.
  * @param var output buffer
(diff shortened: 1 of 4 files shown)


lib/client.c
```diff
@@ -152,136 +147,6 @@ static void log_lock_func(void *udata, int lock) {
     }
 }
 
-/**
```
smtakeda: oh, you no longer need this function?

notkriswagner: Yes, this function was broken up into two other functions (`snowflake_timestamp_from_epoch_seconds` and `snowflake_timestamp_to_string`) when I did the timestamp changes. I kept that function around incase I needed to reference it later, but I can just look back at an old commit if I want to reference this function so I thought I would just reduce the code in `client.c` since it is becoming quite the large file.

notkriswagner:
@h-chaturvedi please review when you get a chance

PR #115, opened 2018-11-09
Added support for named parameters

h-chaturvedi:


include/snowflake/client.h
```diff
@@ -335,6 +336,12 @@ typedef struct SF_STMT {
     int64 total_fieldcount;
     int64 total_row_index;
     void *params;
+    void **name_list;
+    union
+    {
+        int64 names_len;
+        int64 idx_len;
+    };
```
notkriswagner: Since these are the same type, you can probably just merge them into 1 variable called `params_len`

lib/client.c
```diff
@@ -1071,23 +1074,111 @@ SF_STMT *STDCALL snowflake_stmt(SF_CONNECT *sf) {
...
+        return SF_STATUS_ERROR_OTHER;
+    }
+    /*
+     * For named parameters, populate the name list in the
+     * sfstmt to help extract the params from the treemap
+     * during execute
+     */
+    if (_sf_get_current_param_style(sfstmt) == NAMED)
+    {
+        sfstmt->name_list = SF_REALLOC(sfstmt->name_list,
```
notkriswagner: Could we grow this list more than 1 element at a time? It seems like the realloc function will get called n times (where n is the number of input parameters for a sql query).

lib/client.c
```diff
@@ -1099,11 +1190,35 @@ SF_STATUS snowflake_bind_param_array(
...
+        sf_param_store_set(sfstmt->params, &sfbind_array[i],
+                sfbind_array[i].idx, sfbind_array[i].name);
+        /*
+         * For named parameters, populate the name list in the
+         * sfstmt to help extract the params from the treemap
+         * during execute
+         */
+        if (_sf_get_current_param_style(sfstmt) == NAMED)
+        {
+            sfstmt->name_list = SF_REALLOC(sfstmt->name_list,
```
notkriswagner: You could move this outside of the for loop and just do 1 memory allocation instead of many allocations.

lib/client.c
```diff
@@ -1353,7 +1468,7 @@ SF_STATUS STDCALL _snowflake_execute_ex(SF_STMT *sfstmt,
     sf_bool success = SF_BOOLEAN_FALSE;
     uuid4_generate(sfstmt->request_id);
     URL_KEY_VALUE url_params[] = {
-        {.key="requestId=", .value=sfstmt->request_id, .formatted_key=NULL, .formatted_value=NULL, .key_size=0, .value_size=0}
+            {.key="requestId=", .value=sfstmt->request_id, .formatted_key=NULL, .formatted_value=NULL, .key_size=0, .value_size=0}
```
notkriswagner: Not a big deal, but wondering why you added another tab here?

lib/client.c
```diff
@@ -18,6 +18,7 @@
 #include "chunk_downloader.h"
 
 #define curl_easier_escape(curl, string) curl_easy_escape(curl, string, 0)
+#define MAX_NAME_SIZE 20
```
notkriswagner: 20 is almost definitely too small. I would say at least 256 is your minimum here. There is no max size according to the JDBC standard (or at least I can't find any place that states a max size).

lib/client.c
h-chaturvedi: Just for consistency. There were two tabs for statements spanning multiple lines at other places   but I guess I haven't made that change everywhere. Will revert.

lib/client.c
```diff
@@ -1365,31 +1480,64 @@ SF_STATUS STDCALL _snowflake_execute_ex(SF_STMT *sfstmt,
...
+            type = snowflake_type_to_string(
+                    c_type_to_snowflake(input->c_type, SF_DB_TYPE_TIMESTAMP_NTZ));
+            value = value_to_string(input->value, input->len, input->c_type);
+            binding = snowflake_cJSON_CreateObject();
+
+            /*
+             * TODO - This may not be correct assumption for
+             * max name size
+             */
+            char name[MAX_NAME_SIZE];
```
notkriswagner: You can move this out of the for loop and dynamically allocate this buffer to something small (maybe 20 bytes or so). Then you can grow this buffer as needed. Since the `snowflake_CJSON_AddStringToObject` function just copies the string passed to it, you don't have to worry about persisting these values long term.

lib/client.c
notkriswagner: You can keep this, I see it is the same below. I was just curious mostly.

lib/client_int.h
```diff
@@ -6,7 +6,8 @@
 #define SNOWFLAKE_CLIENT_INT_H
 
 #include "cJSON.h"
-#include "arraylist.h"
+/*#include "arraylist.h"*/
```
notkriswagner: Please delete this line

lib/lib_common.h
```diff
@@ -0,0 +1,18 @@
+#ifndef SF_LIB_COMMON_H
+#define SF_LIB_COMMON_H
+
+typedef enum SF_RET_CODE
```
notkriswagner: Can you merge this file with `client_int.h`

notkriswagner: Also can you rename this to SF_INT*. Just so we know these are for internal use only.

lib/treemap.c
```diff
@@ -0,0 +1,237 @@
+/*
+** Copyright (c) 2017-2018 Snowflake Computing, Inc. All rights reserved.
+*/
+
+#include <string.h>
+#include "treemap.h"
+#include "memory.h"
+
+/* 
+** Hashmap needs to be initialized
```
notkriswagner: Treemap?

tests/test_bind_datastructures/CMakeLists.txt
```diff
@@ -0,0 +1,2 @@
+add_subdirectory(unit_tests)
+add_definitions(-DSF_DEBUG -DSF_ERROR)
```
notkriswagner: Newline please

lib/treemap.h
```diff
@@ -0,0 +1,60 @@
+/*
+ * Copyright (c) 2017-2018 Snowflake Computing, Inc. All rights reserved.
+ */
+
+#ifndef SNOWFLAKE_HASHMAP_H
+#define SNOWFLAKE_HASHMAP_H
```
h-chaturvedi: change to SNOWFLAKE_TREEMAP_H

lib/treemap.h
```diff
@@ -0,0 +1,60 @@
...
+** @return SF_STATUS 
+*/
+TREE_MAP * STDCALL sf_treemap_init(void);
+
+/* sf_treemap_set
+** insert void into treemap
+** @return- void
+*/
+
+/* TODO - Potentially change this from a void
```
notkriswagner: I think this TODO is already done?

lib/treemap.c
```diff
@@ -0,0 +1,237 @@
...
+** to a size which is sufficiently 
+** large and is a prime number
```
notkriswagner: I'm confused here. Here you say it needs to be initialized to a size that is sufficiently large and is a prime number, but `TREE_MAP_MAX_SIZE` is set to 1000 (which is large, but not prime)

h-chaturvedi: Stale comment from the hashmap implementation. Deleted.

lib/lib_common.h
h-chaturvedi: I had thought about putting this in client_int.h too but lib_common is included by both tree_map and rbtree both of which do not need the to know the information present in client_int.h and hence I made this. The file might look small now but the goal is to update the all libs/ to use these return types. This might warrant for a *_common.h for the lib folder.

lib/client.c
h-chaturvedi: This buffer was just a temporary buffer used to store the names. I realized this is not needed however, just pass (char *)(sfstmt->name_list[i]) directly to AddItemToObject.

lib/lib_common.h
notkriswagner: If the idea is that the whole library will use those return types internally, shouldn't we put it in `client_int.h` since the whole library will use those values? `client_int.h` is used to store code that is used across all of the lib files which would work well for the return types. This file is only 100 lines long and even if we end up having 100s of return types, thats still a small file. In the future we can split it up once this file gets larger, but right now I think its a bit overkill.

h-chaturvedi: cool

h-chaturvedi: On second thoughts I think it makes sense for now to have a separate file because including client_int.h as it is currently, exposes a lot of un-necessary code to unrelated files. For instance client_int.h includes client.h the contents of which do not make sense to be exposed to the data structure files like rbtree / treemap / paramstore etc. Essentially the main point here is code separation.

h-chaturvedi:
@smtakeda Just to confirm you're suggesting to update the error printfs and not the Compiler flags themselves.

h-chaturvedi:
@notkriswagner Can you review the latest changes once you find time .. Thanks ..

notkriswagner:
LGTM

PR #117, opened 2018-11-15
Remove constructor of EncryptionMaterial

ChTimTsubasa:
The parameter passed into the struct has the same name as member variable and caused a lot of warning thrown on the ODBC since it uses stricter compile flag. It would be a good chance to meet up with higher standard during the Simba upgrade this time.

Since the constructor seems to not being used anywhere, I just throw it away.

### DIFF
--- a/cpp/StatementPutGet.cpp
+++ b/cpp/StatementPutGet.cpp
@@ -28,7 +28,8 @@ bool StatementPutGet::parsePutGetCommand(std::string *sql,
     ->source_compression;
 
   cJSON *src = (cJSON *) response->src_list;
-  for (int i = 0; i < snowflake_cJSON_GetArraySize(src); i++)
+  int src_size = snowflake_cJSON_GetArraySize(src);
+  for (int i = 0; i < src_size; i++)
   {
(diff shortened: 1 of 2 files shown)


notkriswagner:
Looks like the constructor removal broke something

ChTimTsubasa:
Yes. The constructor is used by emblace_back(). That is why. I will restore it.

ChTimTsubasa:
@notkriswagner Can you review this PR?

cpp/StatementPutGet.cpp
```diff
@@ -28,7 +28,8 @@ bool StatementPutGet::parsePutGetCommand(std::string *sql,
     ->source_compression;
 
   cJSON *src = (cJSON *) response->src_list;
-  for (int i = 0; i < snowflake_cJSON_GetArraySize(src); i++)
+  int src_size = snowflake_cJSON_GetArraySize(src);
```
notkriswagner: +1

notkriswagner:
LGTM. This should help performance a bit too. Thanks for doing this

PR #118, opened 2018-12-07
Azure tests

notkriswagner:
Added Azure testing as well as fixing any issues that popped up because of azure tests


lib/chunk_downloader.c
```diff
@@ -233,7 +233,6 @@ SF_CHUNK_DOWNLOADER *STDCALL chunk_downloader_init(const char *qrmk,
     // We need thread_count, fetch_slots, chunks, and either qrmk or chunk_headers
     if (thread_count <= 0 ||
             fetch_slots <= 0 ||
-            !(chunk_headers || qrmk) ||
```
notkriswagner: Large result set functionality in Azure does not require either chunk_headers or qrmk to be set

lib/connection.c
```diff
@@ -831,10 +831,12 @@ sf_bool STDCALL http_perform(CURL *curl,
             curl_easy_setopt(curl, CURLOPT_VERBOSE, 1);
         }
 
-        res = curl_easy_setopt(curl, CURLOPT_HTTPHEADER, header);
-        if (res != CURLE_OK) {
-            log_error("Failed to set header [%s]", curl_easy_strerror(res));
-            break;
+        if (header) {
```
notkriswagner: If header is NULL, don't set it

PR #121, opened 2019-01-31
Libsnowflakeclient cjwt wrapper

h-chaturvedi:
Adds a C Wrapper for JWT implementation. These APIs will for now be consumed by the ODBC driver.


notkriswagner:
I re-ran the windows tests cause it looked like we were unable to communicate with the server.

h-chaturvedi:
@notkriswagner There was one log line that was being dumped so many times due to a test. I have removed it now. That has been committed independent of this branch [here](https://github.com/snowflakedb/libsnowflakeclient/commit/3f7b62f31011e37d7714c69850bf7382fb74177b).

include/snowflake/jwtWrapper.hpp
```diff
@@ -0,0 +1,52 @@
+#ifndef JWT_WRAPPER_H
```
smtakeda: Add copyright header

cpp/jwt/jwtWrapper.cpp
```diff
@@ -0,0 +1,208 @@
+#include "snowflake/IJwt.hpp"
```
smtakeda: add copyright header.

include/snowflake/jwtWrapper.hpp
smtakeda: I think *.h is better as this is for C.

notkriswagner:
LGTM

PR #122, opened 2019-02-02
JWT Enhancement - support direct token verification

h-chaturvedi:
Enhanced JWT to support verification of a base64 encoded JWT Token directly without the need of parsing it into a JWTObject. This bypasses the steps to serialize Header and ClaimSet Objects while signature verification.

### DIFF
--- a/cpp/jwt/jwtWrapper.cpp
+++ b/cpp/jwt/jwtWrapper.cpp
@@ -130,7 +130,20 @@ const char * CJWT_serialize(CJWT cjwt_obj, EVP_PKEY *key)
 int CJWT_verify(CJWT cjwt_obj, EVP_PKEY *key)
 {
     IJwt *ijwt_obj = static_cast<IJwt *>(cjwt_obj);
-    if ((ijwt_obj->verify(key, false)))
+    if ((ijwt_obj->verify(key)))
+    {
+        return 1;
+    }
+    else
+    {
+        return 0;
+    }
+}
+
+int CJWT_verify_token(CJWT cjwt_obj, EVP_PKEY *key, const char *token)
+{
+    IJwt *ijwt_obj = static_cast<IJwt *>(cjwt_obj);
+    if ((ijwt_obj->verify(key, token)))
     {
         return 1;
     }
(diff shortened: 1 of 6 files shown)


cpp/jwt/jwtWrapper.cpp
```diff
@@ -130,7 +130,20 @@ const char * CJWT_serialize(CJWT cjwt_obj, EVP_PKEY *key)
...
+    {
+        return 1;
+    }
+    else
+    {
+        return 0;
+    }
+}
+
+int CJWT_verify_token(CJWT cjwt_obj, EVP_PKEY *key, const char *token)
```
ChTimTsubasa: IMO it would probably be better if we can hide this `token` from the interface and make it a member of the JWTobject. Therefore the JWT object could be self contained once established without requiring the user to pass in the token.

include/snowflake/IJwt.hpp
```diff
@@ -213,7 +213,16 @@ class IJwt
    * Verify the JWT is valid using the public key
    * @usedBy authenticator
    */
-  virtual bool verify(EVP_PKEY *key, bool format) = 0;
```
ChTimTsubasa: Great!

h-chaturvedi:
> Great work on making this C wrapper on top.
> As for this specific PR, I would argue that let's hide the token inside the JWTObject instead of exposing it to user side. We could have less API then.
> Also it would be great if you can write a test case to address the base64/JWT encode issue, just as the one you gave me in the sample would be awesome.

Consolidating the two APIs into one would mean that we need to have a way of knowing what functionality do we need. Currently there are two ways token verification can happen :
1. Via verification of serialized claimset and serialized header.
2. Via direct verification of the string token (without serializing it into the above).
The extra "format" argument in the API could have potentially helped to distinguish between the needed functionality but that was removed based on prior comments.

Note - For SSD Verification the first approach is not only not required, but it changes the token in a way that fails verification. 

I agree we can ingrain the token in the JWT Object itself to reduce moving parts, I am not sure however about whether consolidating the two APIs into one would help a lot. The latter would just be used for SSD Verification anyway.

smtakeda:
@h-chaturvedi , @ChTimTsubasa and @notkriswagner can you guys update the status of this PR? Do we still changes? @h-chaturvedi is this blocking your work?

h-chaturvedi:
Closing this pull request since OCSP SSD Project has been decommissioned.

PR #123, opened 2019-02-08
Visual Studio 2017 Support

notkriswagner:
Updated Windows build scripts to be able to build with Visual Studio 14 (2015) or 15 (2017). Ensures all builds happen with the same VS version. Updated build scripts to work with both Visual Studio 14.0 and 15.0 in both 32 and 64 bit versions. Built dependency libraries for VS14/x86, VS14/x64, VS15/x86, VS15/x64. Reworked appveyor.yml to test a VS15 build.


smtakeda:
hmm, Appveyor didn't pass the tests... Have you identified the cause?

notkriswagner:
Yes this is the cause: ` IP 67.225.164.53 is not allowed to access Snowflake.`

Seems like I have to update the IP addresses again.

notkriswagner:
@smtakeda The IP whitelisting is updated and now all tests are passing

PR #124, opened 2019-02-12
Ssl upgrade

rvishureddy:
Upgrade to openssl 1.1.1a


notkriswagner:
This is great. Can you also add builds for OS X, and Win32/64 in both VS 2015 and VS 2017?

smtakeda:
Built OSX binaries on Sierra and pushed.

I could not build it on High Sierra due to this error:
```
    ld: dynamic main executables must link with libSystem.dylib for architecture i386
```
I didn't dig into the root cause, but sounds like a sign of removing 32 bit support from OSX?

Anyway, this year we can survive as long as Sierra exists. Next year after Sierra is dropped from Apple support, we may want to drop OSX 32 support. Hopes SimbaEngineSDK aligns soon.

notkriswagner:
Looks like appveyor changed the IPs that need to be whitelisted. I'll update them today and re-run the failed job

smtakeda:
@notkriswagner appveyor tests are green. Go ahead to merge and set the tag if you are good.

notkriswagner:
LGTM

PR #128, opened 2019-04-23
Azure put get support

smtakeda:


CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ enable_testing()
 add_definitions(-DLOG_USE_COLOR)
 
 option(BUILD_TESTS "True if build tests" on)
+set(OPENSSL_VERSION_NUMBER  0x11100000L)
```
notkriswagner: Will this cause the build to fail if the version of OpenSSL that a customer is building with doesn't match this version of OpenSSL?

scripts/build_azuresdk.sh
```diff
@@ -0,0 +1,101 @@
...
+    azure_configure_opts+=("-DCMAKE_BUILD_TYPE=Release")
+fi
+azure_configure_opts+=(
+    "-DCMAKE_VERBOSE_MAKEFILE:BOOL=OFF"
+    "-DCMAKE_C_COMPILER=$GCC"
+    "-DCMAKE_CXX_COMPILER=$GXX"
+    "-DCMAKE_INSTALL_PREFIX=$AZURE_BUILD_DIR"
+    "-DBUILD_SHARED_LIBS=OFF"
+    "-DUSE_OPENSSL=true"
+    "-DOPENSSL_VERSION_NUMBER=0x11100000L"
```
smtakeda: @rvishureddy , Better moving this version to the top of this file or config. Related to @notkriswagner question; does this matter to Azure SDK build? I mean without the version number, what will happen? Ideally version neutral build scrip would be great for maintenance.

rvishureddy: Openssl API's changed with version 1.1.1. libSnowflakeclient and ODBC both use openssl 1.1.1b and to have corresponding API's invoked in Azure sdk it should be compiled with the above flag.

smtakeda: Ok, thanks for confirmation.

smtakeda:
LGTM

CMakeLists.txt
rvishureddy: They should not fail. As we statically link and do archive all libs into libSnowflake.so the application should link with the client specific openssl.

lib/client.c
```diff
@@ -456,6 +456,7 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
     log_debug("timezone: %s", sf->timezone);
     log_debug("login_timeout: %d", sf->login_timeout);
     log_debug("network_timeout: %d", sf->network_timeout);
+    log_debug("snowflake_test_ca_bundle_file from getenv: %s", getenv("SNOWFLAKE_TEST_CA_BUNDLE_FILE"));
```
notkriswagner: This looks like a test debugging line, could this be moved out of client.c and into one of the test files maybe?

notkriswagner:
Besides minor issue, LGTM.

smtakeda:
ok, let's merge the change.

PR #129, opened 2019-04-30
Service name

notkriswagner:
Added support for service name as well as completed a refactoring of request headers. A mock test framework was also added to verify that service name works.

### DIFF
--- a/lib/client_int.h
+++ b/lib/client_int.h
@@ -16,7 +16,7 @@
 #define HEADER_ACCEPT_TYPE_APPLICATION_JSON "accept: application/json"
 #define HEADER_C_API_USER_AGENT "User-Agent: c_api/0.1"
 #define HEADER_DIRECT_QUERY_TOKEN_FORMAT "Authorization: %s"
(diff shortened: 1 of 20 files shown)


notkriswagner:
@smtakeda please review when you get a chance

lib/client_int.h
```diff
@@ -16,6 +16,7 @@
 #define HEADER_ACCEPT_TYPE_APPLICATION_JSON "accept: application/json"
 #define HEADER_C_API_USER_AGENT "User-Agent: c_api/0.1"
 #define HEADER_DIRECT_QUERY_TOKEN_FORMAT "Authorization: %s"
+#define HEADER_SERVICE_NAME_FORMAT "Service-Name: %s"
```
smtakeda:
```
X-Snowflake-Service
```
?

smtakeda:
LGTM other than the header name.

lib/client_int.h
notkriswagner: Good catch

PR #136, opened 2019-05-28
Remove generated file Makefile 

rvishureddy:
Remove generated files
distclean zlib before configure.

### DIFF
--- a/deps/zlib-1.2.11/.gitignore
+++ b/deps/zlib-1.2.11/.gitignore
@@ -1,4 +1,4 @@
 zconf.h.included
-cmake-build*a
+cmake-build
 zconf.h
 Makefile
--- a/deps/zlib-1.2.11/.gitignore
+++ b/deps/zlib-1.2.11/.gitignore
@@ -1,2 +1,4 @@
 zconf.h.included
-cmake-build*
+cmake-build*a
+zconf.h
+Makefile
--- a/deps/zlib-1.2.11/Makefile
+++ b/deps/zlib-1.2.11/Makefile
@@ -1,5 +1,0 @@
-all:
-	-@echo "Please use ./configure first.  Thank you."
-
-distclean:
-	make -f Makefile.in distclean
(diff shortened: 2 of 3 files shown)


deps/zlib-1.2.11/.gitignore
```diff
@@ -1,2 +1,4 @@
 zconf.h.included
-cmake-build*
\ No newline at end of file
+cmake-build*a
```
smtakeda: what is *a?

rvishureddy: Looks like a typo. I will fix it.

rvishureddy: Fixed it.

smtakeda:
LGTM. Ensure the tests pass.

deps/zlib-1.2.11/.gitignore
```diff
@@ -1,2 +1,4 @@
 zconf.h.included
-cmake-build*
\ No newline at end of file
+cmake-build
+zconf.h
```
ChTimTsubasa: The zconf.h and Makefile are part of the zlib source and removing it may leads to build failure AFAIK.

rvishureddy:
Makefile zconf.h are generated when configure is run. However zconf.h is
not part of the code base. Makefile had a target for distclean which in
turn invokes Makefile.in for distclean.

deps/zlib-1.2.11/.gitignore
notkriswagner: Removing the Makefile should be fine since the build script now uses Makefile.in instead. I am curious as to why you removed zconf.h though

notkriswagner:
lgtm

deps/zlib-1.2.11/.gitignore
rvishureddy: zconf.h is not part of code base in the first place.

ChTimTsubasa: zconif.h is part of the code base: https://github.com/madler/zlib
Anyway, let's wait for the test result and see. I remembered it failed on Windows once I removed them.

rvishureddy: zconf.h.cmakein      zconf.h template for cmake
zconf.h.in                 zconf.h template for configure
Yeah let's wait for the results.

PR #138, opened 2019-05-31
Session gone

notkriswagner:
Added logic to catch session gone error and ignore if this occurs during session termination. Otherwise the error is returned to the user.


tests/mock/test_mock_session_gone.c
```diff
@@ -0,0 +1,96 @@
+/*
+ * Copyright (c) 2017-2019 Snowflake Computing, Inc. All rights reserved.
+ */
+
+#include <string.h>
+#include "../utils/test_setup.h"
+#include "../utils/mock_setup.h"
+
+void test_session_gone_during_query(void **unused) {
```
smtakeda: Don't ignore the error while running query.

tests/mock/test_mock_session_gone.c
```diff
@@ -0,0 +1,96 @@
...
+        ++counter;
+    }
+    if (status != SF_STATUS_EOF) {
+        dump_error(&(sfstmt->error));
+    }
+    assert_int_equal(status, SF_STATUS_EOF);
+    snowflake_stmt_term(sfstmt);
+
+    setup_mock_delete_connection_session_gone();
+    status = snowflake_term(sf);
```
smtakeda: This is the only case we want to ignore the session gone error.

tests/mock/test_mock_session_gone.c
notkriswagner: Yea I don't ignore the error while running a query. The purpose of this test is to ensure that if this occurs when we run a query, then we return an error to the user. See below for when we check for an error.

tests/mock/test_mock_session_gone.c
```diff
@@ -0,0 +1,96 @@
...
+    assert_int_equal(status, SF_STATUS_SUCCESS);
+
+    /* query */
+    setup_mock_query_session_gone();
+    SF_STMT *sfstmt = snowflake_stmt(sf);
+    status = snowflake_query(sfstmt, "select 1;", 0);
+    if (status != SF_STATUS_ERROR_CONNECTION_NOT_EXIST) {
+        dump_error(&(sfstmt->error));
+    }
+    assert_int_equal(status, SF_STATUS_ERROR_CONNECTION_NOT_EXIST);
```
notkriswagner: This is where we check to make sure that an error is returned to the user if we get a session gone error code.

tests/mock/test_mock_session_gone.c
smtakeda: oh, sorry about that. I took wrong.

PR #139, opened 2019-06-05
Platform-Specific macOS Binaries

notkriswagner:


notkriswagner:
Not sure why this shows up as being by two users. Probably something misconfigured on my macOS VM

CMakeLists.txt
```diff
@@ -245,14 +244,26 @@ if (UNIX)
...
+
 if (WIN32)
     include_directories(
             deps-build/${PLATFORM}/${VSDIR}/curl/include
             deps-build/${PLATFORM}/${VSDIR}/openssl/include
             deps-build/${PLATFORM}/${VSDIR}/zlib/include
             deps-build/${PLATFORM}/${VSDIR}/aws/include
             deps-build/${PLATFORM}/${VSDIR}/cmocka/include
-	    deps-build/${PLATFORM}/${VSDIR}/azure/include
+	        deps-build/${PLATFORM}/${VSDIR}/azure/include
```
notkriswagner: This has been reverted

scripts/build_azuresdk.sh
```diff
@@ -68,18 +68,36 @@ azure_configure_opts+=(
...
     "-DOPENSSL_CRYPTO_LIBRARY=$DEPENDENCY_DIR/openssl/lib/libcrypto.a"
     "-DOPENSSL_SSL_LIBRARY=$DEPENDENCY_DIR/openssl/lib/libssl.a"
-    "-DUUID_INCLUDE_DIR=$DEPENDENCY_DIR/uuid/include"
-    "-DUUID_LIBRARIES=$DEPENDENCY_DIR/uuid/lib/libuuid.a"
     "-DCURL_INCLUDE_DIRS=$DEPENDENCY_DIR/curl/include"
     "-DCURL_LIBRARIES=$DEPENDENCY_DIR/curl/lib/libcurl.a"
     "-DEXTRA_INCLUDE=$DEPENDENCY_DIR/zlib/include"
 )
 
+if [[ "$PLATFORM" == "linux" ]]; then
```
notkriswagner: @smtakeda macOS has libuuid by default and including our prebuilt libuuid causes a build error in the Azure Storage Lite SDK. The SDK also recommends that you use the system libuuid for macOS, so I dropped this from macOS.

scripts/build_openssl.sh
```diff
@@ -27,6 +27,7 @@ openssl_config_opts=()
 openssl_config_opts+=(
     "no-shared"
     "--prefix=$OPENSSL_BUILD_DIR"
+    "--openssldir=$OPENSSL_BUILD_DIR"
```
notkriswagner: The OpenSSL build page recommends that we set both of these parameters for the version of OpenSSL that we are using (not doing so can lead to build inconsistencies).

scripts/build_azuresdk.sh
rvishureddy: Just like windows. The unreferenced error for gettext can be fixed by removing path to the gettext package on macos. by running  
brew unlink gettext
This will mark gettext as unavailable and we do not see the unreferenced error.
If libuuid is available on macos that should be fine.

smtakeda: +1

scripts/build_openssl.sh
```diff
@@ -42,19 +43,37 @@ if [[ "$PLATFORM" == "linux" ]]; then
...
-    make -j 4 > /dev/null
-    make install_sw install_ssldirs > /dev/null
-    make distclean clean &> /dev/null || true
-    ./Configure darwin64-x86_64-cc "${openssl_config_opts[@]}"
-    make -j 4 > /dev/null
-    lipo -create $OPENSSL_BUILD_DIR/lib/libssl.a    ./libssl.a    -output $OPENSSL_BUILD_DIR/lib/../libssl.a
-    lipo -create $OPENSSL_BUILD_DIR/lib/libcrypto.a ./libcrypto.a -output $OPENSSL_BUILD_DIR/lib/../libcrypto.a
-    mv $OPENSSL_BUILD_DIR/lib/../libssl.a    $OPENSSL_BUILD_DIR/lib/libssl.a
-    mv $OPENSSL_BUILD_DIR/lib/../libcrypto.a $OPENSSL_BUILD_DIR/lib/libcrypto.a
+    openssl_config_opts+=("-mmacosx-version-min=10.11")
```
smtakeda: 10.12?

scripts/build_cmocka.sh
```diff
@@ -26,8 +26,22 @@ config_opts=(
...
+	    config_opts+=("-DCMAKE_OSX_ARCHITECTURES=x86_64;i386")
+	else
+	    if [[ "$ARCH" == "x86" ]]; then
+	        echo "[INFO] Building x86 Binary"
+	        config_opts+=("-DCMAKE_OSX_ARCHITECTURES=i386")
+	    else
+	        echo "[INFO] Building x64 Binary"
+	        config_opts+=("-DCMAKE_OSX_ARCHITECTURES=x86_64")
+	    fi
+    fi
```
smtakeda: We really want to build ODBC with cmake. @ChTimTsubasa see the above change.

smtakeda:
@notkriswagner Is this still open? Can we commit and close?

PR #140, opened 2019-06-07
Append UUID string to the tmp dir to create unique dir

rvishureddy:
Create a uuid string append to the temp dir to create a unique temp directory.

### DIFF
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -300,9 +300,8 @@ void Snowflake::Client::FileTransferAgent::compressSourceFile(
 {
   CXX_LOG_DEBUG("Starting file compression");
   
-  char tempDir[150]={0};
-  memset(tempDir, 0, 150);
-  sf_get_tmp_dir(tempDir);
+  char tempDir[MAX_PATH]={0};
+  sf_get_uniq_tmp_dir(tempDir);
   std::string stagingFile(tempDir);
   stagingFile += fileMetadata->destFileName;
 
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -254,7 +254,8 @@ RemoteStorageRequestOutcome Snowflake::Client::FileTransferAgent::uploadSingleFi
 
   if (fileMetadata->requireCompress)
   {
-    remove(fileMetadata->srcFileToUpload.c_str());
+    //Remove the uniq temp directory created (force remove).
+    sf_delete_uniq_dir_if_exists(fileMetadata->srcFileToUpload.c_str());
   }
 
   m_executionResults->SetTransferOutCome(outcome, resultIndex);
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -303,7 +303,6 @@ void Snowflake::Client::FileTransferAgent::compressSourceFile(
   char tempDir[150]={0};
   memset(tempDir, 0, 150);
   sf_get_tmp_dir(tempDir);
-  sf_create_directory_if_not_exists(tempDir);
   std::string stagingFile(tempDir);
   stagingFile += fileMetadata->destFileName;
 
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -300,8 +300,10 @@ void Snowflake::Client::FileTransferAgent::compressSourceFile(
 {
   CXX_LOG_DEBUG("Starting file compression");
   
-  char tempDir[100];
+  char tempDir[150]={0};
+  memset(tempDir, 0, 150);
   sf_get_tmp_dir(tempDir);
+  sf_create_directory_if_not_exists(tempDir);
   std::string stagingFile(tempDir);
   stagingFile += fileMetadata->destFileName;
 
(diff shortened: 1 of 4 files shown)


cpp/FileTransferAgent.cpp
```diff
@@ -300,7 +300,8 @@ void Snowflake::Client::FileTransferAgent::compressSourceFile(
 {
   CXX_LOG_DEBUG("Starting file compression");
   
-  char tempDir[100];
+  char tempDir[150]={0};
```
smtakeda: can this be larger? 150 is too small for full path. could this be 4096?

lib/platform.c
```diff
@@ -543,20 +546,24 @@ int STDCALL sf_delete_directory_if_exists(const char * directoryName)
...
-  }
+    sprintf(tmpDir,"/tmp/%s/",uuid_cstr);
   else
-  {
-    strncpy(tmpDir, tmpEnv, strlen(tmpEnv));
-    size_t oldLen = strlen(tmpDir);
-    tmpDir[oldLen] = PATH_SEP;
-    tmpDir[oldLen+1] = '\0';
-  }
+    sprintf(tmpDir, "%s/%s/",tmpEnv,uuid_cstr);
```
smtakeda: add curly braces even for online statement.

lib/platform.c
```diff
@@ -543,20 +546,24 @@ int STDCALL sf_delete_directory_if_exists(const char * directoryName)
...
-
+  uuid_t uuid;
+  char uuid_cstr[37]; // 36 byte uuid plus null.
+  uuid_generate(uuid);
+  uuid_unparse(uuid, uuid_cstr);
   if (!tmpEnv)
-  {
-    strncpy(tmpDir, "/tmp/", sizeof("/tmp/"));
-  }
+    sprintf(tmpDir,"/tmp/%s/",uuid_cstr);
```
smtakeda: add curly braces.

lib/platform.c
```diff
@@ -543,20 +546,24 @@ int STDCALL sf_delete_directory_if_exists(const char * directoryName)
...
-    strncpy(tmpDir, tmpEnv, strlen(tmpEnv));
-    size_t oldLen = strlen(tmpDir);
-    tmpDir[oldLen] = PATH_SEP;
-    tmpDir[oldLen+1] = '\0';
-  }
+    sprintf(tmpDir, "%s/%s/",tmpEnv,uuid_cstr);
 #endif
-}
\ No newline at end of file
+  sf_create_directory_if_not_exists(tmpDir);
```
smtakeda: If the directory exists, it is very likely the tmp directory is already used by other PUT commands?

cpp/FileTransferAgent.cpp
rvishureddy: Was adding extra 37 bytes of UUID so thought +50 should do. But yeah customers can have long path. Will fix it.

lib/platform.c
rvishureddy: This is going to be a unique temp directory for every put command. Every time the customer tries to put file we create a unique temp directory. which looks like
/tmp/<uuid string>/

cpp/FileTransferAgent.cpp
```diff
@@ -300,7 +300,8 @@ void Snowflake::Client::FileTransferAgent::compressSourceFile(
 {
   CXX_LOG_DEBUG("Starting file compression");
   
-  char tempDir[100];
+  char tempDir[4096]={0};
+  memset(tempDir, 0, 4096);
```
ChTimTsubasa: The previous line should already set the array to 0 and there should be no need to memset again.

lib/platform.c
```diff
@@ -542,21 +543,22 @@ int STDCALL sf_delete_directory_if_exists(const char * directoryName)
 
 void STDCALL sf_get_tmp_dir(char * tmpDir)
```
ChTimTsubasa: Can you add comments to the function declaration informing user the minimum memory required for tmpDir?

cpp/FileTransferAgent.cpp
rvishureddy: True. Removed the later.

lib/platform.c
rvishureddy: Left the original funtion as is and created a new function sf_get_uniq_tmp_dir which has comments.

rvishureddy:
> One suggestion is to have a common dir prefix under tmp instead of directly using uuid under tmp. For example it could be /tmp/snowflake_local_stage/{env}/{uuid}. That way it could be easier if we want to clean up later.
Its already temp dir it does not give any advantage of having one more local directory. Even if you have a local_stage you cannot remove with a wild char as there might be some folders being in use. And then can always be removed by sorting by date.

notkriswagner:
> > One suggestion is to have a common dir prefix under tmp instead of directly using uuid under tmp. For example it could be /tmp/snowflake_local_stage/{env}/{uuid}. That way it could be easier if we want to clean up later.
> > Its already temp dir it does not give any advantage of having one more local directory. Even if you have a local_stage you cannot remove with a wild char as there might be some folders being in use. And then can always be removed by sorting by date.

One advantage of the common dir prefix is that if a customer looks at their /tmp dir then they would know that these temp dirs are created by libsnowflakeclient instead of wondering what the random UUIDs mean

PR #141, opened 2019-06-18
NULL input support

notkriswagner:
Added support for binding NULL input parameters when you don't know that the C type of that column is.


lib/results.c
```diff
@@ -93,6 +95,8 @@ const char *STDCALL snowflake_type_to_string(SF_DB_TYPE type) {
             return "TIME";
         case SF_DB_TYPE_BOOLEAN:
             return "BOOLEAN";
+        case SF_DB_TYPE_ANY:
+            return "ANY";
```
smtakeda: I assume this will be set to the server and accepted?

notkriswagner: Yes this is sent to the server (and accepted) when the user chooses SF_C_TYPE_NULL as their input binding

smtakeda: ok

PR #144, opened 2019-06-21
Set up CI with Azure Pipelines

rvishureddy:
Trying azure pipelines


smtakeda:
@rvishureddy good start. You may want to disable Appveyor as Azure Pipeline should cover windows too.

PR #145, opened 2019-07-09
snow-85722 Perform check for thread index allocation and free index

ChTimTsubasa:
https://snowflakecomputing.atlassian.net/browse/SNOW-85722

### DIFF
--- a/cpp/util/ThreadPool.hpp
+++ b/cpp/util/ThreadPool.hpp
@@ -18,6 +18,7 @@
 #include <cstring>
 #include "snowflake/platform.h"
 #include "ByteArrayStreamBuf.hpp"
+#include "../logger/SFLogger.hpp"
 
 namespace Snowflake
 {
@@ -143,6 +144,12 @@ public:
 
 #ifdef _WIN32
     key = TlsAlloc();
+    if (key == TLS_OUT_OF_INDEXES)
+    {
+      CXX_LOG_ERROR("Thread pool out of TLS index");
+      throw SnowflakeTransferException(TransferError::INTERNAL_ERROR,
+                                               "Out of TLS index in the thread pool");
+    }
 #else
     pthread_key_create(&key, NULL);
 #endif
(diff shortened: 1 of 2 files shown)


cpp/util/ThreadPool.hpp
```diff
@@ -143,6 +145,12 @@ class ThreadPool {
...
 #ifdef _WIN32
     key = TlsAlloc();
+    if (key == TLS_OUT_OF_INDEXES)
+    {
+      CXX_LOG_ERROR("Thread pool out of TLS index");
+      throw SnowflakeTransferException(TransferError::INTERNAL_ERROR,
+                                               "Out of TLS index in the thread pool");
+    }
 #else
     pthread_key_create(&key, NULL);
```
rvishureddy: Can you also do the check for Linux platform. pthread_key_create fails when the system has insufficient memory or if max keys is reached.

tests/test_connect_negative.c
```diff
@@ -121,9 +121,9 @@ int main(void) {
       cmocka_unit_test(test_connect_account_missing),
       cmocka_unit_test(test_connect_user_missing),
       cmocka_unit_test(test_connect_password_missing),
-      cmocka_unit_test(test_connect_invalid_database),
-      cmocka_unit_test(test_connect_invalid_schema),
-      cmocka_unit_test(test_connect_invalid_warehouse),
+      //cmocka_unit_test(test_connect_invalid_database),
```
notkriswagner: Can you create a JIRA to re-enable these tests and assign it to me for the next sprint?

tests/test_connect_negative.c
```diff
@@ -121,9 +121,9 @@ int main(void) {
...
+      //cmocka_unit_test(test_connect_invalid_schema),
+      //cmocka_unit_test(test_connect_invalid_warehouse),
       cmocka_unit_test(test_connect_invalid_role),
```
rvishureddy: Ok to comment out these tests created SNOW-85815 to track.

cpp/util/ThreadPool.hpp
ChTimTsubasa: Good point. I will add it.

tests/test_connect_negative.c
ChTimTsubasa: @rvishureddy created it and I CCed you on SNOW-85815

howryu:
So what do we do if tls alloc failed? We just throw exception to client directly?

PR #149, opened 2019-08-30
SNOW-81418: Avoid Get before put and overwrite the file to prevent S3…

rvishureddy:
SNOW-81418: Avoid Get before put and overwrite the file to prevent S3…


smtakeda:
@rvishureddy I just want to remind you that OVERWRITE mode is optional.
It is applied only if PUT command has the OVERWRITE=true option.
By default, we still do check the file existence before transferring the file.

PR #154, opened 2019-09-27
SNOW-99277: remove openssl-1.1.0g

smtakeda:
more clean up


smtakeda:
@notkriswagner and @rvishureddy can we merge this?

PR #158, opened 2019-10-17
Treat all 0-scale fixed columns as int64

notkriswagner:


lib/results.c
```diff
@@ -132,7 +132,7 @@ const char * STDCALL snowflake_c_type_to_string(SF_C_TYPE type) {
 
 SF_C_TYPE snowflake_to_c_type(SF_DB_TYPE type, int64 precision, int64 scale) {
     if (type == SF_DB_TYPE_FIXED) {
-        if (scale > 0 || precision >= 19) {
```
notkriswagner: This is to fix an issue where the return type from libsnowflakeclient is a float64 for hash function columns when it should really be int64. If the fixed column is too large to fit inside an int, we will return an error code when the user tries to fetch that value

PR #160, opened 2019-10-21
Fixes Azure SDK large file upload

rvishureddy:
Most of the modified code is from Azure SDK. 

And tests/test_simple_put.cpp is modified to enable large file uploads, copying into table and copying from table to staging file.


rvishureddy:
Large file test for linux is done on ODBC/Jenkins.

rvishureddy:
Fixes: large file issue on mac and windows.
Fixes: crash in azure sdk.

Issue #165, opened 2019-12-04
Inconsistent behavior for put command on AWS when overwrite=false


SimbaGithub:
According to the document, https://docs.snowflake.net/manuals/sql-reference/sql/put.html
when overwrite=false, the uploading should be skipped if there is an existing file with the same name.
But on AWS, the uploading is skipped only when both file name and checksum are the same. It also could cause crash if the server doesn't return checksum of the existing file. On Azure, only file name is checked.

We need consistent behavior across the cloud providers, change the behavior on AWS so that only file name is checked.

rvishureddy:
Are you tracking the crash as part of this JIRA.

SimbaGithub:
> Are you tracking the crash as part of this issue.

Yup, add it into the issue description.

PR #166, opened 2019-12-05
Simba Changes

SimbaGithub:
1. Make consistent behavior for put command with overwrite=off
https://github.com/snowflakedb/libsnowflakeclient/issues/165
2. Fix security issue in curl/lib/vtls/sf_ocsp.c
http://cwe.mitre.org/data/definitions/134.html

### DIFF
--- a/cpp/SnowflakeS3Client.cpp
+++ b/cpp/SnowflakeS3Client.cpp
@@ -136,13 +136,7 @@ RemoteStorageRequestOutcome SnowflakeS3Client::upload(FileMetadata *fileMetadata
               s3Client->HeadObject(headObjectRequest);
 
       if (outcome.IsSuccess()) {
-          std::string sfcDigest = outcome.GetResult().GetMetadata().at(
-                  SFC_DIGEST);
-          if (sfcDigest == fileMetadata->sha256Digest) {
-              CXX_LOG_INFO("File %s with same name and sha256 existed. Skipped.",
-                           fileMetadata->srcFileToUpload.c_str());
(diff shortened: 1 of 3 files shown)


cpp/SnowflakeS3Client.cpp
```diff
@@ -136,13 +136,7 @@ RemoteStorageRequestOutcome SnowflakeS3Client::upload(FileMetadata *fileMetadata
               s3Client->HeadObject(headObjectRequest);
 
       if (outcome.IsSuccess()) {
-          std::string sfcDigest = outcome.GetResult().GetMetadata().at(
```
rvishureddy: This also fixes the crash when metadata does not have SFC_DIGEST. 
"Further more, I found in following case the server doesn’t return “sfc-digest” in the result which crashes the driver.

 

create temporary stage sen.harrystagetest

copy into @sen.harrystagetest from sen.varchar_table file_format=(FIELD_OPTIONALLY_ENCLOSED_BY='"')

get @sen.harrystagetest file://c:\snowflake\get

put file://c:\snowflake\get\data_0_0_0.csv.gz @sen.harrystagetest OVERWRITE = false

 

The checksum should be the same because the upload file is downloaded by get command. If the existing file is generated by copy into command, the server doesn’t return “sfc-digest” and the driver doesn’t check if “sfc-digest” exists before get the value. That’s why the driver crashes in such case.

Is it a server issure not returning “sfc-digest”? I’ll fix the driver crash by checking existence of “sfc-digest” before get the value."

PR #173, opened 2020-02-18
SNOW-137586 delete curl-7.58 from libsnowflakeclient

snowstakeda:
I believe we no longer need curl-7.58


snowstakeda:
This is not ready. PR is created to trigger tests.

snowstakeda:
nevemind the previous deleted comment.

PR #177, opened 2020-02-25
Snow 137587 add libsnowflake mac new ci

snowstakeda:
@ankitbhatnagarsf , this is not true PR but just to share how we are making changes for scripts.


scripts/build_zlib.sh
```diff
@@ -6,17 +6,21 @@ function usage() {
     echo "Usage: `basename $0` [-t <Release|Debug>]"
     echo "Builds zlib"
     echo "-t <Release/Debug> : Release or Debug builds"
+    echo "-v: return version"
```
snowstakeda: add `-v` option to return the version of the component. Each build_NNNN.sh should implement this.

scripts/build_zlib.sh
```diff
@@ -6,17 +6,21 @@ function usage() {
...
 set -o pipefail
 
+ZLIB_VERSION=1.2.11
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" && pwd )"
-SOURCE_DIR=$DIR/../deps/zlib-1.2.11
+SOURCE_DIR=$DIR/../deps/zlib-${ZLIB_VERSION}
 source $DIR/_init.sh $@
 
+[[ -n "$GET_VERSION" ]] && echo $ZLIB_VERSION && exit 0
```
snowstakeda: If `-v` is specified, return the version number. This is useful to create zip file.

scripts/_init.sh
```diff
@@ -84,9 +86,11 @@ done
...
-echo "  target       = $target"
-echo "PATH="$PATH
+if [[ -z "$GET_VERSION" ]]; then
+    echo "Options:"
+    echo "  target       = $target"
+    echo "PATH="$PATH
+fi
 
-DEPENDENCY_DIR=$DIR/../deps-build/$PLATFORM
+export DEPENDENCY_DIR=$DIR/../deps-build/$PLATFORM/$target
```
snowstakeda: Added an extra level of path to have both Debug and Release.
In case of Mac, that's the only variant aside from the platform name.

PR #190, opened 2020-03-25
SNOW-100200 use -fPIC for zlib

snowstakeda:
It appeared the generated zlib library file is not compatible for the Snowflake ODBC driver, because it was not built with fPIC option.


scripts/build_zlib.sh
```diff
@@ -36,6 +36,7 @@ cd $SOURCE_DIR
 if [[ "$PLATFORM" == "linux" ]]; then
     # Linux 64 bit
     export CC=gcc52
+    export CFLAGS="-fPIC"
```
snowstakeda: This is the main change.

Others are just bumping up the version

PR #199, opened 2020-04-30
Enable put/get tests on AWS and AZURE

rvishureddy:
Currently enabled put get tests on AWS and AZURE

PR #203, opened 2020-05-20
putget on gcp for libsnowflakeclient

SimbaGithub:
Support put/get on gcp in libsnowflakeclient.
Merged test cases from branch SNOW-156394_run_put_get_tests_on_all_cloud_env and enabled for GCP.
test_simple_put_threshold and test_large_put_threshold are failing on my end with error of "Unsupported feature", not sure that's test environment or test case issue so I keep them for now. Will remove them (as what has been done in SNOW-156394_run_put_get_tests_on_all_cloud_env) if also failed on Snowflake's test environment.


tests/test_simple_put.cpp
```diff
@@ -495,7 +505,17 @@ void test_simple_put_skip(void **unused)
...
-  ITransferResult * results = agent.execute(&putCommand);
+
+  ITransferResult * results = NULL;
+  try
+  {
+    results = agent.execute(&putCommand);
+  }
+  catch (...)
+  {
+    assert_string_equal("", sfstmt->error.msg);
```
smtakeda, line 516: so why no error message even if an exception is raised here?

tests/test_simple_put.cpp
```diff
@@ -427,10 +453,12 @@ void test_large_get(void **unused)
...
   char tempPath[MAX_PATH + 256] = "get @%test_small_put/bigFile.csv.gz file://";
-    if ( ! strncmp(getenv("SNOWFLAKE_CLOUD_ENV"), "AWS", 6) ) {
-        errno = 0;
-        return;
-    }
+  const char *cloud_provider = std::getenv("CLOUD_PROVIDER");
+  if((!cloud_provider) ||
+     ((strcmp(cloud_provider, "GCP") != 0) && (strcmp(cloud_provider, "AWS") != 0))) {
+      errno = 0;
+      return;
```
smtakeda, line 460: Does this mean no AZURE support? @rvishureddy please confirm.

tests/test_simple_put.cpp
```diff
@@ -81,15 +81,24 @@ void test_simple_put_core(const char * fileName,
...
+  ITransferResult * results = NULL;
+  //catch exception to continue with other test cases if query fails.
+  try
+  {
+    results = agent.execute(&putCommand);
+    assert_int_equal(1, results->getResultSize());
+  }
+  catch (...)
+  {
+    assert_string_equal("", sfstmt->error.msg);
```
smtakeda, line 100: so if an error is raised, should not it be non-empty string?

tests/CMakeLists.txt
```diff
@@ -83,18 +80,8 @@ set(SOURCE_UTILS_CXX
...
-
-# Only run PUT/GET tests on Azure deployments
-if ("$ENV{SNOWFLAKE_CLOUD_ENV}" STREQUAL "AZURE")
-    message("SNOWFLAKE_CLOUD_ENV is set to " $ENV{SNOWFLAKE_CLOUD_ENV})
-    message("libazure-storage-lite.a is located at " ${AZURE_STORAGE_LITE_LIB})
-    set(TESTS_CXX ${TESTS_CXX})
-    set(TESTS_CXX ${TESTS_CXX} ${TESTS_AZURE})
-endif()
+message("CLOUD_PROVIDER is set to " $ENV{CLOUD_PROVIDER})
+set(TESTS_CXX ${TESTS_CXX} ${TESTS_PUTGET})
```
smtakeda, line 84: This is good. An idea is build a test executable that can run against AWS, Azure and GCP. Ensure including both AWS and Azure SDK.

tests/test_simple_put.cpp
SimbaGithub: Without this change, if the query execution failed, exception is raised and terminate the process so following test cases won't be run.
This change is to catch the exception, since the error message is not empty when query fails, the assertion will fail this test case with error message out in the test output, and continue with following test cases.

sfc-gh-stakeda: Sorry, I'm confused. I understand the statement, "Without this change, if the query execution failed, exception is raised and terminate the process so following test cases won't be run."

But why this assert ensures the error message is an empty string? Should not the error context have some message instead of stdout?

https://api.cmocka.org/group__cmocka__asserts.html#ga3db7885581668648088fc4eb4cee6d4f

SimbaGithub:
Hi @smtakeda @rvishureddy ,
So far Windows looks good but test fails on Linux. Could you please help me to take a look on the test failures on Linux? 
https://github.com/snowflakedb/libsnowflakeclient/pull/203/checks?check_run_id=697322445
I can't reproduce those failures on my end. 

Thanks,
Harry

SimbaGithub:
Hi @rvishureddy ,

It seems the failures on Linux are caused by test environment. I've removed all my changes and only have the changes from the PR you made https://github.com/snowflakedb/libsnowflakeclient/pull/199 but the failures are still there.
The test case of put/get passed before in the checking of your PR and now it fails, so I guess there are some environment changes causing those failures.
Please help me to investigate on the possible test environment issue when you have time. This PR is not needed for June release so it's not urgent.
I'll leave this PR for now until the test environment issue gets fixed.

Thanks,
Harry

sfc-gh-stakeda:
@sfc-gh-vreddy do we need to keep this opened? If not, let's close this.

sfc-gh-vreddy:
@smtakeda 
In my recent PR made changes to run put/get tests on azure and aws with single binary. 
And want to add GCP to complete this. 
Will definitely take a look and merge this and corresponding PR in snowfake-odbc soon.

sfc-gh-stakeda:
Thanks @sfc-gh-vreddy . Btw, my account name is @sfc-gh-stakeda .

sfc-gh-vreddy:
The changes are already in. This PR does not have any Tests related to GCP. So closing this.

PR #207, opened 2020-06-17
SNOW-165820 throw error when upload fails 

sfc-gh-vreddy:
Throws error when Put fails
Adds timestamps for put speeds


cpp/FileTransferAgent.cpp
```diff
@@ -222,6 +223,11 @@ void Snowflake::Client::FileTransferAgent::upload(string *command)
         renewToken(command);
         i--;
       }
+      else if( outcome == RemoteStorageRequestOutcome::FAILED)
+      {
+        throw SnowflakeTransferException(TransferError::INTERNAL_ERROR,
```
sfc-gh-sshankar, line 228: Can you improve this message ? What is the file we failed to upload, and why did it fail ?

sfc-gh-sshankar: And what about reporting errors and checking for outcomes on small files ?

cpp/FileTransferAgent.cpp
```diff
@@ -251,16 +257,23 @@ void Snowflake::Client::FileTransferAgent::uploadFilesInParallel(std::string *co
...
+            CXX_LOG_DEBUG("Token expired, Renewing token.")
             _mutex_lock(&m_parallelTokRenewMutex);
             this->renewToken(command);
             _mutex_unlock(&m_parallelTokRenewMutex);
           }
           else
           {
+            if( outcome == RemoteStorageRequestOutcome::FAILED)
+            {
+              throw SnowflakeTransferException(TransferError::INTERNAL_ERROR,
```
sfc-gh-sshankar, line 274: Again, add more information to the user-visible error message

cpp/SnowflakeAzureClient.cpp
```diff
@@ -49,14 +49,16 @@ SnowflakeAzureClient::SnowflakeAzureClient(StageInfo *stageInfo,
       snowflake_global_get_attribute(SF_GLOBAL_CA_BUNDLE_FILE, caBundleFile, sizeof(caBundleFile));
       CXX_LOG_TRACE("ca bundle file from SF_GLOBAL_CA_BUNDLE_FILE *%s*", caBundleFile);
   }
-if( caBundleFile[0] == 0 ) {
+  else if( caBundleFile[0] == 0 ) {
```
sfc-gh-sshankar, line 52: Why the else if, don't you always want to check this.

sfc-gh-vreddy: Thats true, The else in line 48 also needs to be removed.

cpp/FileTransferAgent.cpp
sfc-gh-vreddy: @sfc-gh-sshankar For small files its done on line 272.
We did add at multiple locations the reason for error. 
Will add more detail to the exception.

cpp/FileTransferAgent.cpp
sfc-gh-vreddy: Yes, will add the reason for failure and the file name.

sfc-gh-sshankar:
How are we testing whether an error is thrown when a put fails ?

sfc-gh-stakeda:
Would you add tests in libsnowflakeclient?

sfc-gh-vreddy:
@smtakeda Added integration tests here to validate different FILE URIS.

sfc-gh-vreddy:
These tests fail on docker. Looking into why its failing in docker.

PR #211, opened 2020-06-23
SNOW-170115 ODBC put on windows fails when file paths have forward slash

sfc-gh-vreddy:
On windows file path C:/Users/vreddy/TestData/testput.csv fails to upload.

This fix will address this issue.

### DIFF
--- a/tests/test_unit_file_metadata_init.cpp
+++ b/tests/test_unit_file_metadata_init.cpp
@@ -70,30 +70,30 @@ std::vector<std::string> getListOfTestFileMatchDir()
 void test_file_pattern_match_core(std::vector<std::string> *expectedFiles,
                                   const char *filePattern)
 {
-	std::vector<std::string> listTestDir = getListOfTestFileMatchDir();
-	for (auto testDir : listTestDir)
-	{
-		std::vector<FileMetadata> smallFileMetadata;
-		std::vector<FileMetadata> largeFileMetadata;
+  std::vector<std::string> listTestDir = getListOfTestFileMatchDir();
+  for (auto testDir : listTestDir)
+  {
+    std::vector<FileMetadata> smallFileMetadata;
+    std::vector<FileMetadata> largeFileMetadata;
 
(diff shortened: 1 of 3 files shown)


sfc-gh-stakeda:
you may want to add test cases?

sfc-gh-vreddy:
@smtakeda Added test cases for different file types and URIs in snowflake-odbc

sfc-gh-kwagner:
@sfc-gh-vreddy can you add a unit test for this change in libsnowflakeclient in this file: https://github.com/snowflakedb/libsnowflakeclient/blob/master/tests/test_unit_file_metadata_init.cpp

sfc-gh-stakeda:
@sfc-gh-vreddy ,the best practice is to have test cases next to the the code so that we can catch any error or behavior change as soon as we make the code change. If we have tests in ODBC, then it will be delayed to the stage. Please add the test cases to libsnowflakeclient. Thanks.

tests/test_unit_file_metadata_init.cpp
```diff
@@ -28,33 +28,75 @@ std::string getTestFileMatchDir()
...
+std::vector<std::string> getListOfTestFileMatchDir()
+{
+  std::vector<std::string> listOfTestDirs; 
+  #ifdef _WIN32
+  std::string srcLocation = getTestFileMatchDir();
+  replaceStrAll(srcLocation,"\\","/");
+  // Directory path looks like C:/Users/vreddy/App/Temp/adkf-kasdfk-adsfl/
+  // Testing forward slash for put/get support
+  listOfTestDirs.push_back(srcLocation);
+  #endif
```
sfc-gh-stakeda, line 63: Looks like you are deleting the test case including backslashes? Don't we want to test both cases on Windows?

sfc-gh-vreddy: @smtakeda 
The above code pushes new directory to the vector of directories. And line 64,65 is the old directory pushed to vector. We did not delete we are adding all the directories to the vector and running tests on all. 
Line 74 goes through all directories and runs tests.

PR #212, opened 2020-06-23
SNOW-163154 Fixes auto compression guess

sfc-gh-vreddy:
When a file with no extension is uploaded, the driver tries to guess the file type compression and fails. 

This fixes the type and returns unknown so we handle it as a normal file.


sfc-gh-stakeda:
I think this should have unit test case to cover all filename cases.

sfc-gh-vreddy:
@smtakeda Unit Tests are added in snowflake-odbc

sfc-gh-kwagner:
@sfc-gh-vreddy you need to update this libsnowflakeclient unit test
https://github.com/snowflakedb/libsnowflakeclient/blob/master/tests/test_unit_file_type_detect.cpp

PR #213, opened 2020-06-26
SNOW 162610: Do not use thread pool when parallel=1

sfc-gh-vreddy:
When using applications with no thread support to upload many small files in parallel. The threadpool gets initiated and as the threads are not supported this adds delay of 2 seconds for every upload file. 
Given customer had 500 files it adds around 1000seconds of delay. 

To reduce that delay, the thread pool is not used when parallel is set to 1. 

And we should recommend customers using applications with no thread support to set parallel=1 to the query to reduce delay.


cpp/FileTransferAgent.cpp
```diff
@@ -248,6 +248,30 @@ void Snowflake::Client::FileTransferAgent::uploadFilesInParallel(std::string *co
     FileMetadata * metadata = &m_smallFilesMeta[i];
     m_executionResults->SetFileMetadata(&m_smallFilesMeta[i], resultIndex);
     metadata->overWrite = response.overwrite;
+
+    // workaround for incident 00212627
+    // Do not use thread pool when parallel = 1
+    if ((unsigned int)response.parallel <= 1)
```
sfc-gh-kwagner, line 254: Can response.parallel ever be less than 1?

sfc-gh-vreddy: @sfc-gh-kwagner The approved values are 1 to 99 with 1 no parallelism and 99 highest level and default 4.

PR #214, opened 2020-07-01
SNOW-169969 Add DSN and Connection Parameters to ODBC OOB

sfc-gh-jbahk:
DSN Test result: 

```
[{
		"Created_On":	"2020-07-13 20:47:44",
		"Name":	"ConnectionDSNConfig",
		"SchemaVersion":	1,
		"Tags":	{
			"UUID":	"69d11590-3401-4228-a61c-7583b81ed688",
			"connectionString":	"snowflake.local.snowflakecomputing.com:443/session/v1/login-request?requestId=c4d53986-ee7a-4f01-9fac-2604653e9c41&request_guid=abbab0e5-5c77-4102-9d29-d44efde6a050&databaseName=testdb&schemaName=testschema&warehouse=regress",
			"ctx_account":	"sfctest0",
			"ctx_port":	"443",
			"ctx_protocol":	"https",
			"ctx_user":	"snowman",
			"driver":	"ODBC",
			"version":	"99.0.0",
			"hostOs":	"Linux",
			"telemetryServerDeployment":	"dev"
		},
		"Type":	"Log",
		"UUID":	"69d11590-3401-4228-a61c-7583b81ed688",
		"Urgent":	false,
		"Value":	{
			"DSN":	{
				"SERVER":	"snowflake.local.snowflakecomputing.com",
				"PORT":	"443",
				"ACCOUNT":	"testaccount",
				"Uid":	"***",
				"PWD":	"***",
				"Token":	"***",
				"Authenticator":	"snowflake",
				"DATABASE":	"TestDB",
				"SCHEMA":	"TestSchema",
				"WAREHOUSE":	"Regress",
				"ROLE":	"AccountAdmin",
				"SSL":	"1"
			},
			"request":	"snowflake.local.snowflakecomputing.com:443/session/v1/login-request?requestId=c4d53986-ee7a-4f01-9fac-2604653e9c41&request_guid=abbab0e5-5c77-4102-9d29-d44efde6a050&databaseName=testdb&schemaName=testschema&warehouse=regress"
		}
	}]
```

### DIFF
--- a/deps/oob-1.0.2/jsonutil.c
+++ b/deps/oob-1.0.2/jsonutil.c
@@ -98,6 +98,10 @@ char *prepareOOBevent(oobOcspData *ocspevent)
   if( ! tags ) goto end;
   cJSON_AddItemToObject(list, "Tags", tags);
 
+  dsn = cJSON_CreateObject();
+  if (! dsn ) goto end;
+  cJSON_AddItemToObject(list, "DSN", dsn);
+
   getuuid(uuid);
   key = cJSON_CreateString(uuid);
   if( ! key ) goto end;
@@ -125,24 +129,28 @@ char *prepareOOBevent(oobOcspData *ocspevent)
     key = cJSON_CreateString( connectionInfo.account );
     if( ! key ) goto end;
     cJSON_AddItemToObject(tags, "ctx_account", key);
+    cJSON_AddItemToObject(dsn, "account", key);
   }
 
   if( connectionInfo.host[0] != 0 ){
     key = cJSON_CreateString( connectionInfo.host );
     if( ! key ) goto end;
     cJSON_AddItemToObject(tags, "ctx_host", key);
+    cJSON_AddItemToObject(dsn, "host", key);
   }
 
   if( connectionInfo.port[0] != 0 ){
     key = cJSON_CreateString( connectionInfo.port );
     if( ! key ) goto end;
     cJSON_AddItemToObject(tags, "ctx_port", key);
+    cJSON_AddItemToObject(dsn, "port", key);
   }
 
   if( connectionInfo.protocol[0] != 0 ){
     key = cJSON_CreateString( connectionInfo.protocol );
     if( ! key ) goto end;
     cJSON_AddItemToObject(tags, "ctx_protocol", key);
+    cJSON_AddItemToObject(dsn, "protocol", key);
   }
 
   if( connectionInfo.user[0] != 0 ){
@@ -195,6 +203,30 @@ char *prepareOOBevent(oobOcspData *ocspevent)
     key = cJSON_CreateString( "prod" );
     if( ! key ) goto end;
     cJSON_AddItemToObject(tags, "telemetryServerDeployment", key);
+  }
+
+  if (connectionInfo.role[0] != 0) {
+      key = cJSON_CreateString(connectionInfo.role);
+      if (!key) goto end;
+      cJSON_AddItemToObject(dsn, "role", key);
+  }
+
+  if (connectionInfo.warehouse[0] != 0) {
+      key = cJSON_CreateString( connectionInfo.warehouse);
+      if ( ! key ) goto end;
+      cJSON_AddItemToObject(dsn, "warehouse", key);
+  }
+
+  if (connectionInfo.dbName[0] != 0) {
+      key = cJSON_CreateString( connectionInfo.dbName);
+      if ( ! key ) goto end;
+      cJSON_AddItemToObject(dsn, "database", key);
+  }
+
+  if (connectionInfo.schema[0] != 0) {
+      key = cJSON_CreateString( connectionInfo.schema);
+      if ( ! key ) goto end;
+      cJSON_AddItemToObject(dsn, "schema", key);
   }
 
   key = cJSON_CreateString("Log");
--- a/deps/oob-1.0.2/jsonutil.c
+++ b/deps/oob-1.0.2/jsonutil.c
@@ -413,7 +413,7 @@ void setoobConnectioninfo(const char* host,
 }
 
 void setoobDsninfo(const char *key, const char *val) {
-    char* caps = malloc(strlen(key));
+    char* caps = calloc(strlen(key), sizeof(char));
     upper(caps, key);
 
     if (dsn == NULL) {
@@ -421,6 +421,9 @@ void setoobDsninfo(const char *key, const char *val) {
     }
     cJSON *value = cJSON_CreateString(val);
     if (strcmp(caps, "PWD")) {
+        if (!strcmp(caps, "SERVER")) {
+            setdeployment(val);
+        }
         cJSON_AddItemToObject(dsn, key, value);
     } else {
         cJSON_AddItemToObject(dsn, key, cJSON_CreateString("***"));
--- a/deps/oob-1.0.2/jsonutil.c
+++ b/deps/oob-1.0.2/jsonutil.c
@@ -1,4 +1,3 @@
-#include <ctype.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
@@ -411,15 +410,16 @@ void setoobConnectioninfo(const char* host,
     copyString("http", connectionInfo.protocol, 8);
 }
 
-void setOOBDsninfo(const char* key, const char* val) {
-    if (dsn == NULL) {
-        dsn = cJSON_CreateObject();
+void setOOBDsninfo(struct dsnKeyValue kvPair[], int num) {
+    dsn = cJSON_CreateObject();
+    cJSON* val;
+    for (int i = 0; i < num; ++i) {
+        val = cJSON_CreateString(kvPair[i].val);
+        cJSON_AddItemToObject(dsn, kvPair[i].key, val);
+        if (!strcasecmp(kvPair[i].key, "server")) {
+            setdeployment(kvPair[i].val);
+        }
     }
-    cJSON* value = cJSON_CreateString(val);
-    if (strcasecmp(key, "server")) {
-        setdeployment(val);
-    }
-    cJSON_AddItemToObject(dsn, key, value);
     return;
 }
 
--- a/deps/oob-1.0.2/jsonutil.c
+++ b/deps/oob-1.0.2/jsonutil.c
@@ -323,30 +323,6 @@ char *prepareOOBevent(oobOcspData *ocspevent)
     cJSON_AddItemToObject(tags, "telemetryServerDeployment", key);
   }
 
-  if (connectionInfo.role[0] != 0) {
-      key = cJSON_CreateString(connectionInfo.role);
-      if (!key) goto end;
-      cJSON_AddItemToObject(dsn, "role", key);
-  }
-
-  if (connectionInfo.warehouse[0] != 0) {
-      key = cJSON_CreateString( connectionInfo.warehouse);
-      if ( ! key ) goto end;
-      cJSON_AddItemToObject(dsn, "warehouse", key);
-  }
-
-  if (connectionInfo.dbName[0] != 0) {
-      key = cJSON_CreateString( connectionInfo.dbName);
-      if ( ! key ) goto end;
-      cJSON_AddItemToObject(dsn, "database", key);
-  }
-
-  if (connectionInfo.schema[0] != 0) {
-      key = cJSON_CreateString( connectionInfo.schema);
-      if ( ! key ) goto end;
-      cJSON_AddItemToObject(dsn, "schema", key);
-  }
-
   key = cJSON_CreateString("Log");
   if( ! key ) goto end;
   cJSON_AddItemToObject(list, "Type", key);
@@ -358,7 +334,7 @@ char *prepareOOBevent(oobOcspData *ocspevent)
   if(oobevent.urgent) {
     key = cJSON_CreateTrue();
   }
-  else
+  else {
     key = cJSON_CreateFalse();
   }
   if( ! key ) goto end;
@@ -618,7 +594,6 @@ void setoobdsninfo(const char *key, const char *val) {
         copyString(key, dsnKeys.privKey, 16);
         copyString(val, dsnValues.privKey, 1024);
     } else {
-        // TODO handle error/misc cases
         free(caps);
         return;
     }
--- a/deps/oob-1.0.2/jsonutil.c
+++ b/deps/oob-1.0.2/jsonutil.c
@@ -1,3 +1,4 @@
+#include <ctype.h>
 #include <stdio.h>
 #include <string.h>
 #include <stdlib.h>
@@ -20,7 +21,9 @@ static struct conStr connectionInfo = {{0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {
 
 static struct logDetails oobevent = {{0}, {0}, {0}, {0}, 0, 0};
 
-static struct dsnStr dsnInfo = {{0},{0},{0},{0},{0},{0},{0}};
+static struct dsnKey dsnKeys = {{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0}};
+
+static struct dsnVal dsnValues = {{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0},{0}};
 
 void setdeployment(const char *host);
 
@@ -29,13 +32,15 @@ int sendOOBevent(char *event);
 //connStr will be copied into another string
 void setConnectionString(char const* connStr);
 
-char * getConnectionString(void);
+char* getConnectionString(void);
 
 void gettime( char *buffer);
 
 void getuuid(char *buffer);
 
 char* getConnectionInfo(enum OOBINFO id);
+
+void upper(char* caps, const char* word);
 
 void getCabundle(char *cabundle, int maxlen);
 
@@ -63,7 +68,8 @@ char *prepareOOBevent(oobOcspData *ocspevent)
   cJSON *tags = NULL;
   cJSON *dsn = NULL;
   cJSON *vals = NULL; 
-  cJSON *key  = NULL;
+  cJSON *key = NULL;
+  cJSON *val = NULL;
   char *str = NULL;
   char buffer[TMP_BUF_LEN]={0};
   char uuid[TMP_BUF_LEN]={0};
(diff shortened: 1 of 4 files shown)


deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,7 +554,85 @@ void setoobConnectioninfo(const char* host,
...
+        copyString(val, dsnValues.port, 10);
+    } else if (!strcmp(caps, "ACCOUNT")) {
+        copyString(key, dsnKeys.account, 8);
+        copyString(val, dsnValues.account, 256);
+    } else if (!strcmp(caps, "UID")) {
+        copyString(key, dsnKeys.user, 8);
+        copyString(val, dsnValues.user, 256);
+    } else if (!strcmp(caps, "PWD")) {
+        copyString(key, dsnKeys.password, 8);
+        copyString(val, dsnValues.password, 8);
```
sfc-gh-stakeda, line 586: please don't include the secret.

sfc-gh-stakeda:
Add unit test here to cover new codes. Don't defer tests to ODBC.

deps/oob-1.0.2/jsonutil.h
```diff
@@ -51,6 +95,8 @@ void setoobConnectioninfo(const char* host,
     short ssl
     );
 
+void setoobdnsinfo(const char *key, const char *val);
```
sfc-gh-azhan, line 98: function name style doesn't match!

deps/oob-1.0.2/jsonutil.c
```diff
@@ -60,8 +66,10 @@ char *prepareOOBevent(oobOcspData *ocspevent)
   cJSON *root = NULL;
   cJSON *list = NULL;
   cJSON *tags = NULL;
-  cJSON *vals = NULL; 
-  cJSON *key  = NULL;
+  cJSON *vals = NULL;
```
sfc-gh-sshankar, line 69: I don't think we need any of this: This PR seems redundant.

deps/oob-1.0.2/jsonutil.c
```diff
@@ -60,8 +66,10 @@ char *prepareOOBevent(oobOcspData *ocspevent)
...
+  cJSON *dns = NULL;
```
sfc-gh-sshankar, line 70: dns or dsn ?

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,7 +530,84 @@ void setoobConnectioninfo(const char* host,
...
+        copyString(val, dnsValues.privKey, 1024);
+    } else {
+        free(caps);
+        return;
+    }
+    free(caps);
+    return;
+}
+
+void upper(char* caps, const char* word){
```
sfc-gh-sshankar, line 604: You don't need this, and you should never have to write this.

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,7 +530,84 @@ void setoobConnectioninfo(const char* host,
     copyString("https", connectionInfo.protocol, 8);
   else
     copyString("http", connectionInfo.protocol, 8);
+}
 
+void setoobdnsinfo(const char *key, const char *val) {
+    char* caps = malloc(strlen(key));
+    upper(caps, key);
+
+    if (!strcmp(caps, "DESCRIPTION")) {
```
sfc-gh-sshankar, line 539: Why are we doing this ? Can't we just store this as an array of pairs ? And pass in an array of pairs ?

deps/oob-1.0.2/jsonutil.c
```diff
@@ -217,6 +225,124 @@ char *prepareOOBevent(oobOcspData *ocspevent)
   vals = cJSON_CreateObject();
   cJSON_AddItemToObject(list, "Value", vals);
 
+  dns = cJSON_CreateObject();
```
sfc-gh-sshankar, line 228: Why are you doing this on each call to prepare ?
setoobeventdsninfo should take a map, or an array of pairs or something like that. You should construct this cjson object once. Instead of preparing it repeatedly.

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: DNS

deps/oob-1.0.2/jsonutil.h
```diff
@@ -36,6 +36,50 @@ struct logDetails{
...
+    char schema[256];
+    char warehouse[256];
+    char role[256];
+    char ssl[10];
+    char tracing[10];
+    char jwtTimeout[256];
+    char timezone[64];
+    char privKey[1024];
+};
+
```
sfc-gh-vreddy, line 82:
```suggestion
enum {KEY=0,VALUE=1};

#define MAX_KEY 40
#define MAX_VAL 1024

typedef struct keyVal{
  char key[MAX_KEY];
  char val[MAX_VAL];
}keyval;

typedef struct dsnfields
{
  //You can define in a single array but then again you need another enum
  keyval description;
  keyval driver;
  keyval host;
  keyval port;
  //Add more
} dsnFields;
//dsnFields dsn ; 
//You can access them like dsn.driver.key 
```

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: There aren't any maps in C and using a struct is the next best thing. This manual creation is unavoidable and already done in the existing code if you look at connectioninfo.

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: We do because you wanted case sensitive keys and the only way to consolidate the difference between something like "server" and "SERVER".

deps/oob-1.0.2/jsonutil.c
sfc-gh-azhan: I'm confused by the naming. DNS stands for domain name system right? What DSN stands for? It seems the info includes all connection settings from ODBC right? Why not call it `connectionSettings`?

deps/oob-1.0.2/jsonutil.h
sfc-gh-jbahk: I was thinking about this too but I was just afraid of setting a general memory limit for each field. For example description or authenticator need 1024 but most cases require far less so I thought it'd be less efficient memory wise. Also the existing code also seems to point to customizing the memory limit but this can be changed. @sfc-gh-sshankar  What do you think?

sfc-gh-vreddy: Your event message is short lived. 
variable should be a local and goes out of scope after the oob event is sent, thus freeing the stack memory.
And for DSN this is done once at the time of connection.

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: Same issue that these "pairs" don't exist in C so I'd have to create a struct regardless. The best we can do is change it to Viswa's suggestion.

deps/oob-1.0.2/jsonutil.h
sfc-gh-vreddy: Can you also check the oob event with dsn fields is json parsable. 
one way is to check some online parsers.
When i checked the event in the description it was not parsable.

sfc-gh-jbahk: I've said this privately but if you're just copy-pasting the GitHub description there are errors because the single quote (') is recognized differently.

deps/oob-1.0.2/jsonutil.c
sfc-gh-vreddy: It should be DSN which mean Data Source Name

deps/oob-1.0.2/jsonutil.c
sfc-gh-sshankar: That doesn't mean you should do it in each prepare

deps/oob-1.0.2/jsonutil.c
sfc-gh-azhan: aha, DSN should be database source name, that makes sense!

deps/oob-1.0.2/jsonutil.c
sfc-gh-sshankar: By pair  I simply mean const char* kvpair[2]

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: How do you suggest I go about this then? Could you propose an alternative?

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: Ok, I can change this back.

sfc-gh-sshankar: Yes

tests/test_unit_oob.cpp
```diff
@@ -166,9 +171,71 @@ void test_oob(void **) {
     }
 }
 
+void test_dsn(void **) {
```
sfc-gh-azhan, line 174: nit: add comment

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,7 +410,35 @@ void setoobConnectioninfo(const char* host,
...
+        dsn = cJSON_CreateObject();
+    }
+    cJSON *value = cJSON_CreateString(val);
+    if (strcmp(caps, "PWD")) {
+        if (!strcmp(caps, "SERVER")) {
+            setdeployment(val);
+        }
+        cJSON_AddItemToObject(dsn, key, value);
+    } else {
+        cJSON_AddItemToObject(dsn, key, cJSON_CreateString("***"));
```
sfc-gh-azhan, line 429: I know you try to mask sensitive value. Question: do you limit the keys from DSN? If yes, where is the list? If no, that may be a security issue, e.g., if customer set `proxy_pwd` in the key value pair, we are going to miss it.

deps/oob-1.0.2/jsonutil.c
```diff
@@ -16,26 +17,29 @@
 #define TMP_BUF_LEN 40
 #endif
 
-//static struct conStr connectionInfo ;
 static struct conStr connectionInfo = {{0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0} };
 
 static struct logDetails oobevent = {{0}, {0}, {0}, {0}, 0, 0};
 
+cJSON *dsn = NULL;
```
sfc-gh-sshankar, line 24: What is the correct style we adopt ? I see both `type* var` and `type *var`

deps/oob-1.0.2/jsonutil.c
```diff
@@ -16,26 +17,29 @@
...
-char * getConnectionString(void);
+char* getConnectionString(void);
 
 void gettime( char *buffer);
 
 void getuuid(char *buffer);
 
 char* getConnectionInfo(enum OOBINFO id);
 
+void upper(char* caps, const char* word);
```
sfc-gh-sshankar, line 41: Delete

deps/oob-1.0.2/jsonutil.c
sfc-gh-sshankar: Use stricmp or strcasecmp instead. As a rule you should never be writing such basic string processing functions.

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,7 +410,35 @@ void setoobConnectioninfo(const char* host,
...
 
+void setoobDsninfo(const char *key, const char *val) {
+    char* caps = calloc(strlen(key), sizeof(char));
+    upper(caps, key);
+
+    if (dsn == NULL) {
+        dsn = cJSON_CreateObject();
+    }
+    cJSON *value = cJSON_CreateString(val);
+    if (strcmp(caps, "PWD")) {
```
sfc-gh-sshankar, line 423: You shouldn't be doing this here. ODBC maintains a list of "sensitive keys" and you should make sure you don't pass in the real value for this in the first place.

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,7 +410,35 @@ void setoobConnectioninfo(const char* host,
     copyString("https", connectionInfo.protocol, 8);
   else
     copyString("http", connectionInfo.protocol, 8);
+}
 
+void setoobDsninfo(const char *key, const char *val) {
```
sfc-gh-sshankar, line 415: How do you ensure that you don't repeatedly add the same key for multiple connections ?

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,7 +410,35 @@ void setoobConnectioninfo(const char* host,
...
+
+    if (dsn == NULL) {
+        dsn = cJSON_CreateObject();
+    }
+    cJSON *value = cJSON_CreateString(val);
+    if (strcmp(caps, "PWD")) {
+        if (!strcmp(caps, "SERVER")) {
+            setdeployment(val);
+        }
+        cJSON_AddItemToObject(dsn, key, value);
```
sfc-gh-sshankar, line 427: @sfc-gh-azhan who owns the memory for the string that is passed in to AddItemToObject ?

sfc-gh-azhan: The ODBC's m_connectionString owns the strings here. https://github.com/snowflakedb/snowflake-odbc/pull/83/files#diff-8a7d154ab6fecb24f34bcf3a036b2ddaR640
The `cJSON_AddItemToObject` will do a memcpy of the string so should be fine.

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: I learned it as type* var but cJSON *dsn seems to be the convention within the prepareOOBevent function call. I'd be happy to change to type* var for this file.

deps/oob-1.0.2/jsonutil.c
```diff
@@ -217,6 +221,8 @@ char *prepareOOBevent(oobOcspData *ocspevent)
   vals = cJSON_CreateObject();
   cJSON_AddItemToObject(list, "Value", vals);
 
+  cJSON_AddItemToObject(vals, "DSN", dsn);
```
sfc-gh-sshankar, line 224: What is dsn is NULL ? Wouldn't this crash ? Do you have a test for this ? cc @sfc-gh-azhan

deps/oob-1.0.2/jsonutil.c
```diff
@@ -16,27 +17,28 @@
 #define TMP_BUF_LEN 40
 #endif
 
-//static struct conStr connectionInfo ;
 static struct conStr connectionInfo = {{0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0}, {0} };
 
 static struct logDetails oobevent = {{0}, {0}, {0}, {0}, 0, 0};
 
-void setdeployment(const char *host);
+cJSON* dsn = NULL;
```
sfc-gh-azhan, line 24: You need free the data this pointer refers to after the event has been sent. Check out https://github.com/snowflakedb/libsnowflakeclient/blob/994a2b2c63837c8c41eff3113100602d3ee656f5/deps/oob-1.0.2/jsonutil.c#L487

deps/oob-1.0.2/jsonutil.c
```diff
@@ -217,6 +219,8 @@ char *prepareOOBevent(oobOcspData *ocspevent)
   vals = cJSON_CreateObject();
   cJSON_AddItemToObject(list, "Value", vals);
 
+  cJSON_AddItemToObject(vals, "DSN", dsn);
```
sfc-gh-azhan, line 222: Why not check ths is null or not?

tests/test_unit_oob.cpp
```diff
@@ -166,9 +175,89 @@ void test_oob(void **) {
...
+    setConnectionString((char const *) url);
+
+    setOOBeventdata(OOBEVENTNAME, (const char*)"ConnectionDSNConfig", 0);
+    setOOBeventdata(ERRORCODE, NULL, 0);
+    setOOBeventdata(EXCPMSG, "", 0);
+    setOOBeventdata(REQUESTURL, url, 0);
+    setOOBeventdata(OOBSQLSTATE, (const char*)"", 0);
+    setOOBeventdata(OOBCABUNDLE, CABundlePath.c_str(), 0);
+    char *oobevent = prepareOOBevent(nullptr);
+
```
sfc-gh-azhan, line 227: add an assertion that once the event has been prepared, the dsn object should be freed.

deps/oob-1.0.2/jsonutil.c
sfc-gh-azhan: After the usage of dsn, we should free the object.

tests/CMakeLists.txt
```diff
@@ -215,7 +215,7 @@ else()
             )
             target_link_libraries(${T} ${TESTLIB_OPTS_C})
             add_test(${T} ${T})
-            set_tests_properties(${T} PROPERTIES ENVIRONMENT CABUNDLE_PATH=${CABUNDLE_PATH})
+            set_tests_properties(${T} PROPERTIES ENVIRONMENT CABUNDLE_PATH=${SNOWFLAKE_TEST_CA_BUNDLE_FILE})
```
sfc-gh-azhan, line 218: @rvishureddy is this safe to change?

tests/test_include_aws.cpp
```diff
@@ -15,6 +15,30 @@ void test_include_aws(void **unused) {
 }
 
 int main(void) {
+
+#ifdef __APPLE__
```
sfc-gh-azhan, line 19: why add this in this PR?

sfc-gh-vreddy: @sfc-gh-azhan 
The master has this changes for a put/get issue. And his branch was cut before that got merged. 
Ideally jenkins merges this PR with master and runs tests. But somehow this PR checks still see put/get test failing. 
So recommended him to merge with master to see if that fixes the Checks failing.

deps/oob-1.0.2/jsonutil.c
sfc-gh-vreddy: @sfc-gh-azhan This dsn object becomes part of cJSON* root which is freed after preparing the oobevent and converted to string. 
Line 320 in jsonutil.c 

```
  if (root)
  {
    str = cJSON_Print(root);
    cJSON_Delete(root);
  }
  freeAll();
  return str;
```

tests/CMakeLists.txt
sfc-gh-vreddy: @sfc-gh-azhan These is result of merging master to this PR.

deps/oob-1.0.2/jsonutil.c
sfc-gh-jbahk: As discussed, the test_oob test case within test_unit_oob.cpp covers the case when dsn is null; it simply does not add this object to the overall oob payload

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,12 +408,24 @@ void setoobConnectioninfo(const char* host,
     copyString("https", connectionInfo.protocol, 8);
   else
     copyString("http", connectionInfo.protocol, 8);
+}
 
+void setOOBDsninfo(struct dsnKeyValue kvPair[], int num) {
```
sfc-gh-sshankar, line 413: setOODBDsn**I**nfo

deps/oob-1.0.2/jsonutil.c
```diff
@@ -404,12 +408,24 @@ void setoobConnectioninfo(const char* host,
...
+    dsn = cJSON_CreateObject();
+    cJSON* val;
+    for (int i = 0; i < num; ++i) {
+        val = cJSON_CreateString(kvPair[i].val);
```
sfc-gh-sshankar, line 417: Nit: Just do 
```
cJSON* val = cJSON_CreateString(kvPair[i].val)
```

deps/oob-1.0.2/jsonutil.h
```diff
@@ -36,6 +36,11 @@ struct logDetails{
   int urgent;
 };
 
+typedef struct dsnKeyValue{
```
sfc-gh-sshankar, line 39: typedef struct KeyValuePair

a. Please capitalize type names
b. The type name is not specific to DSn, so just call is KeyBaluePair, or better yet
KeyValueStringPair

deps/oob-1.0.2/jsonutil.h
```diff
@@ -36,6 +36,11 @@ struct logDetails{
...
+  const char* key;
+  const char* val;
+} dsnKeyValue;
```
sfc-gh-sshankar, line 42: Here too captialize.

deps/oob-1.0.2/jsonutil.c
sfc-gh-sshankar, line 413: Do you need the struct keyword here if you have a typdef ?
You should just do 
```
void setOOBDsnInfo(KeyValuePair kvPair[], int num)
```

tests/test_unit_oob.cpp
```diff
@@ -166,9 +175,89 @@ void test_oob(void **) {
     }
 }
 
+// Tests DSN logging in OOB
+void test_dsn(void **) {
```
sfc-gh-sshankar, line 179: Why void** ?

PR #219, opened 2020-07-15
add BUILD_CLEAN flag

sfc-gh-stakeda:
I found the local build test doesn't detect inconsistency between dependencies and libsnowflakeclient.
This change is to clean up the dependency's staging area before running the build.

`BUILD_CLEAN` can be set to `false` if the download/build can be skipped for faster dev test iteration optionally.


sfc-gh-vreddy:
@sfc-gh-stakeda BUILD_CLEAN is set to true on local dev vms this will clean the existing dependencies and downloads fresh. 
And on jenkins irrespective its a clean build so always downloaded from s3 bucket. Is my understanding correct.

sfc-gh-stakeda:
In all build environemtns, if `BUILD_CLEAN` is not set, the dependency's staging area will be cleaned. ONLY if `BUILD_CLEAN` is 'false` for your local test, it won't.

Issue #221, opened 2020-07-18
SNOW-176651:  Wrong result with timestamp with precision 0


sfc-gh-smirzaei:
For a simple queries like following:
`select '2020-07-1'::timestamp_ltz(0)`
`select '2020-07-1'::timestamp_ntz(0)`
`select '2020-07-1'::timestamp_tz(0)`
`select '2020-07-1'::timestamp(0)`
The results is `Dec 30 1899...`. I.e., tm_year, tm_mon ... fields of tm are all uninitialized.

sfc-gh-smirzaei:
I have a fix, will send it for review soon

sfc-gh-jtang:
@sfc-gh-igarish can you check if it's still an open issue? looks we had a fix to review about 3 years ago :)

sfc-gh-igarish:
PR#222 has the fix.

PR #222, opened 2020-07-20
Handling timestamps with zerp precision

sfc-gh-smirzaei:
The current implementation is not dealing with TSs with precision 0 and considers them as malformed.
closing #221


sfc-gh-kwagner:
@sfc-gh-smirzaei can you add a test case for this change? Otherwise it looks good to me

PR #223, opened 2020-07-21
Use uRandom device for Athena and put it behind a flag in simbaini

sfc-gh-vreddy:
The fix includes using a /dev/urandom instead of /dev/random based on the flag in SIMBAINI (ODBC). 

This can be set using transferagent method.


cpp/FileTransferAgent.hpp
```diff
@@ -59,6 +59,16 @@ class FileTransferAgent : public IFileTransferAgent
...
+  /**
+   * Set useUrand to true to use /dev/urandom device
+   * Set it to false to use /dev/random device
+   * @param useUrand
+   */
+  virtual void setRandomDeviceAsUrand(bool useUrand)
+  {
+    m_useDevUrand = useUrand;
+  }
+
```
sfc-gh-vreddy, line 71: This setRandomDeviceAsUrand() is invoked by filetransferagent and is used from ODBC and applications directly. 

The other setRandomDev is internal func in cpp/FileMetadataInitializer.hpp which sets the real random device to be used and the Encryptor uses the device value from this func.

cpp/FileTransferAgent.cpp
```diff
@@ -139,6 +145,7 @@ void Snowflake::Client::FileTransferAgent::initFileMetadata(std::string *command
   m_FileMetadataInitializer.setAutoCompress(response.autoCompress);
   m_FileMetadataInitializer.setSourceCompression(response.sourceCompression);
   m_FileMetadataInitializer.setEncryptionMaterials(&response.encryptionMaterials);
+  m_FileMetadataInitializer.setRandomDev(m_useDevUrand);
```
sfc-gh-vreddy, line 148: setRandomDev is internal func in cpp/FileMetadataInitializer.hpp which sets the real random device to be used and the Encryptor uses the device value from this func.

The other setRandomDeviceAsUrand() is invoked by filetransferagent and is used from ODBC and applications directly.

sfc-gh-stakeda:
How does the customer enable / disable the new random function? What is the simba parameter?

sfc-gh-vreddy:
> How does the customer enable / disable the new random function? What is the simba parameter?

@sfc-gh-stakeda Here is the ODBC PR https://github.com/snowflakedb/snowflake-odbc/pull/108/files

SIMBA setting is `UseURandomDevice`

PR #228, opened 2020-08-11
SNOW-57187 Refactor row metadata in response (libsnowflakeclient changes)

sfc-gh-kwang:
Description
See snowflakedb/snowflake#11348 for how response changed.
Screen Shot 2020-08-10 at 4 36 20 PM

Testing
test_stats passed but still can't enable until changes go out in GS (see screenshot)

### DIFF
--- a/include/snowflake/Statement.hpp
+++ b/include/snowflake/Statement.hpp
@@ -32,7 +32,7 @@ namespace Snowflake {
 
             SF_COLUMN_DESC *desc();
 
-            SF_ROW_METADATA *rowMetadata();
+            SF_STATS *stats();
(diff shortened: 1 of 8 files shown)


sfc-gh-kwang:
> Just curious, why the name change?

name change to keep the field name consistent with other responses like query monitoring details response. But the response change also includes the stats field being an object rather than an array of an object.
See https://docs.google.com/document/d/1GfSLu66RrHCcUNC2GLr2s63y6g4EhD5bNcq66g9ewWA/edit?ts=5ee198f9#heading=h.z33igma7kv2l

PR #241, opened 2020-11-06
Snow 218104 fast fail put cmd

sfc-gh-vreddy:
When uploading multiple files with wild char * in put command, 
This feature enables to stop uploading the pending files when one file fails to upload. 

In case of multiple threads uploading files, 
Given 100 files to upload and the thread pool size is 4
If the thread pool size is 4 then at any point 4 files will be uploaded in parallel. And if one of the file fails to upload then the 3 files which are already being uploaded will continue to upload. The rest of the 96 files will not be uploaded.
And the corresponding error code is set.


cpp/FileTransferAgent.cpp
```diff
@@ -283,19 +293,31 @@ void Snowflake::Client::FileTransferAgent::uploadFilesInParallel(std::string *co
...
       {
-        RemoteStorageRequestOutcome outcome = uploadSingleFile(m_storageClient, metadata,
-          resultIndex);
+        RemoteStorageRequestOutcome outcome = RemoteStorageRequestOutcome::SUCCESS;
+        if(isPutFastFailEnabled() && !failedTransfers.empty())
+        {
+          m_executionResults->SetTransferOutCome(RemoteStorageRequestOutcome::SKIP_UPLOAD_FILE, resultIndex);
+          CXX_LOG_DEBUG("Sequential upload, put fast fail enabled, Skipping file.");
+          break;
+        }
```
sfc-gh-vreddy, line 302: When a put fails, we append to the string the filename that failed to upload. 
If fast fail is enabled and there already failed a file to upload then we skip the upload of the rest of the files.

cpp/FileTransferAgent.cpp
```diff
@@ -283,19 +293,31 @@ void Snowflake::Client::FileTransferAgent::uploadFilesInParallel(std::string *co
...
         {
           if (outcome == RemoteStorageRequestOutcome::FAILED) {
-            CXX_LOG_DEBUG("Failed to upload %s ", metadata->srcFileName.c_str());
+            CXX_LOG_DEBUG("Sequential upload %s FAILED.", metadata->srcFileName.c_str());
             failedTransfers.append(metadata->srcFileName) + ", ";
+            //Fast fail, return when the first file fails to upload.
+            if(isPutFastFailEnabled())
+            {
+              CXX_LOG_DEBUG("Sequential upload, put fast fail enabled, Skip uploading rest of the files.");
+            }
```
sfc-gh-kwagner, line 320: Shouldn't you have a break or return statement here?

cpp/FileTransferAgent.hpp
```diff
@@ -170,6 +172,25 @@ class FileTransferAgent : public IFileTransferAgent
     m_useDevUrand = useUrand;
   }
 
+  virtual void setPutFastFail(bool fastFail)
+  {
+    if(fastFail)
+    {
+      CXX_LOG_DEBUG("Enabling put fast fail.");
+    }
+    m_fastFail = fastFail;
```
sfc-gh-kwagner, line 181: Shouldn't `m_fastFail = fastFail` be inside the if statement? Or do you even need this if statement at all?

sfc-gh-vreddy: The function intention is to set the value of fastFail into m_fastFail which gets done irrespective of the if condition. 
And the log message is for debugging purpose.

cpp/FileTransferAgent.cpp
sfc-gh-vreddy: If returned from here then we do not throw the exception and ODBC driver will not know of the error. 
break statement in 325 already covers this. So did not add break again.

sfc-gh-kwagner (approved):

sfc-gh-vreddy:
@sfc-gh-stakeda Merge conflicts resolved can you take a look at this.

sfc-gh-stakeda (approved):

sfc-gh-vreddy:
http://172.16.20.10:8080/job/LibSnowflakeClient-Mac-Universal-Release-Test/789/
@sfc-gh-stakeda Mac builds also passed. Could you merge this.

sfc-gh-vreddy:
@sfc-gh-stakeda Could you please merge this. 
http://172.16.20.10:8080/job/LibSnowflakeClient-Mac-Universal-Release-Test/833/ 
http://172.16.20.10:8080/job/LibSnowflakeClient-Mac-Universal-Debug-Test/995/

PR #242, opened 2020-11-06
SNOW-219403 Add configurable retries for put fails

sfc-gh-vreddy:
Adds the ability to configure the number of retries to perform when a put command fails.

Test case is added to test if the set number of retries are performed.

### DIFF
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -70,7 +70,8 @@ Snowflake::Client::FileTransferAgent::FileTransferAgent(
   m_transferConfig(transferConfig),
   m_uploadStream(nullptr),
   m_uploadStreamSize(0),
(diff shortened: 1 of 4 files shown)


sfc-gh-kwagner (approved):

sfc-gh-stakeda:
How does the customer set the maximum retry counter?

sfc-gh-vreddy:
> How does the customer set the maximum retry counter?
ODBC driver will invoke the transferagent api to set this value. 
And ODBC reads this value from DSN .

sfc-gh-stakeda:
Ok. If you add a parameter to the DSN, consider making it a session parameter as well so that either customer / support can change the behaviors without changing the actual DSN.

sfc-gh-stakeda (approved):

PR #243, opened 2020-11-12
retry curl couldnt connect

sfc-gh-smirzaei:
We have a CRITICAL ticket (SNOW-208699) with several customers affected. It seems we have some XP to GS connection issue that almost always (>99%) happens on AZURE which cause stored proc's child queries fails with `curl_easy_perform() failed: Couldn't connect to server`.  As you know in stored proc we use LibSfClient to prepare and execute children query statements. What we are seeing is that occasionally for a proc that runs let say 10 children jobs (just an example from many), 8th jobs only fails (1, ..7, 9 and 10 are good) while we use the same SF_CONNECT.
My guess is it's going to take a while to figure out what's going on with AZURE (I believe this issue is not new an we have seen similar connection issues else where), and until then, retrying the same curl request can help.


sfc-gh-kwagner (approved):

lib/http_perform.c
```diff
@@ -131,10 +131,13 @@ sf_bool STDCALL http_perform(CURL *curl,
                              int64 network_timeout,
                              sf_bool chunk_downloader,
                              SF_ERROR_STRUCT *error,
-                             sf_bool insecure_mode) {
+                             sf_bool insecure_mode,
+                             sf_bool retry_on_curle_couldnt_connect,
+                             int8 retry_on_curle_couldnt_connect_count) {
```
sfc-gh-vreddy, line 136: This is a transient failure and retry logic should work, It would be better if retry context is used with exponential backoff and sleep with jitter is used.

lib/http_perform.c
```diff
@@ -318,8 +334,12 @@ sf_bool STDCALL http_perform(CURL *curl,
             } else if (http_code != 200) {
                 retry = is_retryable_http_code(http_code);
                 if (!retry) {
-                    SET_SNOWFLAKE_ERROR(error, SF_STATUS_ERROR_RETRY,
-                                        "Received unretryable http code",
+                    char msg[1024];
```
sfc-gh-smirzaei, line 337: This change is new and is due to some customer cases that we see that get `Received unretryable http code`. I thought it make sense to know the `http_code` to give us a better view of why we are getting this!

sfc-gh-kwagner (approved):

lib/http_perform.c
```diff
@@ -294,19 +297,32 @@ sf_bool STDCALL http_perform(CURL *curl,
...
+                                              retry_on_curle_couldnt_connect_count)
+            {
+              retry = SF_BOOLEAN_TRUE;
+              uint32 next_sleep_in_secs = retry_ctx_next_sleep(curl_retry_ctx);
+              log_error(
+                      "curl_easy_perform() failed connecting to server on attempt %d, "
+                      "will retry after %d second",
+                      curl_retry_ctx->retry_count,
+                      next_sleep_in_secs);
+              sleep(next_sleep_in_secs);
```
sfc-gh-vreddy, line 310: libsnowflakeclient builds on windows/mac/linux 
sleep may not work on winodws. 
You can try this 
#ifdef _WIN32
            Sleep(sleepTime);  // Sleep for sleepTime milli seconds (Sleep(<time in milliseconds>) in windows)
#else
            std::this_thread::sleep_for(std::chrono::milliseconds (std::chrono::milliseconds(sleepTime)));
or you can use the sleep(number of seconds) API as it works on mac/linux
#endif

lib/http_perform.c
```diff
@@ -12,6 +12,7 @@
 #endif
 
 #include <string.h>
+#include <unistd.h>
```
sfc-gh-vreddy, line 15: Header not available on windows
#include <windows.h> this header in windows has Sleep() API defined.

sfc-gh-vreddy (changes requested):
unistd header not available in windows
And sleep API not available in windows

sfc-gh-smirzaei:
I'll be merging this soon (by the end of the day). Please let me know if there is any changes needed.

sfc-gh-vreddy (approved):
LGTM

PR #248, opened 2020-11-24
Smirzaei get column string representation from raw value

sfc-gh-smirzaei:
This is a change that we need for our use case in SQL stored procedures. 
Basically here is a short background on this change: Inside SQL stored procs (which are translated to JS stored procs) we will be evaluating expressions as much as we can on the XP side (through XP expression evaluation framework) instead of issuing an individual child job through LibSnowflakeClient. E.g. consider expression `'2020-10-10' + 1` which we can via two approaches, i.e., 1) either directly in XP or 2) or with a child query `SELECT '2020-10-10' + 1` through libSfClient where the result is extracted using `snowflake_column_as_str` as a string. But if this expression is evaluated on XP side we get the internal representation which is an epoch value (the same raw value representation in LibSfClient). So to have consistent representation between this two approach, we should be able to go from raw representation to LibSfClient's string representation.

Edit: unit testing is missing which I will add it soon

### DIFF
--- a/include/snowflake/client.h
+++ b/include/snowflake/client.h
@@ -845,6 +845,32 @@ SF_STATUS STDCALL snowflake_column_as_timestamp(SF_STMT *sfstmt, int idx, SF_TIM
 SF_STATUS STDCALL snowflake_column_as_const_str(SF_STMT *sfstmt, int idx, const char **value_ptr);
 
 /**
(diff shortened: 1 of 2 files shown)


sfc-gh-vreddy (approved):
LGTM

sfc-gh-kwagner (approved):

PR #250, opened 2020-12-01
Add retry for the parts in the multipart upload

sfc-gh-vreddy:
Huge files are split into multi parts and uploaded in parts. These are parts are uploaded by the corresponding cloud SDK's. Though there is a retry logic in the SDK's it still fails. 

Adding a retry at a higher level. Similar fix has helped single file uploads when the file fails to upload even after a retry in the cloud SDK. 

The retry logic is added only for aws client. Once the customer confirms this has helped them we can extend this to the other cloud providers.


sfc-gh-wshangguan:
Hi @sfc-gh-vreddy, the fix looks good to me. But I just wanted to double check with you that in that specific failure, libsnowflake will raise an exception, which then make ODBC throw an exception to our customer, is that correct?

sfc-gh-wshangguan (approved):
LGTM

sfc-gh-hchaturvedi:
What is the total time a user might end up waiting? Also do we give any form of feedback to the customer while they're waiting that we're retrying?

PR #251, opened 2020-12-03
SNOW-218025 catch exception in threads for upload and download

SimbaGithub:
File upload and download are not exception safe. Exception could be thrown on renew token, encryption, compression, file I/O and upload/download for gcp through REST request.
When do file transfer in parallel with multiple threads (small files), exception can't be caught on main thread and could crash the application.
1. Catch exception in threads for upload/download small files.
2. Add test cases
3. Fix missing default value of fast fail (false)


sfc-gh-kwagner (approved):
LGTM

sfc-gh-vreddy (approved):
LGTM

sfc-gh-vreddy (approved):
LGTM

PR #252, opened 2020-12-14
Snow-237995: Add force_arrow support to the C API library, and add a field to SF_STMT to extract response headers

sfc-gh-ema:
For Java/Scala Stored procs, we wish to extract the response headers from libsnowflakeclient queries so that the JDBC instance in the UDF server can interpret them. Also, force_arrow is enabled from XP so that the client driver can parse Arrow results directly.

https://snowflakecomputing.atlassian.net/browse/SNOW-237995


sfc-gh-ema:
@sfc-gh-kwagner , @sfc-gh-smirzaei , @sfc-gh-vreddy , could you review? It seems that I can't request you guys to review since I am not part of the repo owners list.

sfc-gh-ema:
Abandoned for: https://github.com/snowflakedb/libsnowflakeclient/pull/253

PR #254, opened 2020-12-15
Add Arrow lib to project

SimbaGithub:
First PR for Arrow support.
This is a minimal set of changes that only modifies the build files to add Arrow as a dependency. It makes no changes to the existing source code.

- Update build files to download pre-build Arrow binaries from SFC Dev1 (temporary workaround in place of a proper build script for macOS and Windows).
- Update CMakeLists.txt files to include Arrow as a dependency.


SimbaGithub:
@sfc-gh-kwagner @rvishureddy Looking for your input on the matter of adding Arrow to the project.

Ideally, we would like to write a proper build script to compile Arrow from source, but we have some concerns.


The cleanest solution would be to write a proper build script to compile Arrow from source. The concerns with that are as follows:

- The Snowflake Arrow repository is currently private. Customers (and ourselves) would not be able to run the build scripts locally as they wouldn't have the credentials to clone the source.
- We would like to reduce duplicate work as much as possible. Presumably, there's already an existing set of build scripts for Arrow that is being used internally (otherwise I'm unsure how there are binaries for macOS and Windows). Is it possible to share any such files and add them straight to this PR?

Alternatively, we could upload pre-built binaries directly into the repository. This is unfavorable, but probably the quick and dirty solution. The only concern with this approach is that we don't have access to any binaries built with VS16 (we only have VS14).

In terms of integrating it with your CI, I see that as part of the CI workflow, dependencies are downloaded from SFC. We could replicate this flow for Arrow to get the merge checks working, but locally, this would force any customers to build strictly using VS14 until we can add more binaries to cover the gaps.

sfc-gh-kwagner:
@SimbaGithub why can't we use the build script that ODBC uses to build arrow for the C API? I don't think we should just store the binary in this repo. I would prefer if we uploaded the binary to S3 and then download from there as opposed to building it everytime

SimbaGithub:
@sfc-gh-kwagner That's a good point. Let me try to capture my thoughts as best as I can.

### ODBC and libsnowflakeclient build & test workflow

I'm trying to reference the OBDC workflow as closely as possible. For what it's worth, the two projects are pretty similar, but things that work on OBDC don't seem to work on libsnowflakeclient?

For ODBC, the relevant build files are [here](https://github.com/snowflakedb/snowflake-odbc/blob/master/Installer/build_arrow.sh) and [here](https://github.com/snowflakedb/snowflake-odbc/blob/master/Installer/build_dependencies.bat). As you can see, there's only a proper build script for Arrow on Linux. Windows downloads from S3 whereas macOS doesn't run at all. I'm unsure as to where and how Arrow binaries are being compiled/sourced. All in all, ODBC and libsnowflakeclient have a similar workflow, but there are some issues that make the workflow fail here. I'll try to go over the issues I'm currently facing.

### Issues on Linux

On Linux, there are some issues with what I believe to be the AWS credentials on the images running these merge checks. Here are my findings after running these merge checks several times. I'll start with a brief snippet of the `scripts/arrow.mk` file that goes over the critical build logic.

```
# Check if the src directory exists. If not, we have to download the source.
SRC_EXISTS := $(shell if [ -d src ]; then echo 1; fi)
ifneq ($(SRC_EXISTS), 1)
	DOWNLOAD_SRC := 1
# --- snip ---
download-src:
	@echo $(DOWNLOAD_SRC) $(SRC_EXISTS)
ifeq ($(DOWNLOAD_SRC),1)
	@$(RM) src && aws s3 cp --region=us-west-2 $(SRC_URL) $(SRC_FILE)
    # --- snip ---
```

When the merge checks get triggered, they fail with the following output:

```
Options:
  target       = Release
PATH=/usr/local/bin:/usr/local/bin:/usr/lib64/ccache:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
mkdir -p build
mkdir -p ../deps-build/linux/Release/arrow_deps/lib
mkdir -p ../deps-build/linux/Release/arrow_deps/include
mkdir -p ../deps-build/linux/Release/boost/lib
mkdir -p ../deps-build/linux/Release/boost/include
mkdir -p ../deps-build/linux/Release/arrow/lib64
mkdir -p ../deps-build/linux/Release/arrow/include
1
fatal error: Unable to locate credentials
make: *** [download-src] Error 1
Error: Process completed with exit code 2.
```

So it seems to me that `$DOWNLOAD_SRC` is correctly set to `1`, but the `aws s3 cp ...` command fails.
I'm not super familiar with AWS, but here's the most relevant [support article](https://aws.amazon.com/premiumsupport/knowledge-center/s3-locate-credentials-error/) I could find. To me, it suggests there is something wrong about the AWS configuration on these images.

Right now, I'm in the middle of changing the Linux workflow to also pull from S3 instead of building from source. I don't think this approach is ideal in that it blocks any customer who builds libsnowflakeclient from source, but theoretically it should work.

### Issues on Windows

On Windows, there doesn't seem to be any Arrow binaries that match the same VS version as the ones being used in the merge checks. Harsh communicated that the current binaries on the S3 bucket are compiled using VS14, whereas Github Actions is currently defined to use VS16.

I tried modifying the workflow to use VS14 instead (see [here](https://github.com/snowflakedb/libsnowflakeclient/commit/dafb9168c1f0707305b5bbf3da659b55b17ca149#diff-063ca77e012f959eba648db60e6868875fd1e70e5f5ac081193d848f8d61ef32)), but I see that the checks are failing with no logs or anything. To me, this suggests there aren't any machines that can run using VS14, but perhaps I made a mistake in my changes. I'm not super familiar with GA so it's more than possible.

SimbaGithub:
I took the suggestions of @sfc-gh-kwagner and tried to resolve the issues I was previously facing, but in doing so I'm running into new issues.

### Linux

I've added the following lines to `.github/workflows/build-test.yml`:

```
jobs:
    build-test-linux:
        # --- snip ---
            - name: Build
              shell: bash
              env:
                  BUILD_TYPE: ${{ matrix.build_type }}
                  WHITESOURCE_API_KEY: ${{ secrets.WHITESOURCE_API_KEY }}
                + AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
                + AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
              run: ci/build_linux.sh
            # --- snip ---
            - name: Test
              shell: bash
              env:
                  BUILD_TYPE: ${{ matrix.build_type }}
                  CLOUD_PROVIDER: ${{ matrix.cloud_provider }}
                  PARAMETERS_SECRET: ${{ secrets.PARAMETERS_SECRET }}
                + AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
                + AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
              run: ci/test_linux.sh
```

This results in the following:

```
+ /mnt/host/scripts/build_arrow.sh -t Release
Options:
  target       = Release
PATH=/usr/local/bin:/usr/local/bin:/usr/lib64/ccache:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin
[INFO] For linux, use pre-built binaries for Arrow
=== downloading s3://sfc-dev1-data/dependency/arrow/arrow_linux_Release-0.15.0.tar.gz to /mnt/host/scripts/../deps-build/linux/Release
fatal error: An error occurred (403) when calling the HeadObject operation: Forbidden
Error: Process completed with exit code 1.
```

I suspect that there might be some additional configuration we need to do to meet the bucket policy, but I'm not certain.

### Windows

To bring us back up to speed, I'm mainly having issues pulling Arrow from S3. Arrow binaries archived on S3 are built with VS14, but Build and Test workflow run on VS16.

To fix this, I tried modifying `.github/workflows/build-test.yml` to change the Windows machines to use VS14.

#### Changing `runs-on` and `vs_version`

At first, I tried changing two lines:

1. from `runs-on: windows-2019` to `runs-on: windows-2019`
2. from `vs_version: [ 'VS16' ]` to `vs_version: [ 'VS14' ]`

The workflow does not run at all. I see no logs available when looking at the checks, only a blank interface with an 'X'. I can only guess that there is machine that qualifies as `windows-2015`.

#### Changing only `vs_version`

Next, I reverted `windows-2015` back to `windows-2019`, keeping only the changes to `vs_version`.
This time, I got to see the build output, but this time, it failed at `scripts/build_zlib.bat` which comes even earlier than anything regarding Arrow! The exact error message is: `MSB4019: The imported project "D:\Microsoft.Cpp.Default.props" was not found.`

#### Adding `setup-msbuild`

I was suggested to try using [setup-msbuild](https://github.com/marketplace/actions/setup-msbuild). I modified `.github/workflows/build-test.yml` as follows:

```
    build-test-win:
        # --- snip ---
        steps:
          + - name: Add msbuild to PATH
          +   uses: microsoft/setup-msbuild@v1.0.2
          +   with:
          +     vs-version: '[ 14.0, 15.0 )'
```

but this didn't seem to accomplish much. The tool `vswhere.exe` tries to find the MSBuild executable on the Windows machine running the workflow, but it can't find it. I've also tried the following values for `vs-version`: `'[ 14.0, 14.1 ]'` and `[ 14 ]`.

I'm a little confused as to how to proceed. Any help would be greatly appreciated.

Julian

sfc-gh-stakeda:
@SimbaGithub can you wait for @sfc-gh-vreddy to come back next year?

Looks like you are in trouble with downloading artifacts from the internal S3 bucket. Adding `AWS_ACCESS_KEY_ID` won't help.

`VS16` is used in github actions, while `VS14` is used internally in Jenkins, because the library needs to be linked with the ODBC driver. All artifacts including libsnowflakeclient.tar and zip are staged in the internal S3 bucket so that the ODBC can download them and build later. `VS16` jobs here is merely to verify build and tests in `windows-2019` Github Action because there is no older VM available.  For production build and test, the internal Jenkins is used. At this moment, that is not accessible to your team.

Anyway please work with @sfc-gh-vreddy and @sfc-gh-abhatnagar for the build script updates.

SimbaGithub:
Thanks for the response!
Sure, that's fine. I'll continue on some additional work with the source code while I wait for @sfc-gh-vreddy to come back.

Cheers,
Julian

PR #253, opened 2020-12-14
SNOW-237995: Add force_arrow support to the C API library, and add a field to SF_STMT to extract response headers

sfc-gh-ema:
For Java/Scala Stored procs, we wish to extract the response headers from libsnowflakeclient queries so that the JDBC instance in the UDF server can interpret them. Also, force_arrow is enabled from XP so that the client driver can parse Arrow results directly.

https://snowflakecomputing.atlassian.net/browse/SNOW-237995

Note that there is also a companion PR, that's necessary to make this work:
https://github.com/snowflakedb/snowflake/pull/18521

PR #256, opened 2020-12-17
SNOW-249762: Refactor Query-Result Capture

sfc-gh-ema:
In SNOW-237995, and in PR #253, I modified SF_STMT to contain a field `query_response_text`, which used memory allocated by libsnowflakeclient.

This task is to track refactoring this code to instead use caller-managed memory via a new struct called SF_QUERY_RESPONSE_CAPTURE and a method `snowflake_execute_with_capture`.

### DIFF
--- a/include/snowflake/client.h
+++ b/include/snowflake/client.h
@@ -421,8 +421,11 @@ typedef struct SF_TIMESTAMP {
 
 /**
  * Initializes an SF_QUERY_RESPONSE_CAPTURE struct.
+ * Note that these need to be released by calling snowflake_query_result_capture_term().
+ *
+ * @param input pointer to an uninitialized SF_QUERY_RESULT_CAPTURE struct pointer.
  */
-SF_QUERY_RESULT_CAPTURE snowflake_query_result_capture_init();
+void STDCALL snowflake_query_result_capture_init(SF_QUERY_RESULT_CAPTURE **input);
 
 /**
  * Checks whether the client is running in force_arrow mode.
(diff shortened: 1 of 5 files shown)


sfc-gh-smirzaei (commented):
The parts that we discuss last night and I can make sense of look good. But Kris is a better person to review and approve this.

sfc-gh-smirzaei (approved):

include/snowflake/client.h
```diff
@@ -403,6 +419,11 @@ typedef struct SF_TIMESTAMP {
     SF_DB_TYPE ts_type;
 } SF_TIMESTAMP;
 
+/**
+ * Initializes an SF_QUERY_RESPONSE_CAPTURE struct.
+ */
+SF_QUERY_RESULT_CAPTURE sf_query_result_capture_init();
```
sfc-gh-kwagner, line 425: Can you follow the naming convention for the public methods? Just replace `sf` with `snowflake`

sfc-gh-kwagner: Also why not dynamically allocate this struct and pass it back through a pointer? Just to keep it consistent with the other init functions

sfc-gh-ema: Done.

sfc-gh-ema: Sounds good, will do.

include/snowflake/client.h
```diff
@@ -403,6 +419,11 @@ typedef struct SF_TIMESTAMP {
     SF_DB_TYPE ts_type;
 } SF_TIMESTAMP;
 
+/**
+ * Initializes an SF_QUERY_RESPONSE_CAPTURE struct.
+ */
+SF_QUERY_RESULT_CAPTURE snowflake_query_result_capture_init();
```
sfc-gh-kwagner, line 425: Also this is missing STDCALL from the function definition (don't forget to add that in the .c file as well). And can we ad a term function for this?

include/snowflake/client.h
```diff
@@ -403,6 +419,11 @@ typedef struct SF_TIMESTAMP {
...
+
 /**
  * Checks whether the client is running in force_arrow mode.
  *
```
sfc-gh-kwagner, line 429: I missed this in the last review, but can we prefix this with `snowflake` and add STDCALL the the function definition

sfc-gh-kwagner (approved):

PR #257, opened 2020-12-31
Simba incident 00284803: crash with named parameters more than 8

SimbaGithub:
This is for the issue reported for PHP driver https://github.com/snowflakedb/pdo_snowflake/issues/220
The root cause is that the new size is incorrect when growing the buffer for named parameters.
Add test cases for more than 8 named parameters and positional parameters as well.


sfc-gh-vreddy (approved):
LGTM

PR #264, opened 2021-02-17
Avoid empty log file generation.

SimbaGithub:
Simba incident 00295760: PHP driver creates empty log files.
Delay log file creation to when actual log output is needed.


sfc-gh-kwagner (approved):
LGTM

PR #268, opened 2021-02-26
SNOW-289058 Refactor RESULT_CAPTURE to allocate response space in-library

sfc-gh-zli:
Previously we require the caller of `snowflake_execute_with_capture` to allocate memory for `result_capture->capture_buffer`. The initial motivation is to make it easy for xp to manage dynamically allocated memories. However, the draw back is that caller have no idea about how much memory should be allocated. 

This PR aims to refactor RESULT_CAPTURE to allocate response space in library. The memory is allocated in `snowflake_execute_with_capture` to variable `result_capture->capture_buffer`. Then the memory is freed in `snowflake_query_result_capture_term`.

We rely on caller following the pattern of calling `snowflake_query_result_capture_init` to init result_capture and calling `snowflake_query_result_capture_term` to free the memory.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -1340,7 +1340,6 @@ void STDCALL snowflake_query_result_capture_init(SF_QUERY_RESULT_CAPTURE **init)
     SF_QUERY_RESULT_CAPTURE *capture = (SF_QUERY_RESULT_CAPTURE *) SF_CALLOC(1, sizeof(SF_QUERY_RESULT_CAPTURE));
 
     capture->capture_buffer = NULL;
-    capture->buffer_size = 0;
     capture->actual_response_size = 0;
 
     *init = capture;
@@ -1375,7 +1374,7 @@ void STDCALL snowflake_bind_input_init(SF_BIND_INPUT * input)
  */
 void STDCALL snowflake_query_result_capture_term(SF_QUERY_RESULT_CAPTURE *capture) {
     if (capture) {
-        capture->capture_buffer = NULL;
+        SF_FREE(capture->capture_buffer);
         SF_FREE(capture);
     }
 }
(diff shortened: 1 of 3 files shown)


lib/client.c
```diff
@@ -1853,17 +1852,14 @@ SF_STATUS STDCALL _snowflake_execute_ex(SF_STMT *sfstmt,
     if (request(sfstmt->connection, &resp, queryURL, url_params,
                 url_paramSize , s_body, NULL,
                 POST_REQUEST_TYPE, &sfstmt->error, is_put_get_command)) {
-        s_resp = snowflake_cJSON_Print(resp);
```
sfc-gh-ema, line 1856: This changes the behavior, no? Now we don't log `s_resp` unless there is a result capture buffer.

sfc-gh-zli: Yes this does change the behavior. So is it necessary for us to always log `s_resp`?

sfc-gh-ema: @sfc-gh-kwagner ? @sfc-gh-vreddy ? What do you think?

sfc-gh-kwagner: Yea we should keep it to always log the response when logging in trace mode. Trace mode shouldn't be enabled unless there is an issue and we need a customer to collect more info for us to fix a bug.

sfc-gh-zli: Makes sense. I'll change it back.

sfc-gh-zli: Yes I agree. I would prefer to make this modification in a separate PR, so that modifications of this PR is more clear.
I created a new JIRA to track this improvement: SNOW-294232.

sfc-gh-kwagner (approved):

lib/client.c
```diff
@@ -1853,18 +1852,14 @@ SF_STATUS STDCALL _snowflake_execute_ex(SF_STMT *sfstmt,
     if (request(sfstmt->connection, &resp, queryURL, url_params,
                 url_paramSize , s_body, NULL,
                 POST_REQUEST_TYPE, &sfstmt->error, is_put_get_command)) {
+        // s_resp will be freed by snowflake_query_result_capture_term
         s_resp = snowflake_cJSON_Print(resp);
```
sfc-gh-vreddy, line 1856: @sfc-gh-zli I would not recommend the trace log of the response as it contains the credentials. And libsnowflakeclient is run on github actions.

PR #271, opened 2021-03-01
SNOW-293305 Add describe only api

sfc-gh-zli:
Currently libsnowflakeclient only support snowflake_execute, which will execute a snowflake statement. In JDBC, apart from statement.execute(), there is another API called statement.describe(), which will execute the query in describe only mode. Describe only means GS will only return some metadata associated with the sql, but won’t actually execute the query.

Statement.describe() is a key dependency for Java Stored Procedures, thus we want to add a new api in libsnowflakeclient, snowflake_describe, for Java SP.

### DIFF
--- a/tests/test_get_describe_only_query_result.c
+++ b/tests/test_get_describe_only_query_result.c
@@ -1,0 +1,69 @@
+/*
+ * Copyright (c) 2018-2020 Snowflake Computing, Inc. All rights reserved.
+ */
+#include "utils/test_setup.h"
+#include <cJSON.h>
+#include <string.h>
+#include <connection.h>
+#include <memory.h>
+#include <error.h>
+
+/**
+ * Tests fetching the describe only query result response directly from SFSTMT
+ * @param unused
+ */
+void test_get_describe_only_query_result(void **unused) {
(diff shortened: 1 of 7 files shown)


sfc-gh-kwagner (approved):
LGTM

tests/test_get_describe_only_query_result.c
```diff
@@ -0,0 +1,69 @@
...
+    json_copy_bool(&success, parsedJSON, "success");
+    assert_int_equal(success, SF_BOOLEAN_TRUE);
+
+    cJSON *data = snowflake_cJSON_GetObjectItem(
+            parsedJSON, "data");
+    assert_int_equal(
+            strlen(snowflake_cJSON_GetObjectItem(data, "queryID")->valuestring) + 1,
+            SF_UUID4_LEN);
+
+    // Make sure that the query is run in describe only mode and the actual result is empty
```
sfc-gh-ema, line 52: Could you add a few more asserts to check that what we need in Snowpark prepare is in this response? I.e., column metadata info.

sfc-gh-zli: ack

sfc-gh-ema (approved):

PR #273, opened 2021-03-05
allow put to a subfolder without ending with /

SimbaGithub:
Simba incident 00298808
When execute put command on AWS, having a subfolder in the stage and not ending it with / will cause failure.
example: put file://c:\puttest\data.tar.gz @stage/subfolder
In the case of Azure, the last charater of the subfolder name is truncated, so the file is uploaded to stage/subfulde
The reason is that the S3Client and AzureClient expect stage location end with /, while in the case of subfolder, it's not.


sfc-gh-vreddy (changes requested):
Test case to make sure its working

SimbaGithub:
> Test case to make sure its working

Test already added test_simple_put_create_subfolder

sfc-gh-vreddy (approved):
LGTM

sfc-gh-vreddy:
http://172.16.20.10:8080/job/LibSnowflakeClient-Mac-Universal-Release-Test/1199/
http://172.16.20.10:8080/job/LibSnowflakeClient-Mac-Universal-Debug-Test/1346/
Mac tests passed.

PR #277, opened 2021-03-16
Fix compilation issues on centos7

sfc-gh-vreddy:
Provide ability to compile on Centos 7

### DIFF
--- a/cpp/SnowflakeAzureClient.cpp
+++ b/cpp/SnowflakeAzureClient.cpp
@@ -38,11 +38,15 @@ SnowflakeAzureClient::SnowflakeAzureClient(StageInfo *stageInfo,
   m_parallel(std::min(parallel, std::thread::hardware_concurrency()))
 {
   const std::string azuresaskey("AZURE_SAS_KEY");
-  char caBundleFile[MAX_PATH] ={0};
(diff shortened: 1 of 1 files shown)


cpp/SnowflakeAzureClient.cpp
```diff
@@ -38,11 +38,15 @@ SnowflakeAzureClient::SnowflakeAzureClient(StageInfo *stageInfo,
...
+  char caBundleFile[MAX_PATH] = {0};
   if(transferConfig && transferConfig->caBundleFile) {
-      int len = std::min((int)strlen(transferConfig->caBundleFile), MAX_PATH - 1);
-      sb_strncpy(caBundleFile, sizeof(caBundleFile), transferConfig->caBundleFile, len);
-      caBundleFile[len]=0;
+      if ( strlen(transferConfig->caBundleFile) > MAX_PATH -1) {
+        throw SnowflakeTransferException(TransferError::INTERNAL_ERROR, 
+            "CA bundle file path too long.");
+      }
+      if (0 != strcpy(caBundleFile, transferConfig->caBundleFile)) {
```
sfc-gh-hchaturvedi, line 47: I think this check might be unnecessary. `strcpy` will return the address of the destination which since is a stack variable will always be non null IMO.

PR #300, opened 2021-03-23
SNOW-303289 Fix error in xp

sfc-gh-zli:
In XP, when the query we issued to GS failed, we want to propagate the error message to UDF server.
Currently in libsfclient, if the execution failed, we free the memory space where we store the GS response. This PR changed this behavior that even if GS returns a failed message, snowflake_execute_with_capture will always stores the response in the passed in result_capture struct. 
This won't cause memory leak, because caller will always call snowflake_result_capture_term to free the memory in result_capture struct.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -2090,13 +2090,9 @@ cleanup:
     if (result_capture == NULL) {
         // If no result capture, we always free s_resp
         SF_FREE(s_resp);
-    } else if (ret != SF_STATUS_SUCCESS) {
-        // If result capture failed, free s_resp.
-        SF_FREE(s_resp);
-        // reset states of result_capture to avoid potential double free of capture_buffer.
-        result_capture->capture_buffer = NULL;
-        result_capture->actual_response_size = 0;
-    } // If result capture succeeded, we don't need to free s_resp.
+    }
+    // Caller should always call result_capture_term to free s_resp,
+    // if result_capture is not NULL
 
     return ret;
 }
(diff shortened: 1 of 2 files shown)


sfc-gh-ema (approved):

lib/client.c
```diff
@@ -2090,13 +2090,9 @@ SF_STATUS STDCALL _snowflake_execute_ex(SF_STMT *sfstmt,
...
-    } else if (ret != SF_STATUS_SUCCESS) {
-        // If result capture failed, free s_resp.
-        SF_FREE(s_resp);
-        // reset states of result_capture to avoid potential double free of capture_buffer.
-        result_capture->capture_buffer = NULL;
-        result_capture->actual_response_size = 0;
-    } // If result capture succeeded, we don't need to free s_resp.
+    }
+    // Caller should always call result_capture_term to free s_resp,
+    // if result_capture is not NULL
```
sfc-gh-kbhatia, line 2095: Do we have positive and negative tests for this API in this repo ?
 If not, we should add them now.
  I would have expected some test to also be updated with this behavior change in API

sfc-gh-zli: ack I changed the test.

sfc-gh-vreddy (changes requested):
@sfc-gh-zli Please add description to the PR. provide relevant context for reviewers.

sfc-gh-vreddy:
@sfc-gh-zli Holding merge until we confirm if this could be issue client drivers.

sfc-gh-vreddy:
@sfc-gh-kwagner There is a behavior change here. Does this affect php-driver? 
Now the onus is on application developer to free the memory. 

This C driver is used by some of our clients, Is there a location where we can add these behavior changes?

sfc-gh-zli:
@sfc-gh-vreddy This change should not affect any other users of libsfclient except xp. XP is the only one that is using the api `snowflake_execute_with_capture`. And we don't change any behavior when `result_capture` is NULL. This PR only affect the behavior when `result_capture` is not NULL.

sfc-gh-vreddy (approved):
got it! Looks good to me

sfc-gh-kbhatia (approved):

PR #303, opened 2021-04-07
AWS s3 private links now get regional URL when relevant parameter is set

sfc-gh-vreddy:
When the parameter FORCE_REGIONAL_S3_ENDPOINTS_FOR_PRESIGNED_URLS is set to true then invoke a different API to generate S3 regional url.
This does NOT break any existing customers. The above parameter will be set true to private link parameters in the AWS us-east-1 regions.


SimbaGithub (commented):
LGTM

PR #305, opened 2021-04-13
Make openssl build compatible with ARM builds

sfc-gh-vreddy:
Makes it compatible with ARM builds


sfc-gh-emusser (approved):
thanks!

sfc-gh-cshi (commented):
Look good to me. But should we have ARM test suite first?

PR #308, opened 2021-04-19
SNOW 345673 Rotated secrets

sfc-gh-hchaturvedi:
Rotated Secrets


sfc-gh-kwagner (approved):
LGTM

sfc-gh-kwagner:
@sfc-gh-hchaturvedi looks like the user is locked. Can you reset the password for the user in the proper accounts and restart the tests?

sfc-gh-hchaturvedi:
> @sfc-gh-hchaturvedi looks like the user is locked. Can you reset the password for the user in the proper accounts and restart the tests?

Resetting the password is going to be a very painful exercise. Let me try unlocking the user

sfc-gh-mkeller:
I think we should be okay to merge this without the tests completing.
It won't get worse than this 😅
But seriously none of the tests fail because of wrong password.
Maybe we could find the reason for the Locking up on Jenkins after

PR #309, opened 2021-04-22
Fixing the bug backoff mechanism

sfc-gh-smirzaei:
With backoff mechanism, we should make sure to return error on timeout otherwise the query is silently skipped (SUCCESS is returned)

Description

Testing


sfc-gh-vreddy (approved):
LGTM

PR #312, opened 2021-04-30
Snow 301725 Build libsnowflakeclient for XP

sfc-gh-vreddy:
Remove gcc49 as it cannot be used to compile snowflakeclient (It cannot compile)
Add gcc92 as XP can use it in near future. 
We will continue to use the gcc52 for ODBC and libsnowflakeclient


sfc-gh-abhatnagar (approved):

sfc-gh-mdionne (approved):
LGTM

PR #313, opened 2021-05-03
Bug fix of arrow support: broken resultset of DML query

SimbaGithub:
DML query should return resultset and that was broken when adding arrow support. Fix it and add test case.
The test case passed on older version of 0.5.10 so the behavior should be the same as before.


sfc-gh-vreddy (approved):
LGTM

sfc-gh-vreddy:
http://172.16.20.10:8080/job/LibSnowflakeClient-Mac-Universal-Release-Test/1456/

PR #315, opened 2021-05-04
snowflakeclient Disable secretDetection for XP

sfc-gh-vreddy:
To improve performance, We want to skip log secret detector when running in XP mode.


sfc-gh-vreddy:
@SimbaGithub Is this good enough to not run secret detector screening on logs when running in XP mode. Give the compilation macro is defined when running in XP mode

SimbaGithub:
> @SimbaGithub Is this good enough to not run secret detector screening on logs when running in XP mode. Give the compilation macro is defined when running in XP mode

Yes, I think so.

SimbaGithub (approved):
LGTM

PR #316, opened 2021-05-04
Do not retry when the file already exists in the stage

sfc-gh-vreddy:
The goal of this PR is to skip file retry if the file uploaded already exists in the Staging area.

### DIFF
--- a/cpp/FileTransferAgent.hpp
+++ b/cpp/FileTransferAgent.hpp
@@ -61,9 +61,8 @@ class RetryContext
      */
     bool isRetryable(RemoteStorageRequestOutcome putStatus)
     {
-        //If putStatus is not SUCCESS and not TOKEN_EXPIRED then put is retryable
-        bool isPutInRetryableState = ((putStatus != RemoteStorageRequestOutcome::SUCCESS) &&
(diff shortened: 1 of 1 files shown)


cpp/FileTransferAgent.hpp
```diff
@@ -61,9 +61,11 @@ class RetryContext
      */
     bool isRetryable(RemoteStorageRequestOutcome putStatus)
     {
-        //If putStatus is not SUCCESS and not TOKEN_EXPIRED then put is retryable
+        //If putStatus is not SUCCESS, not TOKEN_EXPIRED and not SKIP_UPLOAD_FILE then put is retryable
         bool isPutInRetryableState = ((putStatus != RemoteStorageRequestOutcome::SUCCESS) &&
-                (putStatus != RemoteStorageRequestOutcome::TOKEN_EXPIRED)) ;
```
sfc-gh-cshi, line 66: So, seems isPutInRetryableState is equivalent to putStatus == RemoteStorageRequestOutcome ::FAILED

sfc-gh-vreddy: Yes it is, Just wanted to be explicit on when we will be retrying.

sfc-gh-cshi (approved):

PR #317, opened 2021-05-07
Fix crash issue with aws logging when using multiple trheads.

SimbaGithub:
Fix for simba ticket 00313852.

### DIFF
--- a/cpp/logger/SFAwsLogger.cpp
+++ b/cpp/logger/SFAwsLogger.cpp
@@ -22,32 +22,7 @@ namespace
     return data.find(toSearch, pos);
   }
 
-  struct awslog_mutex
-  {
-    awslog_mutex()
-    {
-      _critical_section_init(&logMutex);
-    }
-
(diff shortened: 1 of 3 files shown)


cpp/logger/SFAwsLogger.cpp
```diff
@@ -21,6 +21,33 @@ namespace
     // Find sub string in given string
     return data.find(toSearch, pos);
   }
+
+  struct awslog_mutex
```
sfc-gh-cshi, line 25: why do we have two structs for awslog_mutex and awssdk_mutex, These definitions look pretty common to me.

SimbaGithub: > why do we have two structs for awslog_mutex and awssdk_mutex, These definitions look pretty common to me and should be deduplicated

Fixed to use common definition.

sfc-gh-cshi (approved):

PR #319, opened 2021-05-12
Update curl version to 7.74.0

SimbaGithub:


sfc-gh-hchaturvedi (approved):
LGTM. I was mostly concerned about the OCSP specific changes. Looks like proper CURLOPT specific to snowflake have been ported over. I also see that the sf_ocsp file hasn't change and has been ported over. I assume no changes have been made in the OCSP module.

PR #325, opened 2021-06-30
Update for FIPS support

sfc-gh-hchaturvedi:
This change only affects PUT / GET functionality in Libsnowflakeclient and ODBC Driver. 

Till now the driver could only support default stage URLs or region based stage URLs. However while using Gov Cloud, there are certain security requirements that only few deployments and stages can adhere to at the moment. These stages / deployments are FIPS enabled. To ensure that the driver only connects to stages that are FIPS enabled, the GS would send to the driver exact URL that it needs to connect to. 

This fix updates the driver to accept the new information and use that instead of the default S3 URL.

### DIFF
--- a/cpp/SnowflakeS3Client.cpp
+++ b/cpp/SnowflakeS3Client.cpp
@@ -111,7 +111,7 @@ SnowflakeS3Client::SnowflakeS3Client(StageInfo *stageInfo,
   // FIPS mode check
   if (!(m_stageInfo->endPoint.empty())) {
     // FIPS mode is enabled, use the endpoint provided by GS directly
-    clientConfiguration.endPointOverride = Aws::String(stageInfo->endPoint);
+    clientConfiguration.endpointOverride = Aws::String(stageInfo->endPoint);
   } else if (transferConfig != nullptr && transferConfig->useS3regionalUrl) {
     clientConfiguration.endpointOverride = Aws::String("s3.")
         + Aws::String(clientConfiguration.region)
--- a/cpp/SnowflakeS3Client.cpp
+++ b/cpp/SnowflakeS3Client.cpp
@@ -107,8 +107,12 @@ SnowflakeS3Client::SnowflakeS3Client(StageInfo *stageInfo,
   clientConfiguration.caFile = caFile;
   clientConfiguration.requestTimeoutMs = 40000;
   clientConfiguration.connectTimeoutMs = 30000;
-  if(transferConfig != nullptr && transferConfig->useS3regionalUrl)
-  {
+
+  // FIPS mode check
+  if (stageInfo->endPoint != nullptr) {
+    // FIPS mode is enabled, use the endpoint provided by GS directly
+    clientConfiguration.endPointOverride = Aws::String(stageInfo->endPoint);
+  } else if (transferConfig != nullptr && transferConfig->useS3regionalUrl) {
     clientConfiguration.endpointOverride = Aws::String("s3.")
         + Aws::String(clientConfiguration.region)
         + Aws::String(".amazonaws.com");
--- a/cpp/SnowflakeS3Client.cpp
+++ b/cpp/SnowflakeS3Client.cpp
@@ -599,7 +599,7 @@ void SnowflakeS3Client::setMaxRetries(unsigned int maxRetries)
   m_maxRetries = maxRetries;
 }
 
-char * SnowflakeS3Client::GetClientConfigStageEndpoint()
+const char * SnowflakeS3Client::GetClientConfigStageEndpoint()
 {
   return clientConfiguration.endpointOverride.c_str();
 }
--- a/cpp/SnowflakeS3Client.cpp
+++ b/cpp/SnowflakeS3Client.cpp
@@ -109,7 +109,7 @@ SnowflakeS3Client::SnowflakeS3Client(StageInfo *stageInfo,
   clientConfiguration.connectTimeoutMs = 30000;
 
   // FIPS mode check
-  if (stageInfo->endPoint != nullptr) {
+  if (!(m_stageInfo->endPoint.empty())) {
     // FIPS mode is enabled, use the endpoint provided by GS directly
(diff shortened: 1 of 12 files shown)


sfc-gh-hchaturvedi:
@sfc-gh-cshi / @SimbaGithub can you please take a look at this PR?

SimbaGithub (commented):
The code change looks good to me. Do we know what caused the test failures?

SimbaGithub (changes requested):
Also I think test case needs to be add for this feature. Is there any option to force the server send endpoint for aws?

sfc-gh-hchaturvedi:
> The code change looks good to me. Do we know what caused the test failures?

@SimbaGithub Access Violation for AWS on Windows.  Not sure what is causing that.

SimbaGithub:
> > The code change looks good to me. Do we know what caused the test failures?
> 
> @SimbaGithub Access Violation for AWS on Windows. Not sure what is causing that.

@sfc-gh-hchaturvedi I think that might because you added free for endPoint and storageAccount while those pointers could be not initialized for case of AWS? Please ensure those pointer is set to null when server doesn't send them in response. That's just my guess, please try reproduce the failure in you local environment and fix it.

SimbaGithub:
ODBC has it's own implementation of IStatementPutGet so won't be affected by this PR. 
https://github.com/snowflakedb/snowflake-odbc/blob/master/Source/Platform/ODBCStatementPutGet.cpp
Please make another PR on ODBC side or make a Simba support ticket for feature request so I can work on it.

sfc-gh-hchaturvedi:
@SimbaGithub I think ODBC too would get affected by this PR. The meat of the change is in the implementation of the S3Client which I think ODBC refers to libsnowflakeclient for. Please correct me if I am wrong. ODBC ofcourse would need to be updated to support this change.

SimbaGithub:
@sfc-gh-hchaturvedi Yes, the change in cpp/SnowflakeS3Client.cpp would affect ODBC as well. I meant https://github.com/snowflakedb/snowflake-odbc/blob/master/Source/Platform/ODBCStatementPutGet.cpp needs to be updated on ODBC side to parse the extra information in the response from server side.

cpp/SnowflakeS3Client.cpp
```diff
@@ -107,8 +107,12 @@ SnowflakeS3Client::SnowflakeS3Client(StageInfo *stageInfo,
   clientConfiguration.caFile = caFile;
   clientConfiguration.requestTimeoutMs = 40000;
   clientConfiguration.connectTimeoutMs = 30000;
-  if(transferConfig != nullptr && transferConfig->useS3regionalUrl)
-  {
+
+  // FIPS mode check
+  if (!(m_stageInfo->endPoint.empty())) {
```
sfc-gh-kwagner, line 112: Don't you also need to check to make sure `endPoint` isn't null before calling `.empty()` ?

cpp/StatementPutGet.cpp
```diff
@@ -74,6 +74,10 @@ bool StatementPutGet::parsePutGetCommand(std::string *sql,
   {
     putGetParseResponse->stageInfo.stageType = StageType::S3;
     putGetParseResponse->stageInfo.region = response->stage_info->region;
+    // FIPS Support
+    if (response->stage_info->endPoint != NULL) {
```
sfc-gh-kwagner, line 78: Should you also check that endPoint isn't empty?

lib/client_int.h
```diff
@@ -76,7 +76,7 @@ typedef struct SF_STAGE_INFO {
   char *path;
   char *region;
   char *storageAccount; // For Azure only
-  char *endPoint; //For Azure only.
+  char *endPoint; //For FIPS support
```
sfc-gh-kwagner, line 79: nit: mention that this is for azure and fips support (like comment in PutGetParseResponse.hpp)

lib/connection.c
```diff
@@ -670,10 +670,19 @@ json_copy_string(char **dest, cJSON *data, const char *item) {
     size_t blob_size;
     cJSON *blob = snowflake_cJSON_GetObjectItem(data, item);
     if (!blob) {
+        // We don't check the return status everywhere
+        // make sure that the value is set to NULL to enable
+        // NULL checks.
+        SF_FREE(*dest);
+        *dest = NULL;
```
sfc-gh-kwagner, line 677: The SF_FREE macro already sets *dest to NULL after it frees the data, so this line is redundant

sfc-gh-kwagner:
@sfc-gh-hchaturvedi added some comments

cpp/SnowflakeS3Client.cpp
sfc-gh-hchaturvedi: I don't think so - [Here's](https://github.com/snowflakedb/libsnowflakeclient/blob/efde8ea8df69bc4dbe7d7e14c93d331be9879c28/include/snowflake/PutGetParseResponse.hpp#L62) the definition of `string StageInfo` which `endPoint` is a member of.  `endPoint` itself isn't a pointer.

cpp/StatementPutGet.cpp
sfc-gh-hchaturvedi: I can put a check there but it would be redundant because the empty check is present in SnowflakeS3Client.cpp.

sfc-gh-hchaturvedi:
> Also I think test case needs to be add for this feature. Is there any option to force the server send endpoint for aws?

@SimbaGithub I have been working on this on and off but there is no clean way to write a unit test for this. We don't have an E2E test account to test it out for regressions on a regular basis. Going to reach out off line to discuss some potential options for testing this.

SimbaGithub:
Hi @sfc-gh-hchaturvedi,
The unit test case makes sense to me. Please solve the build issue and make the test case pass.
-Harry

sfc-gh-hchaturvedi:
@SimbaGithub The tests are fine now. Please review.

SimbaGithub (approved):
LGTM

PR #326, opened 2021-06-30
SNOW-373485 Adding more logging around createStmt and execute

sfc-gh-wdai:


lib/client.c
```diff
@@ -1965,6 +1969,7 @@ SF_STATUS STDCALL _snowflake_execute_ex(SF_STMT *sfstmt,
                     "localLocation");
 
             } else {
+                // todo: Deadlock possible?
```
sfc-gh-wdai, line 1972: I do not think it could cause deadlock here. But still ping here in case I forgot some use scenarios.

lib/connection.c
```diff
@@ -349,6 +352,7 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
         }
 
         if (strcmp(query_code, SESSION_TOKEN_EXPIRED_CODE) == 0) {
+            log_debug("Session token expired");
```
sfc-gh-smirzaei, line 355: TODO: change to log_info

lib/connection.c
```diff
@@ -420,13 +425,30 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
                                     SF_SQLSTATE_UNABLE_TO_CONNECT);
                 break;
             }
+
+            if(counter_for_code == 1 && sf->log_spec_info){
```
sfc-gh-smirzaei, line 429: shouldn't `counter_for_code == 1` be `counter_for_code == 1000`?

lib/connection.c
```diff
@@ -420,13 +425,30 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
...
+            counter_for_code++;
         }
 
         if (stop) {
             break;
         }
 
         ret = SF_BOOLEAN_TRUE;
+        counter++;
+        if(counter == 1000){
```
sfc-gh-smirzaei, line 448: Let's add the param check here as well (similar to what I suggested for previous one)

sfc-gh-smirzaei (commented):
Left some comments.
Also, let change the couple of places that we were already using `log_info` to `log_debug` (since those places are almost useless). And then change the default logging level for all customers to `INFO` by default.

lib/connection.c
sfc-gh-smirzaei: Let's do it as following so that we don't keep incrementing `counter_for_code` indefinitely:
```
if (sf->log_spec_info){
  if(counter_for_code == 1000) {
    .... // logging
    }
  counter_for_code++;
}
```

sfc-gh-wdai: I changed to 1 to let me able to test it locally, will revert it back after some local tests

lib/client.c
```diff
@@ -726,7 +727,7 @@ SF_STATUS STDCALL snowflake_connect(SF_CONNECT *sf) {
     char os_version[128];
     sf_os_version(os_version, sizeof(os_version));
 
-    log_info("Snowflake C/C++ API: %s, OS: %s, OS Version: %s",
+    log_debug("Snowflake C/C++ API: %s, OS: %s, OS Version: %s",
```
sfc-gh-smirzaei, line 730: NIT: this can stay `log_info`

include/snowflake/client.h
```diff
@@ -265,6 +266,7 @@ typedef struct SF_CONNECT {
 
     char *passcode;
     sf_bool passcode_in_password;
+    sf_bool log_spec_info;
```
sfc-gh-smirzaei, line 269: NIT: I'm not sure if the name for `log_spec_info` is clear what it does.

sfc-gh-smirzaei (approved):
Besides some minor comments, LGTM, thanks.
So I assume now we have logging at different stages of executing a child job, i.e., beginning, middle and end, right?

include/snowflake/client.h
sfc-gh-wdai: I think this attributes will be used widely when we plan to add more logs in libsnowflakeClient. Therefore, I choose this name means we need to add some specific logs in libsnowflakeClient. Please help me find some more meaningful name :)

sfc-gh-wdai:
> Besides some minor comments, LGTM, thanks.
> So I assume now we have logging at different stages of executing a child job, i.e., beginning, middle and end, right?

Yah

include/snowflake/client.h
sfc-gh-smirzaei: Hopefully we won't need to add attributes like this once libSfClient is ported to XP. So we can go with something more specific here, a name like `log_query_exec_steps_info` or anything that you can think of.

lib/client.c
```diff
@@ -1859,6 +1869,7 @@ SF_STATUS STDCALL _snowflake_execute_ex(SF_STMT *sfstmt,
     if (request(sfstmt->connection, &resp, queryURL, url_params,
                 url_paramSize , s_body, NULL,
                 POST_REQUEST_TYPE, &sfstmt->error, is_put_get_command)) {
+        log_info("Query execution receives response");
```
sfc-gh-smirzaei, line 1872: I'm trying to be super safe and have all bases covered, with that in mind is it possible that the execution may get stuck anywhere between this point and `log_info("Query execution done");`?

sfc-gh-wdai: The only space I could image is that I point out in previous comment for the dead lock issue. But I could not image the case, but we still could add logs as check-points. Please check my first comments.

lib/client.c
sfc-gh-wdai: @sfc-gh-smirzaei To make it super safe, we could add some logs here as check-point. In case it causes deadlock(interrupt before the lock is released)

PR #329, opened 2021-08-03
fast fail and max retries for get

SimbaGithub:
simba ticket 00342243. Add fast fail and max retries for get to be inline with put. Fix inconsistent behavior of put for large and small files as well.
Another PR will be made on ODBC side to pass the setting.


cpp/FileTransferAgent.cpp
```diff
@@ -259,12 +272,16 @@ void Snowflake::Client::FileTransferAgent::upload(string *command)
...
-        std::string localCommand = *command;
-        getPresignedUrlForUploading(m_smallFilesMeta[i], localCommand);
+        for (size_t i = 0; i < m_smallFilesMeta.size(); i++)
+        {
+          std::string localCommand = *command;
+          getPresignedUrlForUploading(m_smallFilesMeta[i], localCommand);
+        }
       }
     }
     uploadFilesInParallel(command);
```
sfc-gh-hchaturvedi, line 287: If fast fail is enabled and we've had a failure it would mean that we won't call `getPresignedUrlForUploading` from the logic between line 278 and 284. If that is the case should we even call `uploadFilesInParallel` at all?

SimbaGithub: In such case, calling uploadFilesInParallel() is just for setting transfer outcome to SKIP_UPLOAD_FILE for following files. We could add a loop here to do that but I think it doesn't hurt to reuse the existing logic.

cpp/FileTransferAgent.cpp
```diff
@@ -610,21 +627,56 @@ void Snowflake::Client::FileTransferAgent::download(string *command)
     for (size_t i=0; i<m_largeFilesMeta.size(); i++)
     {
       m_executionResults->SetFileMetadata(&m_largeFilesMeta[i], i);
+      CXX_LOG_DEBUG("Putget serial large file download, %s file",
+                    m_largeFilesMeta[i].srcFileName.c_str());
+      if (isGetFastFailEnabled() && !m_failedTransfers.empty())
+      {
+        m_executionResults->SetTransferOutCome(RemoteStorageRequestOutcome::FAILED, i);
+        CXX_LOG_DEBUG("Sequential download, get fast fail enabled, Skipping file.");
+        continue;
```
sfc-gh-cshi, line 636: when fast fail is enabled, why do we need to continue here? These two conditions isGetFastFailEnabled && m_failedTransfers do not look like will change?

SimbaGithub: Copied the same behavior from uploadFilesInParallel(). Continue to set transfer outcome for all following files and also have log for each file so we know which files are downloaded, which one failed and which are skipped afterwards.

sfc-gh-hchaturvedi (approved):
Approved on @sfc-gh-cshi 's behalf.

PR #330, opened 2021-08-09
Simplify the log in response json

sfc-gh-wdai:
Issue: Snowhouse has the limitation of the log message size. Therefore, we need to keep the useful information and delete inessential attributes in response json. 
On the other hand, I just check the original text if it exist the "code" attribute to triage the "miss code" issue happened in libsfclient or GS side.

Add a parameter to avoid string search for each query, I plan to continue using the previous parameter because the issue still happened in query execution step( which the parameter name is make sense). Also since we are roll out the port libsfclient to XP, I do not want to add too much temporary parameter. cc @sfc-gh-kwang Do you think we need to make the same change in your branch(XP part of libsfclient).

### DIFF
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -330,7 +330,7 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
         if ((json_error = json_copy_string_no_alloc(query_code, *json, "code",
                                                     QUERYCODE_LEN)) !=
             SF_JSON_ERROR_NONE &&
-            json_error != SF_JSON_ERROR_ITEM_NULL)  {
+            json_error != SF_JSON_ERROR_ITEM_NULL) {
             //Log the useful response information
             create_json_resp_log(json);
             JSON_ERROR_MSG(json_error, error_msg, "Query code");
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -327,10 +327,10 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
             // Error is set in the perform function
             break;
         }
-        sf_bool flag = SF_BOOLEAN_TRUE;
-        if (flag) {
-            json_error = json_copy_string_no_alloc(query_code, *json, "code",
-                                                   QUERYCODE_LEN);
+        if ((json_error = json_copy_string_no_alloc(query_code, *json, "code",
+                                                    QUERYCODE_LEN)) !=
+            SF_JSON_ERROR_NONE &&
+            json_error != SF_JSON_ERROR_ITEM_NULL)  {
             //modify the new Json since we need to keep the original json information
             //cJSON *newJson = snowflake_cJSON_Duplicate(*json, cJSON_True);
             //const char* del = "rowset";
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -1140,5 +1140,4 @@ void STDCALL create_json_resp_log(cJSON **json){
     log_error("Missing query code:\n %s", snowflake_cJSON_Print(newJson));
     //free the memory
     snowflake_cJSON_free(newJson);
-    free(dels);
 }
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -332,13 +332,13 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
             json_error = json_copy_string_no_alloc(query_code, *json, "code",
                                                    QUERYCODE_LEN);
             //modify the new Json since we need to keep the original json information
-            cJSON *newJson = snowflake_cJSON_Duplicate(*json, cJSON_True);
-            const char* del = "rowset";
+            //cJSON *newJson = snowflake_cJSON_Duplicate(*json, cJSON_True);
+            //const char* del = "rowset";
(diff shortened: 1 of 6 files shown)


lib/connection.c
```diff
@@ -1142,3 +1124,20 @@ void STDCALL sf_header_destroy(SF_HEADER *sf_header) {
     curl_slist_free_all(sf_header->header);
     SF_FREE(sf_header);
 }
+
+//Help function to log the useful response json data
+void STDCALL create_json_resp_log(cJSON **json){
```
sfc-gh-wdai, line 1129: @sfc-gh-kwagner Could you help me take a look about the code standard since I am not familiar with it. Not sure if I need to add STDCALL or not. Thanks!

lib/http_perform.c
```diff
@@ -242,7 +242,6 @@ sf_bool STDCALL http_perform(CURL *curl,
             log_error("Failed to set writer [%s]", curl_easy_strerror(res));
             break;
         }
-
```
sfc-gh-kwang, line 245: nit: add back removed line?

lib/http_perform.c
```diff
@@ -396,6 +395,10 @@ sf_bool STDCALL http_perform(CURL *curl,
 
     // We were successful so parse JSON from text
     if (ret) {
+        //Check if the "code" attribute exist in the response texts
+        if(!strstr(buffer.buffer, "code")){
```
sfc-gh-kwang, line 399: Should we search for `"code"` (with the double quotes) to eliminate false positives (ie. `code` could appear in the rowset response)

lib/connection.c
sfc-gh-kwagner: You need STDCALL for when people compile this on windows

lib/connection.c
```diff
@@ -331,12 +332,9 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
...
             json_error != SF_JSON_ERROR_ITEM_NULL) {
-            //modify the new Json since we need to keep the original json information
-            cJSON *newJson = snowflake_cJSON_Duplicate(*json, cJSON_True);
-            const char* del = "rowset";
-            //delete the sensitive information in case it leaks to customer
-            snowflake_cJSON_DeleteItemFromObject(newJson, del, cJSON_True);
-            log_error("Missing query code:\n %s", snowflake_cJSON_Print(newJson));
+            //Log the useful response information
+            cJSON *newJson = create_json_resp_log(json);
+            log_error("Missing query code in post call:\n %s", snowflake_cJSON_Print(newJson));
```
sfc-gh-wdai, line 337: I need to distinguish the log so I just move out this part from function.

lib/connection.c
```diff
@@ -411,12 +409,9 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
...
               json_error != SF_JSON_ERROR_ITEM_NULL) {
-                //modify the new Json since we need to keep the original json information
-                cJSON *newJson = snowflake_cJSON_Duplicate(*json, cJSON_True);
-                const char* del = "rowset";
-                //delete the sensitive information in case it leaks to customer
-                snowflake_cJSON_DeleteItemFromObject(newJson, del, cJSON_True);
-                log_error("Query code missing: %s", snowflake_cJSON_Print(newJson));
+                //Log the useful response information
+                cJSON *newJson = create_json_resp_log(json);
+                log_error("Missing query code in post call:\n %s", snowflake_cJSON_Print(newJson));
```
sfc-gh-kwang, line 414: This has the same log message as the above case, is this intentional?

lib/logger.c
```diff
@@ -103,16 +103,17 @@ void log_log(int level, const char *file, int line, const char *ns,
       // va_list can only be consumed once. Make a copy so we can use it again
       va_list copy;
       va_copy(copy, args);
-
+      size_t linedFmt_length = strlen(fmt) +
```
sfc-gh-kwang, line 106: Why are we changing the code here and why are we allocating on heap now as opposed to on the stack?

lib/connection.c
```diff
@@ -427,10 +422,8 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
             }
             if(sf->log_query_exec_steps_info){
                 if(counter_for_code == 1000){
-                    cJSON *newJson = snowflake_cJSON_Duplicate(*json, cJSON_True);
-                    const char* del = "rowset";
-                    //delete the sensitive information in case it leaks to customer
-                    snowflake_cJSON_DeleteItemFromObject(newJson, del, cJSON_True);
+                    //Log the useful response information
+                    cJSON *newJson = create_json_resp_log(json);
                     log_info("Query continue in progress, query information: %s", snowflake_cJSON_Print(newJson));
```
sfc-gh-kwang, line 427: Why is this logging on INFO level whereas the rest is logging on ERROR level?

sfc-gh-kwang:
Do we need to update the `http_perform` interface in `mock_http_perform.h` for the unit tests to work?

sfc-gh-kwang (approved):
lgtm

PR #332, opened 2021-08-12
Fix crash issue with OCSP

SimbaGithub:
Fix crash issue with OCSP reported by Simba's customer.
ocsp_log_data is freed inside generateOCSPTelemetryData() which still could be used outside in checkCertOCSP() for the next cert.

### DIFF
--- a/deps/curl-7.74.0/lib/vtls/sf_ocsp.c
+++ b/deps/curl-7.74.0/lib/vtls/sf_ocsp.c
@@ -1533,8 +1533,6 @@ static char * generateOCSPTelemetryData(SF_OTD *ocsp_log)
   setOOBeventdata(OOBEVENTNAME, "OCSPException", 0);
   setOOBeventdata(URGENCY, NULL, 1);
   oobevent = prepareOOBevent(ocsp_log);
-  FREE_OCSP_LOG(ocsp_log);
-  ocsp_log = NULL;
   return oobevent;
 }
 /**
@@ -2124,7 +2122,6 @@ SF_PUBLIC(CURLcode) checkCertOCSP(struct connectdata *conn, STACK_OF(X509) *ch, 
   CURLcode rs = CURLE_OK;
   char *ocsp_fail_open_env = getenv("SF_OCSP_FAIL_OPEN"); // Testing Only
   SF_FAILOPEN_STATUS ocsp_fail_open = ENABLED;
-  SF_OTD *ocsp_log_data = (SF_OTD *)calloc(1, sizeof(SF_OTD));
 
 
 // These end points are Out of band telemetry end points.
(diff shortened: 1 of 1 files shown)


deps/curl-7.74.0/lib/vtls/sf_ocsp.c
```diff
@@ -1533,8 +1533,6 @@ static char * generateOCSPTelemetryData(SF_OTD *ocsp_log)
   setOOBeventdata(OOBEVENTNAME, "OCSPException", 0);
   setOOBeventdata(URGENCY, NULL, 1);
   oobevent = prepareOOBevent(ocsp_log);
-  FREE_OCSP_LOG(ocsp_log);
-  ocsp_log = NULL;
```
sfc-gh-hchaturvedi, line 1537: if we're removing this from here .. I'd like some of the fields of the log data be reset to make sure we are not sending duplicate incorrect data for events related to other certificates.

sfc-gh-hchaturvedi: The following fields should be reset - 

`event_type`
`event_sub_type`
`sfc_peer_host`
`cert_id`
`ocsp_req_b64`
`ocsp_responder_url`
`error_msg`

SimbaGithub: I understand it's important to send clear log data but the free didn't do that either.  The free here just caused crash and did nothing more. ocsp_log is an input parameter so set it to NULL here won't change the value hold by the caller and the invalid pointer will still be used. Sometime invalid pointer still could be used without crash but in such case the content won't be cleared so any duplicate incorrect data would be still there. That means if we don't get duplicate incorrect data before, we won't get it after the fix either.
To make sure any duplicate incorrect data is cleared we need to check the logic in caller like checkCertOCSP() and make change there, which I think the priority is much lower than the crash issue. Please let me know if there is any duplicate incorrect data received on server side and make a simba ticket for that, otherwise I think the existing logic might work well already.

sfc-gh-hchaturvedi: I don't see the reason to split it up especially since it's a pretty straightforward fix. I think a helper function that clears the relevant fields right after we call `prepareOOBevent` should suffice since `GenerateOCSPTelemetryData` is called only once per cert. 

There is a _lot_ of telemetry data in our backend and it might be hard to find the anomalies. I don't see any harm in being a little careful about this.

sfc-gh-hchaturvedi: Also ocsp_log is a pointer that's being sent around the functions so I don't understand what you mean by `ocsp_log is an input parameter so set it to NULL here won't change the value hold by the caller`

SimbaGithub: > I don't see the reason to split it up especially since it's a pretty straightforward fix. I think a helper function that clears the relevant fields right after we call `prepareOOBevent` should suffice since `GenerateOCSPTelemetryData` is called only once per cert.

Added reset for log data. I was worried about if the data intentionally to be reused by the caller could be cleared and I don't have time to go through entire logic to confirm that. I think the reset actually better to be added in prepareOOBevent() if we want to clear it after each telemetry sending but that would need a version bump of oob so let's put it here for now.

sfc-gh-hchaturvedi (approved):
LGTM!

PR #335, opened 2021-08-16
remove potential memory leak

SimbaGithub:
ocsp test case fails on memory leak when applying crash issue fix in curl 7.74.0 [PR332](https://github.com/snowflakedb/libsnowflakeclient/pull/332). It could be a false positive but remove the calloc just in case.


deps/curl-7.74.0/lib/vtls/sf_ocsp.c
```diff
@@ -2175,13 +2176,13 @@ SF_PUBLIC(CURLcode) checkCertOCSP(struct connectdata *conn, STACK_OF(X509) *ch,
     ocsp_fail_open = DISABLED;
   }
 
-  sf_otd_set_insecure_mode(1, ocsp_log_data);
+  sf_otd_set_insecure_mode(1, &ocsp_log_data);
```
sfc-gh-hchaturvedi, line 2179: This should be set to 0. This fix is unrelated to your change but let's just fix it. @SimbaGithub

sfc-gh-hchaturvedi (approved):
Apart from the single comment everything LGTM. Please make the update requested before merging.

PR #337, opened 2021-08-31
Adjust the log in libsfclient

sfc-gh-wdai:
For this PR, I did two things:
1. Add log around the `curl_easy_perform` function because this is a function that may cause the query stuck in libsfclient, we should add log in advance so when the issue happens again, we could direct check it.
2. Delete some useless logs, it should be my fault, I made a small mistake in the curl_get_call and curl_post_call, the outside do while loop, the condition is while(0) so it will not cause an infinite loop here.

Next step:
I hope this is the last time to adjust the log in the third part of libsfclient. I will update the related log in XP's libsfclient


sfc-gh-kwang (approved):
lgtm

PR #338, opened 2021-09-06
SNOW-395243 Support application in connection settings

SimbaGithub:
Simba ticket 00349937. Allow user set the value for CLIENT_ENVIRONMENT::APPLICATION in the logging request via the 'application' connection setting. PHP connector is using libsnowflakeclient so we need the change here.

### DIFF
--- a/lib/client.c
+++ b/lib/client.c
@@ -378,7 +378,7 @@ static void STDCALL log_term() {
 SF_STATUS STDCALL
 _snowflake_check_connection_parameters(SF_CONNECT *sf) {
     if (is_string_empty(sf->account)) {
-        // Invalid connection
+        // Invalid account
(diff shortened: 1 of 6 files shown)


lib/client.c
```diff
@@ -404,6 +410,17 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
         return SF_STATUS_ERROR_GENERAL;
     }
 
+    if (SF_BOOLEAN_FALSE == validate_application(sf->application)) {
+        // Invalid connection
```
sfc-gh-kwagner, line 414: nit: Should say its an invalid application name that was passed in

SimbaGithub: @sfc-gh-kwagner done.

sfc-gh-kwagner (approved):
LGTM other than minor comment

PR #342, opened 2021-10-20
Move to curl 7.78.0

SimbaGithub:
Upgrade curl to 7.78.0 as MSFT requested.
Port change for OCSP bases on commit when doing 7.74.0 https://github.com/snowflakedb/libsnowflakeclient/pull/319/commits/8ee8c6051705622b091437ff19aa6a7d1e329a94
Made minor changes on function interfaces due to interface change in 7.78.0, struct Curl_easy* data is removed from struct connectdata and has to be passed separately.

The list of files I changed:
deps/curl-7.78.0/include/curl/curl.h
deps/curl-7.78.0/lib/Makefile.in
deps/curl-7.78.0/lib/Makefile.inc
deps/curl-7.78.0/lib/curl_addrinfo.c
deps/curl-7.78.0/lib/setopt.c
deps/curl-7.78.0/lib/url.c
deps/curl-7.78.0/lib/urldata.h
deps/curl-7.78.0/lib/vtls/openssl.c
deps/curl-7.78.0/lib/vtls/sf_cJSON.c
deps/curl-7.78.0/lib/vtls/sf_cJSON.h
deps/curl-7.78.0/lib/vtls/sf_ocsp.c
deps/curl-7.78.0/lib/vtls/sf_ocsp.h
deps/curl-7.78.0/src/tool_hugehelp.c
deps/curl-7.78.0/winbuild/MakefileBuild.vc
scripts/build_curl.bat
scripts/build_curl.sh
scripts/build_oob.bat
scripts/build_oob.sh
scripts/build_azuresdk.sh
.github/workflows/build-test.yml
tests/CMakeLists.txt

It would be easier to review with commits especially for these two:
- port OCSP changes https://github.com/snowflakedb/libsnowflakeclient/pull/342/commits/85b03f64c2d67a9de647af51da3ebf32dcc91bfc
- remove newline from format string of curl log https://github.com/snowflakedb/libsnowflakeclient/pull/342/commits/8ac2e4453b8145af17749048fbd8edc785b82399


SimbaGithub:
@sfc-gh-hchaturvedi Added change details for review in PR description. Please take a look.

deps/curl-7.78.0/lib/vtls/sf_cJSON.c
```diff
@@ -0,0 +1,2938 @@
+/*
```
sfc-gh-hchaturvedi, line 1: @SimbaGithub are we keeping track of updates happening on the cJSON source ?  Do we have any checks and balances to ensure that the cJson fork here is in sync with the original? If not I'd suggest we come up with a plan to keep the two in sync or atleast test with the latest changes. There are not many in the past 1 year but it might be worthwhile to consider the delta.

sfc-gh-hchaturvedi (approved):
My only concern is about our plan to update `sf_cJSON`, I don't think we have one right now. Though I don't think it is high priority right now, but we need to have a plan. The fact that `sf_cJSON` is not an external dependency, wherein we've just copied the code, version based tools like Whitesource won't work in pointing out vulnerability in existing code base. Either we always pull the latest `cJSON` with every release or we do it periodically diligently.

deps/curl-7.78.0/lib/vtls/sf_cJSON.c
SimbaGithub: @sfc-gh-hchaturvedi So far we just copy sf_cJSON from older curl to the new one. I agree it's time to sync with the original because it has been 3 years since we picked up version 1.7.4 and now the original is 1.7.15. We can discuss the plan on our sync meeting. Given the fact that there are only bug fixes and performance improvements and no major changes in the past 3 years, for now I think after sync to 1.7.15, in the future we can add sync cJSON as a part of curl update task so we can do it whenever we update to a new curl version.

PR #344, opened 2021-10-26
Initialize sf_error_struct

sfc-gh-wdai:


sfc-gh-dprzybysz:
libsnowflakeclient is not used in XP anymore

PR #352, opened 2021-12-17
SNOW-514869 loading OCSP cache file always fail

SimbaGithub:
The size of buffer for loading OCSP cache file is hardcoded as 24KB which is smaller than usual file size (150KB) and causes the loading always fail.
Tested with unit_ocsp_test manually, all cases passed and confirmed "OCSP cache file was successfully loaded" output in curl log.


sfc-gh-cshi (approved):

sfc-gh-hchaturvedi (commented):
LGTM

PR #353, opened 2021-12-18
Update cJSON to 1.7.15

SimbaGithub:
Update cJSON in curl (was 7.1.4) and libsnowflakeclient (was 7.1.12) to 7.1.15
Compared existing code with original cJSON code to ensure all custom changes are applied.
Also compared changes in original code to confirm any further custom change is needed there.
Make separate commit for custom changes so we can reference to it when upgrading to a new version next time.
Remember to bump up minor version of both curl and libsnowflakeclient in the next release.

### DIFF
--- a/lib/cJSON.c
+++ b/lib/cJSON.c
@@ -57,6 +57,7 @@
 #endif
 
 #include "cJSON.h"
+#include "snowflake/Simba_CRTFunctionSafe.h"
 
 /* define our own boolean type */
 #ifdef true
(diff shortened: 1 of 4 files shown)


lib/cJSON.c
```diff
@@ -91,24 +92,24 @@ typedef struct {
 } error;
 static error global_error = { NULL, 0 };
 
-CJSON_PUBLIC(const char *) cJSON_GetErrorPtr(void)
```
sfc-gh-cshi: What about we keep the original function names? While I understand this could be an already implemented change working in previous cJSON.

SimbaGithub: Yes, this is an already implemented change working in previous cJSON. The reason for that change was to avoid linking conflict: https://github.com/snowflakedb/libsnowflakeclient/pull/73
I think the reason might be that we have another copy of cJSON in curl, without changing the original function names, we would have duplicated implementation for the same functions when linking to both curl and libsnowflakeclient libs.

sfc-gh-cshi (approved):

PR #358, opened 2022-01-27
Arm64 Linux and OSX M1 support

SimbaGithub:
Simba ticket 00364226, 00372658
change build scripts and code to fix build issues on arm64 Linux and Apple M1.
Currently we don't have test/build environment of arm64 or M1 available for github action. So this change allows to manually build & test on arm64 and M1 environment, also keep the compatibility with other platforms the existing build & test on github action still could pass.
Manually build & tested on Ubuntu 20.04+aarch64+gcc9 and OSX11+M1+XCode12.

Some changes need explanation:
1. Changed JWT interface as the original implementation returns buffer of local variable, which cause build error on M1, and could potentially crash the application on all platforms.
2. Disable azure test on aarch64 and M1 because it's using old version of catch.hpp which uses asm code that can't be build on arm. ODBC has similar issue which needs to update catch.hpp to version 2.13.3 above to fix.


tests/utils/test_setup.c
```diff
@@ -27,9 +27,9 @@ void initialize_test(sf_bool debug) {
     if (!test_log_dir) {
         char tmp_dir[4096];
         sf_get_tmp_dir(tmp_dir);
-        snprintf(PERFORMANCE_TEST_RESULTS_PATH, 4096, "%s%s%s", tmp_dir, sep, perf_test_fn);
+        snprintf(PERFORMANCE_TEST_RESULTS_PATH, 5000, "%s%s%s", tmp_dir, sep, perf_test_fn);
```
sfc-gh-cshi, line 30: If the max path len is changed to 5000, should the 'tmp_dir' change too? A constant for this length might be better.

tests/CMakeLists.txt
```diff
@@ -100,6 +100,11 @@ if (UNIX)
 endif ()
 
 if (LINUX)
+    if ($ENV{GCCVERSION} STRGREATER "9")
```
sfc-gh-cshi, line 103: What is this for?

tests/utils/test_setup.c
SimbaGithub: This is to fix the compile error reported by gcc9. It's smarter than previous version to calculate possible max length of the input and compare with the length of output buffer. The error here was one of the input tmp_dir could be 4096 bytes already, after adding sep and perf_test_fn it could exceed the max length of output buffer which was 4096 bytes.

tests/CMakeLists.txt
SimbaGithub: For newer version of gcc like gcc9 it throws more warnings then before. unused-result is thrown when the source code doesn't check the return value of a function call, like here https://github.com/snowflakedb/libsnowflakeclient/blob/master/tests/test_unit_put_fast_fail.cpp#L407. I would fix the source code if it's in the driver, but for test case I think simply ignore the warning would be enough.

sfc-gh-cshi (approved):

PR #360, opened 2022-02-14
Support server side encryption (SNOWFLAKE_SSE)

SimbaGithub:
Simba ticket 00372396

When stage is created with "encryption = (type = 'SNOWFLAKE_SSE')", no encryption materials will be returned from server for put/get commands. ODBC driver doesn't support that and throws error.
Follow the implementation in JDBC, no encryption/decryption when either isClientSideEncrypted is off or encryption material is empty.


sfc-gh-cshi (approved):
LGTM

PR #398, opened 2022-04-26
Disable DEV_RANDOM test on Linux

SimbaGithub:
Disable test using DEV_RANDOM on Linux as it could be blocked for a long time.


sfc-gh-cshi (approved):
LGTM. Would like to know the verification result on jenkins first.

PR #399, opened 2022-04-27
fix build issue on Mac

SimbaGithub:
Fix build issue on Mac:
1. ARCH is already used to control build type, [universal, x64, x86], so use a different name to check platform architecture.
2. On Mac "uname -p" could return i386 so handle such case as well.


sfc-gh-cshi (approved):

sfc-gh-kdama (approved):
👍

PR #400, opened 2022-04-27
remove system() call to prevent command injection

SimbaGithub:
Remove system() call and replace it with boost::filesystem::remove_all()


sfc-gh-cshi (commented):
This should be the way to solve it. However let us make sure it pass for all platforms. Esp. windows and Mac. Right now seems many tests are not running or succeed as expected.

SimbaGithub:
> This should be the way to solve it. However let us make sure it pass for all platforms. Esp. windows and Mac. Right now seems many tests are not running or succeed as expected.

Yes, windows build get error of boost/filesystem.hpp not found, which I can't reproduce locally.

sfc-gh-cshi (approved):

PR #418, opened 2022-05-12
Passdown proxy without env for file transfer

SimbaGithub:
Fix for ODBC proxy issue. So far the ODBC driver pass down proxy settings through environment variables, which would affect all connections in the same process. The customer who report this issue is using multi-threading to handling connections from different users, which would need different proxy setting for each connection.
Add proxy settings for aws/azure api also add proxy setting for file transfer so the odbc driver can use that to pass down proxy settings from connection string.
No test case added as libsnowflakeclient still use environment variables for connection, but test cases added on odbc side in this branch and passed on my local. https://github.com/snowflakedb/snowflake-odbc/tree/passdown-proxy-without-env
Will make ODBC PR from there when this PR is merged and tagged with a new version.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -150,7 +150,6 @@ set (SOURCE_FILES_PUT_GET
         cpp/util/ByteArrayStreamBuf.hpp
         cpp/util/CompressionUtil.cpp
         cpp/util/CompressionUtil.hpp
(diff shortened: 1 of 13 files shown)


sfc-gh-cshi (approved):
LGMT

PR #437, opened 2022-05-24
build arrow from source on Windows and Mac

SimbaGithub:
Build arrow from source code on Windows and Mac and bump up version number to update binaries in dependency repo. We've have that on aarch64 Linux and M1 Mac already.
Can't build arrow from source on x86 Linux for now as the build crashes on Centos6 docker. Hopefully that could be fixed when move to centos7 and eventually will build arrow from source code on all platforms.


sfc-gh-hchaturvedi (approved):
LGTM

sfc-gh-hchaturvedi:
I have a question before we merge this? Will we be building arrow with every release? If yes why? Wouldn't this increase the build time? Would this also be for the case where we need to build drivers locally in dev environment?

PR #466, opened 2022-07-18
Configurable get filesize threshold

SimbaGithub:
Simba ticket 00389874

The downloaded file could be corrupted when using aws multi-parts downloading and getting retry-able error from server side during the process. The root cause likely to be inside AWS SDK, while in the customer case they have multiple files to download and disable multi-parts downloading could get better performance.
Provide a new configuration to allow changing the threshold of multi-parts downloading  (was hardcoded to 5MB before) from application.
Will add a new connection/configuration parameter on ODBC side so we can pass the setting from there.

Eventually will upgrade AWS SDK to latest version (currently it's 1.3.50 which was released 4 years ago) and see if that can fix the issue.

### DIFF
--- a/cpp/FileMetadata.hpp
+++ b/cpp/FileMetadata.hpp
@@ -88,6 +88,10 @@ struct FileMetadata
   /// Enum for different timestamps
   enum putGetTimestamp{PUTGET_START=0,COMP_START, COMP_END, PUT_START, PUT_END, PUTGET_END, GET_START, GET_END};
 
+  // The flag to indicate if the file is larger than the threshold and needs to
+  // be treated as large file using multi-part uploading/downloading
+  bool isLarge;
+
   inline void recordPutGetTimestamp(putGetTimestamp ts=PUTGET_START)
   {
     tstamps[ts] = std::chrono::steady_clock::now();
--- a/cpp/FileMetadataInitializer.cpp
+++ b/cpp/FileMetadataInitializer.cpp
@@ -27,7 +27,8 @@ Snowflake::Client::FileMetadataInitializer::FileMetadataInitializer(
   std::vector<FileMetadata> &largeFileMetadata) :
   m_smallFileMetadata(smallFileMetadata),
   m_largeFileMetadata(largeFileMetadata),
-  m_autoCompress(true)
+  m_autoCompress(true),
+  m_downloadSizeThreshold(DOWNLOAD_DATA_SIZE_THRESHOLD)
 {
 }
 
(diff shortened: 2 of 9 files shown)


sfc-gh-hchaturvedi:
Code wise this is fine but how are we intend to address the file corruption by introducing a threshold on the limit? If I understand the gist of this fix correctly, we're introducing a threshold beyond which anything would be considered a large file and be set up for multipart download. Is the goal of this fix is to allow the customer to set this value to a higher value to prevent multi part download from happening?

SimbaGithub:
> Code wise this is fine but how are we intend to address the file corruption by introducing a threshold on the limit? If I understand the gist of this fix correctly, we're introducing a threshold beyond which anything would be considered a large file and be set up for multipart download. Is the goal of this fix is to allow the customer to set this value to a higher value to prevent multi part download from happening? @SimbaGithub

@sfc-gh-hchaturvedi Yes your understanding is correct. The root cause is likely in AWS SDK in retry-able error handling but for now around it by not using multipart download. Will update AWS SDK to see it the bug can be fixed.

sfc-gh-hchaturvedi:
> > Code wise this is fine but how are we intend to address the file corruption by introducing a threshold on the limit? If I understand the gist of this fix correctly, we're introducing a threshold beyond which anything would be considered a large file and be set up for multipart download. Is the goal of this fix is to allow the customer to set this value to a higher value to prevent multi part download from happening? @SimbaGithub
> 
> @sfc-gh-hchaturvedi Yes your understanding is correct. The root cause is likely in AWS SDK in retry-able error handling but for now around it by not using multipart download. Will update AWS SDK to see it the bug can be fixed.

@SimbaGithub In that case should we be notifying the customers to do this _only for AWS_  since not being able to do a multipart download would have perf ramifications? Also are we going to tell the customers that this is a temporary fix?

SimbaGithub:
> > > Code wise this is fine but how are we intend to address the file corruption by introducing a threshold on the limit? If I understand the gist of this fix correctly, we're introducing a threshold beyond which anything would be considered a large file and be set up for multipart download. Is the goal of this fix is to allow the customer to set this value to a higher value to prevent multi part download from happening? @SimbaGithub
> > 
> > 
> > @sfc-gh-hchaturvedi Yes your understanding is correct. The root cause is likely in AWS SDK in retry-able error handling but for now around it by not using multipart download. Will update AWS SDK to see it the bug can be fixed.
> 
> @SimbaGithub In that case should we be notifying the customers to do this _only for AWS_ since not being able to do a multipart download would have perf ramifications? Also are we going to tell the customers that this is a temporary fix?

@sfc-gh-hchaturvedi
For the customer facing issue with multipart downloading in AWS, I've mentioned in the ticket 00389874 that this fix is a workaround and the root cause is in AWS SDK, hope Snowflake support team would pass this information to the customer. For that customer case no performance impact as they have multiple files with similar size to download so the downloading is using multi-threads and the load of each threads is even.

I think we need to claim the AWS issue as an known issue (not sure if we have such documentation though), that when downloading files with multi-parts downloading and it succeeded after network retry, the downloaded file could be corrupted, and mention this new parameter as a temporary workaround there and the possible performance impact for single large file as well.

For this new parameter, I think it's permanent though. We have threshold for PUT already, it's fair to have threshold for GET as well and customer might need it for other purpose. So I prefer not to mention it's a temporary fix in the description of the new parameter, but mention it as a temporary workaround in the claim of known issues.

sfc-gh-hchaturvedi:
Sounds good. Thanks. @sfc-gh-kterada can we make sure that this is communicated back to the customer?

sfc-gh-hchaturvedi (approved):
LGTM.

PR #467, opened 2022-07-20
Keypair auth support

SimbaGithub:
Simba ticket 00389231
Add support for key pair authentication for PHP. Since PHP is using libsnowflakeclient the feature needs to be added here.

### DIFF
--- a/cpp/lib/Authenticator.cpp
+++ b/cpp/lib/Authenticator.cpp
@@ -1,0 +1,352 @@
+/*
+ * Copyright (c) 2022 Snowflake Computing, Inc. All rights reserved.
+ */
+
+#include <string>
+#include <algorithm>
+#include <cctype>
+#include <chrono>
+#include <regex>
+#include <functional>
+
+#include "Authenticator.hpp"
+#include "../logger/SFLogger.hpp"
+
+#include <openssl/pem.h>
+#include <openssl/evp.h>
+#include <openssl/err.h>
+
+#include <fstream>
+
+#include "snowflake/Simba_CRTFunctionSafe.h"
+
+#ifdef __APPLE__
+#include <CoreFoundation/CFBundle.h>
+#include <ApplicationServices/ApplicationServices.h>
+#endif
+
+#ifdef _WIN32
+#include <Shellapi.h>
+#define strcasecmp _stricmp
+#endif
+
+#define JWT_THROW(msg)                              \
+{                                                   \
+  throw Snowflake::Client::Jwt::JwtException(msg);  \
+}
+
+// wrapper functions for C
+extern "C" {
+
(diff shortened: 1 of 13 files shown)


cpp/lib/Authenticator.cpp
```diff
@@ -0,0 +1,359 @@
...
+    {
+      return 0;
+    }
+  }
+
+  SF_STATUS STDCALL auth_authenticate(SF_CONNECT * conn)
+  {
+    if (!conn || !conn->auth_object)
+    {
+      return SF_STATUS_SUCCESS;
```
sfc-gh-hchaturvedi, line 106: Should we be returning success in this case?

cpp/lib/Authenticator.cpp
```diff
@@ -0,0 +1,359 @@
...
+  {
+    switch(type) {
+      case AUTH_JWT:
+        return new AuthenticatorJWT(conn);
+      default:
+        return nullptr;
+    }
+  }
+
+  void IAuthenticator::renewDataMap(cJSON *dataMap)
```
sfc-gh-hchaturvedi, line 204: @SimbaGithub can you please remind me under what scenario we'd need this?

sfc-gh-hchaturvedi: Nvm .. I remember this is for renewing Token correct?

lib/http_perform.c
```diff
@@ -331,6 +352,15 @@ sf_bool STDCALL http_perform(CURL *curl,
 
         log_trace("Running curl call");
         res = curl_easy_perform(curl);
+
+        // forcely trigger renew timeout on the first attempt
+        if ((renew_injection) && (renew_timeout > 0) &&
+            elapsed_time && (*elapsed_time <= 0))
+        {
+            my_sleep_ms(renew_timeout * 1000);
+            res = CURLE_OPERATION_TIMEOUTED;
```
sfc-gh-hchaturvedi, line 361: nit: *TIMEDOUT

sfc-gh-hchaturvedi (approved):
LGTM

cpp/lib/Authenticator.cpp
SimbaGithub: Yes, it's for renewing jwt token.

cpp/lib/Authenticator.cpp
SimbaGithub: I don't think !conn would happen but !conn->auth_object is a normal case that extra authenticator is not needed, like user/password or oauth.

lib/http_perform.c
SimbaGithub: Will fix. Turns out both of them are defined in curl.h.

PR #468, opened 2022-08-05
SNOW-525207 add log when removing temporary folder

SimbaGithub:
Simba ticket 00388491
Customer see some temporary folders for put command abandoned. Adding logs when removing temporary folders to collect more debug information.


sfc-gh-hchaturvedi (approved):
LGTM

PR #469, opened 2022-08-05
Put 2gb file

SimbaGithub:
Simba ticket 00390901
Put command failed when file size is larger than 2GB.
Root cause: There are several places in libsnowflakeclient using long type variables for file size/offset, while long is 4 bytes on Windows and could cause data overflow when file size is larger than 2GB.


sfc-gh-hchaturvedi (approved):
LGTM

PR #474, opened 2022-09-01
Update openssl 1.1.1q

SimbaGithub:
Update openssl to 1.1.1q with the fix here: https://github.com/openssl/openssl/issues/18720 to solve the build issue on Mac.


sfc-gh-igarish (approved):
Looks good.

PR #479, opened 2022-10-07
support proxy settings in connection attributes

SimbaGithub:
Simba ticket 00402653
Support connection individual proxy settings for PHP connector.
Since PHP is using libsnowflakeclient we need to implement that in libsnowflakeclient first.

### DIFF
--- a/cpp/FileTransferAgent.cpp
+++ b/cpp/FileTransferAgent.cpp
@@ -67,6 +67,7 @@ Snowflake::Client::FileTransferAgent::FileTransferAgent(
   m_executionResults(nullptr),
   m_storageClient(nullptr),
   m_lastRefreshTokenSec(0),
+  m_transferConfig(transferConfig),
   m_uploadStream(nullptr),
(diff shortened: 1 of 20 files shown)


sfc-gh-igarish (approved):
LGTM. Looks like assumption is proxy and no_proxy won't intersect, if it's then give preference to proxy for it. Please comment it somewhere, if that's true.

sfc-gh-ext-simba-hx:
> LGTM. Looks like assumption is proxy and no_proxy won't intersect, if it's then give preference to proxy for it. Please comment it somewhere, if that's true.
@sfc-gh-igarish 
No. Both proxy and no_proxy are passed to curl using curl_easy_setopt() and they can be intersected. The test cases are to confirm both proxy and no_proxy can be passed to curl correctly. The intersection of proxy/no_proxy is curl feature and I think curl can handle it well. It's a bit tricky to add such test cases since we don't have a proxy available in the test environment on github.
- Harry

PR #480, opened 2022-10-11
key pair auth bug fix

SimbaGithub:
simba ticket 00404349
Key pair authentication doesn't work.
The bug itself is a simple typo and the problem is the test case didn't catch it. The reason is that setup_snowflake_connection() is called by all test cases to setup connection attributes from environments while password is set there. The jwt test cases were actually using user/password authentication to make the connection.
Unset password in test cases to ensure key pair auth is actually being used. Same test case issue on PHP side as well will make same fix there.

### DIFF
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -181,7 +181,7 @@ cJSON *STDCALL create_auth_json_body(SF_CONNECT *sf,
     snowflake_cJSON_AddItemToObject(body, "data", data);
 
     // update authentication information
-    auth_update_json(sf, data);
+    auth_update_json(sf, body);
 
     return body;
 }
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -180,8 +180,8 @@ cJSON *STDCALL create_auth_json_body(SF_CONNECT *sf,
     body = snowflake_cJSON_CreateObject();
     snowflake_cJSON_AddItemToObject(body, "data", data);
 
-    // update authentication information
-    auth_update_json(sf, body);
+    // update authentication information to body
+    auth_update_json_body(sf, body);
 
     return body;
 }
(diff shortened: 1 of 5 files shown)


tests/test_jwt.cpp
```diff
@@ -202,6 +202,8 @@ void test_unencrypted_pem(void **unused) {
 
   snowflake_set_attribute(sf, SF_CON_AUTHENTICATOR, SF_AUTHENTICATOR_JWT);
   snowflake_set_attribute(sf, SF_CON_PRIV_KEY_FILE, keyFilePath.c_str());
+  // explicitly unset password to esure we are using key pair auth
+  snowflake_set_attribute(sf, SF_CON_PASSWORD, "");
```
sfc-gh-igarish, line 206: why we need to do it?

lib/connection.c
```diff
@@ -181,7 +181,7 @@ cJSON *STDCALL create_auth_json_body(SF_CONNECT *sf,
     snowflake_cJSON_AddItemToObject(body, "data", data);
 
     // update authentication information
-    auth_update_json(sf, data);
+    auth_update_json(sf, body);
```
sfc-gh-igarish, line 184: Can you put some comment why we are using "body" now instead of "data"?

sfc-gh-igarish (approved):
LGTM. Just comment the code, if you can.

tests/test_jwt.cpp
SimbaGithub: The key pair auth was broken but the test case passed because it was actually using user/password

lib/connection.c
SimbaGithub: Done.

PR #482, opened 2022-10-24
bug fix: curl to use proxy for ocsp

sfc-gh-ext-simba-hx:
OCSP check is not using proxy being set through CURLOPT_PROXY and CURLOPT_NOPROXY so if connection individual proxy settings are set from there the OCSP check could fail.


sfc-gh-ext-simba-hx:
I have read the CLA Document and I hereby sign the CLA

sfc-gh-igarish (approved):
LGTM

PR #486, opened 2022-12-28
[SNOW-480523] use gcs downscoped token for GET/PUT

sfc-gh-ext-simba-hx:
for feature request: https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/91

The changes on libsnowflakeclient side to use down scoped token. Since libsnowflakeclient doesn't support GET/PUT for GCS by it's own no test cases added in this PR.
More changes and test cases are added on ODBC side here: https://github.com/snowflakedb/snowflake-odbc/tree/SNOW-480523-gcp-downscoped-token. Will make ODBC PR from there when this PR is merged and can be picked up from ODBC side.

### DIFF
--- a/cpp/SnowflakeGCSClient.cpp
+++ b/cpp/SnowflakeGCSClient.cpp
@@ -36,6 +36,11 @@ SnowflakeGCSClient::SnowflakeGCSClient(StageInfo *stageInfo, unsigned int parall
   m_statement(statement),
   m_gcsAccessToken(stageInfo ? stageInfo->credentials[GCS_TOKEN_KEY] : "")
 {
+  if (!m_gcsAccessToken.empty())
+  {
+    CXX_LOG_INFO("Using GCS down scoped token.");
+  }
+
   //Ensure the stage location ended with /
   if ((!m_stageInfo->location.empty()) && (m_stageInfo->location.back() != '/'))
   {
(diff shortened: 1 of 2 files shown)


sfc-gh-igarish (changes requested):
Can we add logging info, so when something is not working as expected, we can debug using log files? Are we going to add connection option like JDBC here or in ODBC call? Normally ODBC should pass options to libsfclient. But not sure. Please check.

sfc-gh-ext-simba-hx:
> Can we add logging info, so when something is not working as expected, we can debug using log files? Are we going to add connection option like JDBC here or in ODBC call? Normally ODBC should pass options to libsfclient. But not sure. Please check.

Hi @sfc-gh-igarish ,
Will add loging info.
About the connection option, it would be on ODBC side if we need to add that. I'd like to confirm the specification before adding it though, how this option needs to be sent to server to change the behavior.
I can see "GCS_USE_DOWNSCOPED_CREDENTIAL" is added to connection option in JDBC test cases but can't find logic in driver code handling that. Please point me to the logic if I miss it. For my understanding, in JDBC the connection option "GCS_USE_DOWNSCOPED_CREDENTIAL" will be sent to server as a part of "CLIENT_ENVIRONMENT" which basically a note of all connection options on client side, and I doubt whether the server would actually read settings from there and change the behavior. So far I don't see this option changing the server behavior and we have to ask Snowflake team to enable down scoped credential on our account to get GCS token returned.
Setting session parameter like "GCS_USE_DOWNSCOPED_CREDENTIAL" or "FORCE_GCP_USE_DOWNSCOPED_CREDENTIAL" as mentioned in https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/6 doesn't work either.

sfc-gh-igarish:
Yes please add test case, once option is available for ODBC driver, it should start working.

sfc-gh-igarish (approved):
LGTM

PR #498, opened 2023-03-20
add bug and feature templates

sfc-gh-igarish:
Adding Bug and Feature teamples as we have it in other drivers such as JDBC repo.


sfc-gh-mkeller (approved):
:ship: :ship:

PR #503, opened 2023-03-31
SNOW-761229 Bug fix: encoding ocsp request data in URL to avoid invalid URL issue

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/259
Encode Base64 OCSP Request data in URL for some corner cases that it might have characters that are not URL safe.
Unit test only as it's for for some corner cases that not easy to reproduce.

### DIFF
--- a/deps/curl-7.88.1/lib/vtls/sf_ocsp.c
+++ b/deps/curl-7.88.1/lib/vtls/sf_ocsp.c
@@ -190,7 +190,7 @@ static void printOCSPFailOpenWarning(SF_OTD *ocsp_log, struct Curl_easy *data);
 static char * generateOCSPTelemetryData(SF_OTD *ocsp_log);
 static void clearOSCPLogData(SF_OTD *ocsp_log);
 static SF_TESTMODE_STATUS getTestStatus(SF_OCSP_TEST test_name);
-// Intentially make it global for test purpose.
+/* Intentially make it global for test purpose */
 size_t encodeUrlData(const char *url_data, size_t data_size, char** encoded_ptr);
 
 static int _mutex_init(SF_MUTEX_HANDLE *lock);
@@ -283,9 +283,10 @@ size_t encodeUrlData(const char *url_data, size_t data_size, char** encoded_ptr)
   char* encode_buf = (char*)malloc(buf_len);
   char* cur_ptr = encode_buf;
   size_t enc_len = 0;
+  size_t pos = 0;
 
   // encode all special characters
-  for (size_t pos = 0; pos < data_size; pos++)
+  for (pos = 0; pos < data_size; pos++)
   {
     // if unreserved, put as is
     // RFC 3986 section 2.3 Unreserved Characters
--- a/deps/curl-7.88.1/lib/vtls/sf_ocsp.c
+++ b/deps/curl-7.88.1/lib/vtls/sf_ocsp.c
@@ -190,8 +190,9 @@ static void printOCSPFailOpenWarning(SF_OTD *ocsp_log, struct Curl_easy *data);
 static char * generateOCSPTelemetryData(SF_OTD *ocsp_log);
 static void clearOSCPLogData(SF_OTD *ocsp_log);
 static SF_TESTMODE_STATUS getTestStatus(SF_OCSP_TEST test_name);
+
 /* Intentially make it global for test purpose */
-size_t encodeUrlData(const char *url_data, size_t data_size, char** encoded_ptr);
+CURLcode encodeUrlData(const char *url_data, size_t data_size, char** outptr, size_t *outlen);
 
 static int _mutex_init(SF_MUTEX_HANDLE *lock);
 static int _mutex_lock(SF_MUTEX_HANDLE *lock);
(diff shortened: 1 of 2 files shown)


deps/curl-7.88.1/lib/vtls/sf_ocsp.c
```diff
@@ -638,9 +687,10 @@ static OCSP_RESPONSE * queryResponderUsingCurl(char *url, OCSP_CERTID *certid, c
   {
     if (!ACTIVATE_SSD)
     {
+      encodeUrlData(ocsp_req_base64, ocsp_req_base64_len, &encoded_ocsp_req_base64);
       snprintf(urlbuf, sizeof(urlbuf),
                ocsp_cache_server_retry_url_pattern,
-               host, ocsp_req_base64);
+               host, encoded_ocsp_req_base64);
```
sfc-gh-ext-simba-hx, line 693: Checked entire sf_ocsp.c this seems to be the only place put base64 encoded ocsp request data into URL.

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
```diff
@@ -266,6 +268,53 @@ static const unsigned int MAX_SLEEP_TIME = 16000000;
...
 
 
+/* Encode request data as a part of URL
+ * Encode all characters reserved by URL
+ * @param url_data The source data needs to be encoded_ptr
+ * @param data_size The size of the source data
+ * @param encoded_ptr The allocated buffer filled with encoded data, need to be freed by caller
+ * return the size of encoded data
+ */
+size_t encodeUrlData(const char *url_data, size_t data_size, char** encoded_ptr)
```
sfc-gh-igarish, line 278: what will happen URL is already encoded?

sfc-gh-ext-simba-hx: That won't happen as here we only encode the base64 request data that will be appended as a part of the URL, not entire URL.
encodeUrlData(ocsp_req_base64, ocsp_req_base64_len, &encoded_ocsp_req_base64);

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
```diff
@@ -266,6 +268,53 @@ static const unsigned int MAX_SLEEP_TIME = 16000000;
...
+{
+  // allocate buffer 3 times larger than source data in case every single, add 1 for '\0'
+  // character needs to be encoded as %xx
+  size_t buf_len = data_size * 3 + 1;
+  char* encode_buf = (char*)malloc(buf_len);
```
sfc-gh-igarish, line 283: Can we check for null?

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
```diff
@@ -758,6 +809,7 @@ static OCSP_RESPONSE * queryResponderUsingCurl(char *url, OCSP_CERTID *certid, c
 
 end:
   if (ocsp_req_base64) curl_free(ocsp_req_base64);
+  if (encoded_ocsp_req_base64) curl_free(encoded_ocsp_req_base64);
```
sfc-gh-igarish, line 812: You allocated using malloc and then free using curl_free. is that ok?

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
sfc-gh-igarish, line 283: Make sure malloc is right API here and during free it calls curl_free.

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
```diff
@@ -266,6 +268,53 @@ static const unsigned int MAX_SLEEP_TIME = 16000000;
...
+      snprintf(cur_ptr, buf_len, "%%%.2X", car);
+      cur_ptr += 3;
+      enc_len += 3;
+      buf_len -= 3;
+    }
+  }
+
+  *cur_ptr = '\0';
+  *encoded_ptr = encode_buf;
+  return enc_len;
```
sfc-gh-igarish, line 315: Are we using in the caller function?

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: check added

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: This is following the existing curl implementation here:
https://github.com/snowflakedb/libsnowflakeclient/blob/master/deps/curl-7.88.1/lib/base64.c#L206
The buffer ocsp_req_base64 is allocated with malloc and freed using curl_free as well.

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: Confirmed, following existing curl implementation.

deps/curl-7.88.1/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: Not really but won't hurt to have data length returned. Make it optional output argument.

sfc-gh-igarish (approved):
LGTM

PR #504, opened 2023-04-03
Bug fix: only set proxy username and password options when specified

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/294
ODBC driver sends empty Proxy-Authorization header to proxy server when no proxy username and password options in proxy settings which caused connection failure when using proxy.
The root cause is that when using proxy proxy username and password always been set without checking if they are empty.
Do the same fix [here](https://github.com/aws/aws-sdk-cpp/commit/59fe5b1b4d3cf9d26586a543376818e581f5aafc) for all the places need proxy settings.
No test case added as we don't have a proxy environment that doesn't need user/password authentication. But verify the fix with tcpdump as mentioned in the ticket, confirmed that
- without the fix Proxy-Authorization header was added with empty credentials
- with the fix Proxy-Authorization header was removed

### DIFF
--- a/cpp/util/SnowflakeCommon.cpp
+++ b/cpp/util/SnowflakeCommon.cpp
@@ -108,10 +108,13 @@ CURLcode set_curl_proxy(CURL *curl, const char* proxy, const char* no_proxy)
     if (res != CURLE_OK) return res;
     res = curl_easy_setopt(curl, CURLOPT_PROXYPORT, (long)proxySettings.getPort());
     if (res != CURLE_OK) return res;
-    res = curl_easy_setopt(curl, CURLOPT_PROXYUSERNAME, proxySettings.getUser().c_str());
-    if (res != CURLE_OK) return res;
-    res = curl_easy_setopt(curl, CURLOPT_PROXYPASSWORD, proxySettings.getPwd().c_str());
-    if (res != CURLE_OK) return res;
+    if (!proxySettings.getUser().empty() || !proxySettings.getPwd().empty())
+    {
+      res = curl_easy_setopt(curl, CURLOPT_PROXYUSERNAME, proxySettings.getUser().c_str());
+      if (res != CURLE_OK) return res;
+      res = curl_easy_setopt(curl, CURLOPT_PROXYPASSWORD, proxySettings.getPwd().c_str());
+      if (res != CURLE_OK) return res;
+    }
     return curl_easy_setopt(curl, CURLOPT_NOPROXY, proxySettings.getNoProxy().c_str());
   }
 }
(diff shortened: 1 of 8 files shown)


cpp/util/SnowflakeCommon.cpp
```diff
@@ -108,10 +108,13 @@ CURLcode set_curl_proxy(CURL *curl, const char* proxy, const char* no_proxy)
     if (res != CURLE_OK) return res;
     res = curl_easy_setopt(curl, CURLOPT_PROXYPORT, (long)proxySettings.getPort());
     if (res != CURLE_OK) return res;
-    res = curl_easy_setopt(curl, CURLOPT_PROXYUSERNAME, proxySettings.getUser().c_str());
-    if (res != CURLE_OK) return res;
-    res = curl_easy_setopt(curl, CURLOPT_PROXYPASSWORD, proxySettings.getPwd().c_str());
-    if (res != CURLE_OK) return res;
+    if (!proxySettings.getUser().empty() || !proxySettings.getPwd().empty())
```
sfc-gh-igarish, line 111: Do we need both user name and password? if yes, then it should be AND condition.

scripts/build_awssdk.bat
```diff
@@ -3,7 +3,7 @@
 :: GitHub repo: https://github.com/aws/aws-sdk-cpp.git
 ::
 @echo off
-set aws_version=1.3.50.1
+set aws_version=1.3.50.2
```
sfc-gh-igarish, line 6: why we need this change for this issue?

cpp/util/SnowflakeCommon.cpp
sfc-gh-ext-simba-hx: It's possible that have valid user name and empty password. The current logic is that when either username or password is set, Proxy-Authorization header will be added to with the settings, the one not being set will be empty.

scripts/build_awssdk.bat
sfc-gh-ext-simba-hx: This is for version management for third-parties. The artifacts of third-parties will be uploaded to S3 storage only when the version is new and not found on S3. ODBC driver or other Snowflake application needs those third-parties will pick up them from there.

cpp/util/SnowflakeCommon.cpp
sfc-gh-ext-simba-hx: The fix in aws-sdk-cpp [here](https://github.com/aws/aws-sdk-cpp/commit/59fe5b1b4d3cf9d26586a543376818e581f5aafc) is using OR condition as well so I think using the same logic would be safe.

sfc-gh-igarish: ok. But what will happen user name is empty but the password is provided?

sfc-gh-igarish (approved):
LGTM

PR #505, opened 2023-04-04
SNOW-760226 Bug fix: remove deprecated openssl function calls

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/253
Following the guidance here: https://www.openssl.org/docs/man3.0/man7/OPENSSL_NO_DEPRECATED.html
build libsnowflakeclient with following definition and fix all build errors.(OPENSSL_API_COMPAT set to the version number of 1.1.1q to disable all deprecated functions in previous versions)
OPENSSL_API_COMPAT=0x1010111fL
OPENSSL_NO_DEPRECATED

### DIFF
--- a/cpp/crypto/Cryptor.cpp
+++ b/cpp/crypto/Cryptor.cpp
@@ -125,18 +125,6 @@ static void getRandomBytes(char *const out,
 }*/
 
 /**
- * Callback for getting the calling thread's id.
- */
-static void getThreadId(CRYPTO_THREADID *const threadId) noexcept
-{
-  //static_assert(sizeof(pthread_self()) <= sizeof(threadId->val),
-  //              "Unable to store thread id in CRYPTO_THREADID");
-
-  CRYPTO_THREADID_set_pointer(theardId, nullptr);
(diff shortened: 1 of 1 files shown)


cpp/crypto/Cryptor.cpp
```diff
@@ -124,18 +124,6 @@ static void getRandomBytes(char *const out,
   (void) failure;
 }*/
 
-/**
- * Callback for getting the calling thread's id.
- */
-static void getThreadId(CRYPTO_THREADID *const threadId) noexcept
```
sfc-gh-igarish, line 130: Do we have  any other API do to the same stuff?

sfc-gh-igarish (approved):
LGTM

cpp/crypto/Cryptor.cpp
sfc-gh-ext-simba-hx: This function was used to set threading callback `CRYPTO_THREADID_set_callback(getThreadId);` which was deprecated and commented out already before this change.

PR #510, opened 2023-05-16
Bug fix: crash with OCSP validation

sfc-gh-ext-simba-hx:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/371
The bug was just a typo but not caught by test cases since OCSP test is not run on Windows.
Add OCSP test case to run on all platforms.

Also disabled oob test due to this issue:
https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/376


sfc-gh-wshangguan (approved):
LGTM

PR #513, opened 2023-06-01
Update openssl 3.0.9

sfc-gh-ext-simba-hx:
update openssl to 3.0.9 with fips enabled
generate legacy provider in case customer might need it to use legacy algorithm
bump up minor version for all dependencies using openssl to get them rebuild with openssl3 and update to s3
update build scripts and aws sdk source code to fix build issue.


sfc-gh-igarish (approved):
Too many files to review. But all related to thrid party files. So if test passes then merge it.

PR #515, opened 2023-06-14
Build on centos7

sfc-gh-ext-simba-hx:


sfc-gh-igarish:
@sfc-gh-kdama do we need changes in jenkin jobs? Not for libsfclient but for ODBC.

sfc-gh-igarish (approved):
LGTM

PR #517, opened 2023-06-19
Update util linux 2.39

sfc-gh-ext-simba-hx:
sdk issue 350

Update util-linux to the latest version of 2.39


sfc-gh-igarish (approved):
Too many files. But it's related to third party source, so must be good. Please make sure we test it.

PR #519, opened 2023-06-23
Revert "Update util linux 2.39"

sfc-gh-ext-simba-hx:
Reverts snowflakedb/libsnowflakeclient#517

There is build issue on Linux aarch64 and we need to update the docker image to solve the issue.
Revert to use 2.36.1 for June release and will update to 2.39 in July when docker image is updated.


sfc-gh-igarish (approved):
LGTM

PR #523, opened 2023-07-13
remove file causing snyk issue

sfc-gh-ext-simba-hx:
sdk issue 529
False alarm. snyk scan repoorts [CVE-2014-0114](https://github.com/advisories/GHSA-p66x-2cv9-qq3v) on deps/aws-sdk-cpp-1.3.50/code-generation/generator/pom.xml, while that's for C++ code generation for aws sdk, Apache Struts (if it's used for code generation) won't be used at runtime.

remove the pom.xml file to suppress the false alarm from snyk scan.


sfc-gh-igarish (approved):
LGTM. But do we know why we added it at first place? After removing will it cause any other issue?

sfc-gh-igarish:
@sfc-gh-ext-simba-hx I understand that, but having pom.xml I think snyk was able to detect dependency chain and  detect the vulnerability and suggest the solution. Now we remove that path. right?

sfc-gh-ext-simba-hx:
pom.xml is for java package while the driver is using c++. The java packages in pom.xml are used for code generation only and none of them will be used at runtime. So I think any issue detected there would be false alarm.

PR #527, opened 2023-07-20
SNOW-867941 - Fix LibSnowflakeClient-Win64-VS14-Debug Jenkins job

sfc-gh-anavalos:
The job has been failing when trying to compile Boost with Visual Studio 2022.
The fix for this was to add vc14 when calling the build.bat script.
Another issue was Boost was compiled for x86 and not for x64. The problem was with the environment variable platform. The value should be "x86" or "x64". However, when the script was trying to validate the platform the value was capitalized and the conditions failed. The fix is to update the condition with the correct value.
The fix worked in this test Jenkins job https://ci-dev-122.int.snowflakecomputing.com/job/LibSnowflakeClient-VSTest-Win64-VS14-Debug/11/

### DIFF
--- a/scripts/build_boost_source.bat
+++ b/scripts/build_boost_source.bat
@@ -61,7 +61,7 @@ if /I "%dynamic_runtime%"=="on" (
 
 call "%BOOST_SOURCE_DIR%\bootstrap.bat" --with-libraries=filesystem,regex,system
 if %ERRORLEVEL% NEQ 0 goto :error
-b2 stage --stagedir=%BOOST_INSTALL_DIR% --includedir=%BOOST_INSTALL_DIR%\include --layout=system --with-system --with-filesystem --with-regex link=static runtime-link=%runtimelink% threading=multi address-model=%bitness% variant=%variant% runtime-debugging=%debugging% cxxflags=/ZH:SHA_256 install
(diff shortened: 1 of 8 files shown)


sfc-gh-igarish (approved):
LGTM

sfc-gh-kdama (approved):

scripts/build_boost_source.bat
```diff
@@ -26,11 +26,11 @@ if %ERRORLEVEL% NEQ 0 goto :error
 
 set curdir=%cd%
 
-if "%platform%"=="x64" (
+if "%platform%"=="X64" (
```
sfc-gh-ext-simba-hx, line 29: We can't just change that as same checking in other scripts as well like this one for azure sdk: https://github.com/snowflakedb/libsnowflakeclient/blob/v1.0.0/scripts/build_azuresdk.bat#L32
Do we know why it become upper case and fix that directly? Otherwise we need to fix this in all other scripts as well, maybe make the checking case insensitive.

sfc-gh-anavalos: I couldn't find where this value was updated to upper case. I am going to review one more time. Otherwise I will fix in the other scripts.

sfc-gh-anavalos: Updated the other scripts.

sfc-gh-ext-simba-hx (approved):
LGTM

PR #546, opened 2023-08-10
SNOW-819682 enable oob test

sfc-gh-ext-simba-hx:
sdk issue 376. oob test was commented out due to certificate issue on telemetry endpoint. Enable it to check as the issue has been solved.


sfc-gh-igarish (approved):
LGTM

PR #547, opened 2023-08-17
SNOW-880365 Curl option to disable oob

sfc-gh-ext-simba-hx:
fix for SDK issue 587
add curl option to disable oob telemetry
Change oob timeout from 100 seconds to 5 seconds and change test case accordingly as well.


sfc-gh-igarish (approved):
LGTM. How user can enable it? Please document it.

Here only add a curl option can be set from driver. The parameter will be added on ODBC driver side which would be same as JDBC: `CLIENT_OUT_OF_BAND_TELEMETRY_ENABLED`

sfc-gh-ext-simba-hx:
@sfc-gh-igarish 
Here only add a curl option can be set from driver. The parameter will be added on ODBC driver side which would be same as JDBC: `CLIENT_OUT_OF_BAND_TELEMETRY_ENABLED`

PR #553, opened 2023-08-24
SNOW-872773: ODBC Driver Does not retry Entire OCSP URL

sfc-gh-ext-simba-jl:
Issue 560: https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/560

Add the path and port, if exist, to the OCSP retry URL instead of taking just the host.


deps/curl-8.1.2/lib/vtls/sf_ocsp.c
```diff
@@ -711,9 +711,31 @@ static OCSP_RESPONSE * queryResponderUsingCurl(char *url, OCSP_CERTID *certid, c
...
+      char full_url[4096] = "";
+      strcpy(full_url, host ? host : "");
+
+      // default port numbers will not be added to the url
+      // (443 if the scheme is https, else 80)
+      if (port &&
+          atoi(port) > 0 &&
+          ((use_ssl && atoi(port) != 443) || (!use_ssl && atoi(port) != 80)))
+    {
+        strcat(full_url, ":");
```
sfc-gh-igarish, line 725: If host is empty as per line#717, is it ok to put ":" for port?

deps/curl-8.1.2/lib/vtls/sf_ocsp.c
```diff
@@ -711,9 +711,31 @@ static OCSP_RESPONSE * queryResponderUsingCurl(char *url, OCSP_CERTID *certid, c
         sf_otd_set_event_sub_type(OCSP_RESPONSE_ENCODE_FAILURE, ocsp_log_data);
         goto end;
       }
+
+      // send the entire OCSP URL to the cache server
+      char full_url[4096] = "";
```
sfc-gh-igarish, line 716: Can we make a define const for 4096?

deps/curl-8.1.2/lib/vtls/sf_ocsp.c
```diff
@@ -711,9 +711,31 @@ static OCSP_RESPONSE * queryResponderUsingCurl(char *url, OCSP_CERTID *certid, c
...
+          ((use_ssl && atoi(port) != 443) || (!use_ssl && atoi(port) != 80)))
+    {
+        strcat(full_url, ":");
+        strcat(full_url, port);
+    }
+
+      // path is guaranteed to begin with a / character
+      if (path && (strlen(path) > 1))
+      {
+          strcat(full_url, path);
```
sfc-gh-pfus, line 732: Shouldn't we URL encode path? Does our `retry` proxy handle correctly slashes?

deps/curl-8.1.2/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-jl: Added #define MAX_BUFFER_LENGTH 4096

deps/curl-8.1.2/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-jl: Yes, it's ok because the request will fail if host is empty.

deps/curl-8.1.2/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-jl: Support team has confirmed the fix already so URL encoded is not required.

sfc-gh-igarish (approved):
LGTM

PR #554, opened 2023-08-24
fix build issue with xcode14 on Mac

sfc-gh-ext-simba-hx:
add build option of -Wno-error=deprecated-declarations to suppress the build error with boost header.


sfc-gh-igarish (approved):
LGTM

PR #557, opened 2023-09-11
SNOW-871471: [HTAP] add retry context in retry for query request

sfc-gh-ext-simba-hx:
sdk issue 550

Currently the new connection parameter SF_CON_INCLUDE_RETRY_CONTEXT controls both retryCount and retryReason as for libsnowflakeclient both of them are newly added while for other drivers retryCount already exist so the new connection parameter (like in ODBC) includeRetryReason only control retryReason.

@sfc-gh-igarish Please confirm whether to keep the current implementation or change to be inline with other drivers (always have retryCount and have a connection parameter to control retryReason)

### DIFF
--- a/include/snowflake/client.h
+++ b/include/snowflake/client.h
@@ -233,7 +233,7 @@ typedef enum SF_ATTRIBUTE {
     SF_CON_PROXY,
     SF_CON_NO_PROXY,
     SF_CON_DISABLE_QUERY_CONTEXT_CACHE,
-    SF_CON_INCLUDE_RETRY_CONTEXT,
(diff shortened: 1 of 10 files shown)


sfc-gh-igarish:
@sfc-gh-ext-simba-hx Better to have in line with other drivers.

sfc-gh-ext-simba-hx:
> @sfc-gh-ext-simba-hx Better to have in line with other drivers.

@sfc-gh-igarish Fixed. Please review.

sfc-gh-igarish (approved):
LGTM

PR #558, opened 2023-09-11
fix flaky test case

sfc-gh-ext-simba-hx:
After changing oob timeout from 100 seconds to 5 seconds in https://github.com/snowflakedb/libsnowflakeclient/pull/547
sometime oob test case fails on timeout. Treat that as expected result as we don't want to increase the timeout which could block customer application longer on oob sending.


sfc-gh-igarish (approved):
LGTM

PR #559, opened 2023-09-11
SNOW-871482: [HTAP] add test case for snowflake_sfqid

sfc-gh-ext-simba-hx:
sdk issue 551
last query id for libsnowflakeclient

We have existing function snowflake_sfqid() returns last query id on statement level already while no test case cover it.
Add test case to ensure it works for both success and fail cases.


sfc-gh-igarish (approved):
LGTM

PR #566, opened 2023-10-06
SNOW-928053: Upgrade to openssl3.0.11 and libcurl 8.4.0

sfc-gh-ext-simba-jz:
Security concerns, I think. I was assigned the task in our OEM release 3.1.2 (during the release curl made a new release 8.4.0 from 8.3.0). The next monthly release should be 3.1.3, which make sense to include all those upgrades.

upgrade dependencies openssl-3.0.11 and curl-8.4.0
patch snowflake change of ocsp on curl-8.1.2 to curl-8.4
update source and build versions of related build scripts


sfc-gh-igarish:
There are 5000+ files....

sfc-gh-igarish:
@sfc-gh-ext-simba-jz could you please update description why we are upgrading?

sfc-gh-ext-simba-jz:
Security concerns, I think. I was assigned the task in our OEM release 3.1.2 (during the release curl made a new release 8.4.0 from 8.3.0). The next monthly release should be 3.1.3, which make sense to include all those upgrades.

sfc-gh-igarish (approved):
As it has 5000+ files and all are third party source code. So not reviewing files. If all merge gate pass then merge it.

sfc-gh-ext-simba-jz:
Yes, there are over 5000 files, we replace openssl-3.0.9 and curl-8.1.2 with opessl-3.0.11 and curl-8.4.0. You don't need to pay attention to openssl because we don't change anything in openssl, just grab and use. We do modify curl for OCSP support, and I have patched everything what we have done in curl-8.1.2 to curl-8.4.0. So the actual change doesn't sound that scary.

PR #567, opened 2023-10-06
save diskspace on git

sfc-gh-ext-simba-hx:
Seems github reduced the diskspace for it's runners the debug build on Linux start to fail.
Remove some unnecessary files to save disk space.
The zip files was a pre-build version of arrow. Since currently we build arrow from source code on all platforms it's no use anymore. Anyone needs pre-build artifacts of arrow should get that from s3, like what ODBC is doing currently.
files in deps is not used after all dependencies building finished. The artifacts are in deps-build so deps can be deleted to save some space.

### DIFF
--- a/ci/build/build.sh
+++ b/ci/build/build.sh
@@ -91,7 +91,7 @@ download_build_component arrow "$SCRIPTS_DIR/build_arrow.sh" "$target"
 
 # very tight diskspace limit on github runners, clear deps folder with all .o files
 if [[ -n "$GITHUB_ACTIONS" ]]; then
-    rm -rf $SCRIPTS_DIR/../deps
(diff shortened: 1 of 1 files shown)


sfc-gh-igarish:
If you remove it, make sure nobody references it. If it's find alternate way to access it.

sfc-gh-ext-simba-hx:
> If you remove it, make sure nobody references it. If it's find alternate way to access it.

Sure, it should be fine if the test passes.
The zip files was a pre-build version of arrow. Since currently we build arrow from source code on all platforms it's no use anymore.
files in deps is not used after all dependencies building finished. The artifacts are in deps-build so deps can be deleted to save some space.

sfc-gh-igarish (approved):
LGTM

PR #568, opened 2023-10-12
save more disk space on github

sfc-gh-ext-simba-hx:
Build fails again due to disk space issue.
Use workaround provided here https://github.com/actions/runner-images/issues/2840#issuecomment-790492173 to save more disk space.

### DIFF
--- a/ci/build_linux.sh
+++ b/ci/build_linux.sh
@@ -17,6 +17,12 @@ else
     export GIT_URL=https://github.com/${GITHUB_REPOSITORY}.git
     export GIT_BRANCH=origin/$(basename ${GITHUB_REF})
     export GIT_COMMIT=${GITHUB_SHA}
+    # remove unnecssary files to save disk space on github
+    # base on the workaround provided here:
+    # https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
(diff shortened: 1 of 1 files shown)


ci/build_linux.sh
```diff
@@ -17,6 +17,12 @@ else
     export GIT_URL=https://github.com/${GITHUB_REPOSITORY}.git
     export GIT_BRANCH=origin/$(basename ${GITHUB_REF})
     export GIT_COMMIT=${GITHUB_SHA}
+    # remove unnecssary files to save disk space on github
+    # base on the workaround provided here:
+    # https://github.com/actions/runner-images/issues/2840#issuecomment-790492173
+    sudo rm -rf /usr/share/dotnet
```
sfc-gh-igarish, line 23: are we sure to remove these three directory?

sfc-gh-ext-simba-hx: This is a workaround provided by a person from github for disk space issue.
Remove those directories only when build on github hosted runners and I think it would be restored from their image each time when starting a new job.
Those directories is from github hosted runners and not used by our build, so should be safe to remove them.

sfc-gh-igarish (approved):
LGTM

PR #571, opened 2023-10-20
SNOW-878098: Retry Strategy

sfc-gh-ext-simba-hx:
sdk issu 581

- add retry on http error code 429
- new retry strategy of jittered exponential backoff, max retry count of 7 and retry time of 300 seconds apply on login/authenticator/token requests
- add CLIENT_APP_ID, CLIENT_APP_VERSION in the header of login/authenticator/token requests

### DIFF
--- a/lib/connection.c
+++ b/lib/connection.c
@@ -160,11 +160,11 @@ cJSON *STDCALL create_auth_json_body(SF_CONNECT *sf,
 
     //Create Request Data JSON blob
     data = snowflake_cJSON_CreateObject();
-    snowflake_cJSON_AddStringToObject(data, "CLIENT_APP_ID", int_app_name);
+    snowflake_cJSON_AddStringToObject(data, CLIENT_APP_ID_KEY, int_app_name);
 #ifdef MOCK_ENABLED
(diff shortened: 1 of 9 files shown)


lib/connection.c
```diff
@@ -542,7 +556,9 @@ decorrelate_jitter_next_sleep(DECORRELATE_JITTER_BACKOFF *djb, uint32 sleep) {
     {
       sleep = 4;
     }
-    return ((uint32)(sleep/2) + (uint32) (rand() % (sleep/2)));
+    // (sleep/2) + (random from 0 to sleep) = random from sleep/2 to sleep + sleep/2
+    // = sleep +-50%
+    return ((uint32)(sleep/2) + (uint32) (rand() % (sleep + 1)));
```
sfc-gh-igarish, line 561: nit: jitter should be added separately to make it clear or change method name to reflect this is total next sleep value.

sfc-gh-igarish (approved):
LGTM

lib/connection.c
sfc-gh-ext-simba-hx: changed function name to get_next_sleep_with_jitter

PR #572, opened 2023-10-23
SNOW-932679: reduce delay on ocsp timeout

sfc-gh-ext-simba-hx:
Fix for both sdk issue 692 and 706
Before the fix if the OCSP validate requests failed with timeout it could cause 15 (5 seconds timeout on cache downloading and 5 seconds timeout * 2 retries on fetch request) seconds delay per entry in the certificate chain, with typically 4 entries in the chain the total delay could be up to 60 seconds and could fail the SSL connection even the driver try to establish the connection with FAIL_OPEN mode.
The fix is to remember the host name reached timeout in previous entry and skip further attempt on the same host in following entries to reduce the delay.


deps/curl-8.1.2/lib/vtls/sf_ocsp.c
```diff
@@ -753,6 +763,21 @@ static OCSP_RESPONSE * queryResponderUsingCurl(char *url, OCSP_CERTID *certid, c
...
+      urlbuf, &host, &port, &path, &use_ssl);
+  if (!url_parse_result)
+  {
+    failf(data, "Invalid OCSP fetch URL: %s", urlbuf);
+    goto end;
+  }
+  if (strcasecmp(host, last_timeout_host) == 0)
+  {
+    // skip if we got timeout on the same host for previous certificate entry
+    goto end;
```
sfc-gh-igarish, line 778: nit: if we avoid goto.

sfc-gh-igarish (approved):
LGTM

deps/curl-8.1.2/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: goto is everywhere in the existing OCSP implementation. Leave it for now.

PR #593, opened 2023-10-24
SNOW-935778: Add support for unicode characters in file path in PUT/GET command

sfc-gh-ext-simba-hx:
libsnowflakeclient doesn't has the functionality for encoding conversion so it doesn't support Unicode characters by it's own.
Take the encoding conversion functionality from caller (.e.g ODBC driver) to support Unicode characters in file path for PUT/GET command.


sfc-gh-igarish (approved):
LGTM

PR #594, opened 2023-10-25
SNOW-950725: remove unused arrow code causing snyk issue

sfc-gh-ext-simba-hx:
We actually only use the code under cpp folder. Remove the code with all other languages.
Deleted following language specific folders:
- c_glib
- csharp
- go
- java
- js
- matlab
- python
- r
- ruby
- rust


sfc-gh-igarish (approved):
More than 2K files. All related ARROW third party source code. So not reviewing it. But suggest to create a doc when we upgrade ARROW libs, how to strip it again. Merge it once all gates pass.

PR #595, opened 2023-10-25
SNOW-945857: put with GCP to skip existing file

sfc-gh-ext-simba-hx:
sdk issue 709

libsnowflakeclient itself doesn't support GCP so unit test case only. Will add functional test on ODBC side.

### DIFF
--- a/cpp/SnowflakeGCSClient.cpp
+++ b/cpp/SnowflakeGCSClient.cpp
@@ -66,6 +66,18 @@ RemoteStorageRequestOutcome SnowflakeGCSClient::upload(FileMetadata *fileMetadat
   if (!m_gcsAccessToken.empty())
   {
     std::string filePathFull = m_stageInfo->location + fileMetadata->destFileName;
+
+    //check if file exists if overwrite is not specified.
+    if (!fileMetadata->overWrite)
+    {
+      RemoteStorageRequestOutcome outcome = GetRemoteFileMetadata(&filePathFull, fileMetadata);
+      if (RemoteStorageRequestOutcome::SUCCESS == outcome)
+      {
+        CXX_LOG_DEBUG("File already exists skipping the file upload %s",
+                      fileMetadata->srcFileToUpload.c_str());
+        return RemoteStorageRequestOutcome::SKIP_UPLOAD_FILE;
+      }
+    }
     buildGcsRequest(filePathFull, url, reqHeaders);
   }
   else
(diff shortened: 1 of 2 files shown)


cpp/SnowflakeGCSClient.cpp
```diff
@@ -66,6 +66,18 @@ RemoteStorageRequestOutcome SnowflakeGCSClient::upload(FileMetadata *fileMetadat
   if (!m_gcsAccessToken.empty())
   {
     std::string filePathFull = m_stageInfo->location + fileMetadata->destFileName;
+
+    //check if file exists if overwrite is not specified.
+    if (!fileMetadata->overWrite)
```
sfc-gh-igarish, line 71: Can we add condition, if it's not presigned URL with GCP. All other cloud provider is fine. i.e. condition like if(!overwrite && ((GCP && !presigned url) || AWS || AZURE)) then check for file exist.

cpp/SnowflakeGCSClient.cpp
```diff
@@ -134,12 +146,21 @@ RemoteStorageRequestOutcome SnowflakeGCSClient::GetRemoteFileMetadata(
...
-                             true))
+  try
+  {
+    if (!m_statement->http_get(url,
+                               reqHeaders,
+                               NULL,
+                               headerString,
+                               true))
+    {
+      return RemoteStorageRequestOutcome::FAILED;
```
sfc-gh-igarish, line 157: FAILED can happend when file exist or some IOException happen. How to differentiate it? right now as coded any exception treat as file exist and just SKIPPED.

cpp/SnowflakeGCSClient.cpp
sfc-gh-ext-simba-hx: We have the correct condition already.
SnowflakeGCSClient.cpp is implementation for GCP only. It's under if `(!m_gcsAccessToken.empty())` which means not using presigned URL, so only need to add check for overwrite

cpp/SnowflakeGCSClient.cpp
sfc-gh-ext-simba-hx: SKIPPED would be the case when SUCCESS on getting metadata. If there is anything else wrong the uploading would fail with the same reason and error could be thrown there.

sfc-gh-igarish (approved):
LGTM

PR #601, opened 2023-10-31
more banned function replacement

sfc-gh-ext-simba-jz:
getenv and strerror are banned functions. The replacement allocates extra memory that should be released later.


sfc-gh-ext-simba-hx:
related changed already merged in other PRs. Close

PR #602, opened 2023-10-31
replace ms-banned functions part2

sfc-gh-ext-simba-jz:
getenv and strerror are two banned functions, the replacement allocate memory that need to be released by caller.

(in part1 we only put stubs for the replacement but no effective change)

This is 2nd part of solution of issue #708, SNOW-945768: Addressing Banned functions in ODBC driver


sfc-gh-igarish (approved):
LGTM

PR #604, opened 2023-11-02
fix test case failures on jenkins Windows run

sfc-gh-ext-simba-hx:
On Windows test on Jenkins, FindFirstFile() returns unexpected error ERROR_PATH_NOT_FOUND and break some test cases.
We only handle ERROR_FILE_NOT_FOUND before but since ERROR_PATH_NOT_FOUND also seems to be a reasonable error handle it same as ERROR_FILE_NOT_FOUND.


sfc-gh-igarish (approved):
LGTM

PR #564, opened 2023-09-26
SNOW-918603: Fix the test failures we commented out recently

sfc-gh-ext-simba-hx:
related SDK issue: 658, 340

It's a potential bug that made 3 years ago https://github.com/snowflakedb/libsnowflakeclient/pull/207 .
Worked fine until we got some unit test case hang recently.
Has been there for a long time and no customer report so far likely not a customer facing issue.

PR #605, opened 2023-11-03
SNOW-918603: temporarily comment out hang test case on Jenkins Linux

sfc-gh-ext-simba-hx:
Turns out the fix we made in https://github.com/snowflakedb/libsnowflakeclient/pull/564 doesn't fix the hanging issue on Jenkins Linux run. Comment out the test cases again and will follow with sdk issue 658, 340


sfc-gh-igarish (approved):
LGTM

PR #652, opened 2024-01-08
SNOW-968730: Increased max LOB size complete testing

sfc-gh-ext-simba-hx:
sdk issue 751

This PR is for testing only. make it draft and won't merge for now.
Need to increase MAX_LOB_SIZE definition to test on Snowflake side.
The test case could be merged when the feature is stable and can be released, but at that time we might need to reduce the test cases (especially the number of sizes being tested) as currently it might take too much time to run due to the large data size.

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -41,7 +41,9 @@ SET(TESTS_C
         test_get_query_result_response
         test_get_describe_only_query_result
         test_stmt_functions
-        test_lob
+# FEATURE_INCREASED_MAX_LOB_SIZE_IN_MEMORY is internal switch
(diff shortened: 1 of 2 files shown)


sfc-gh-dprzybysz (approved):

sfc-gh-ext-simba-hx:
@sfc-gh-dprzybysz 
Just to confirm we can merge this PR, right? The test result looks good and the new test case is commented out so it won't break anything.

sfc-gh-dprzybysz:
> @sfc-gh-dprzybysz Just to confirm we can merge this PR, right? The test result looks good and the new test case is commented out so it won't break anything.

@sfc-gh-ext-simba-hx yes, you can merge it

PR #655, opened 2024-01-11
 SNOW-983106: libsnowflakeclient - possible resource leak due to exit call

sfc-gh-ext-simba-jz:
When errors occur in allocation memory, instead of calling exit() to abort the process,

it is more elegant as a library to throw an exception and let the application/caller make decision.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ set(SOURCE_FILES
         lib/rbtree.h
         lib/rbtree.c
         lib/memory.h
-        lib/memory.cpp
+        lib/memory.c
         lib/connection.h
         lib/connection.c
         lib/constants.h
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,7 +115,7 @@ set(SOURCE_FILES
         lib/rbtree.h
         lib/rbtree.c
         lib/memory.h
-        lib/memory.c
+        lib/memory.cpp
         lib/connection.h
         lib/connection.c
         lib/constants.h
--- a/cpp/util/SnowflakeCommon.cpp
+++ b/cpp/util/SnowflakeCommon.cpp
@@ -133,7 +133,7 @@ void STDCALL sf_memory_error_handler()
   if (exception_on_memory_error)
     throw std::bad_alloc();
   
-  exit(1);
+  exit(EXIT_FAILURE);
 }
 
 }
--- a/cpp/util/SnowflakeCommon.cpp
+++ b/cpp/util/SnowflakeCommon.cpp
@@ -15,7 +15,7 @@
 using namespace Snowflake;
 using namespace Snowflake::Client;
 
-static bool exit_on_memory_error = true;
+static bool exception_on_memory_error = false;
 
 /**
  * Validate partner application name.
@@ -122,18 +122,18 @@ CURLcode set_curl_proxy(CURL *curl, const char* proxy, const char* no_proxy)
   }
 }
 
-void STDCALL sf_disable_exit_on_memory_failure()
+void STDCALL sf_exception_on_memory_failure()
 {
   // should be called by odbc driver snowflake_global_init
-  exit_on_memory_error = false;
+  exception_on_memory_error = true;
 }
 
 void STDCALL sf_memory_error_handler()
 {
-  if (exit_on_memory_error)
-    exit(1);
-
-  throw std::runtime_error("fail to alloc memory");
+  if (exception_on_memory_error)
+    throw std::bad_alloc();
+  
+  exit(1);
 }
 
 }
--- a/cpp/util/SnowflakeCommon.cpp
+++ b/cpp/util/SnowflakeCommon.cpp
@@ -10,9 +10,12 @@
 #include "snowflake/platform.h"
 #include "snowflake/Proxy.hpp"
 #include "../logger/SFLogger.hpp"
+#include <stdexcept>
 
 using namespace Snowflake;
 using namespace Snowflake::Client;
+
+static bool exit_on_memory_error = true;
 
 /**
  * Validate partner application name.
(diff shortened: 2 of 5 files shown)


CMakeLists.txt
```diff
@@ -115,7 +115,7 @@ set(SOURCE_FILES
         lib/rbtree.h
         lib/rbtree.c
         lib/memory.h
-        lib/memory.c
+        lib/memory.cpp
```
sfc-gh-ext-simba-hx, line 118: Please look around and make your changes inline with existing implementation. The wrapper should be in cpp folder and added to source list of SOURCE_FILES_CPP_WRAPPER.
I would prefer to have a wrapper on malloc directly, in https://github.com/snowflakedb/libsnowflakeclient/blob/master/cpp/util/SnowflakeCommon.cpp
or if you want create a new file,
check and throw exception there, and call the C wrapper in memory.c (no need to change the existing file name)
In such case we don't need to check the return value in memory.c since it's been checked in the wrapper already.

Please also make the behavior configurable otherwise it would be a breaking change for libsnowflakeclient, changing the error handling from exit to throw exception. That's good for ODBC but might be an issue for any other application using libsfc that doesn't expect exception.

cpp/util/SnowflakeCommon.cpp
```diff
@@ -119,5 +122,19 @@ CURLcode set_curl_proxy(CURL *curl, const char* proxy, const char* no_proxy)
...
+  // should be called by odbc driver snowflake_global_init
+  exit_on_memory_error = false;
+}
+
+void STDCALL sf_memory_error_handler()
+{
+  if (exit_on_memory_error)
+    exit(1);
+
+  throw std::runtime_error("fail to alloc memory");
```
sfc-gh-ext-simba-hx, line 136: Would that be easily handled on ODBC side?
 `throw std::bad_alloc()` in existing implementation would be better than this one. Or maybe we need a new specific exception (could take SnowflakeTransferException as an example) so it could be easier for ODBC to catch and handle it?
 Please check existing implementation and make your changes inline with it.

cpp/util/SnowflakeCommon.cpp
```diff
@@ -10,10 +10,13 @@
 #include "snowflake/platform.h"
 #include "snowflake/Proxy.hpp"
 #include "../logger/SFLogger.hpp"
+#include <stdexcept>
 
 using namespace Snowflake;
 using namespace Snowflake::Client;
 
+static bool exit_on_memory_error = true;
```
sfc-gh-ext-simba-hx, line 18: I would refer the name of `exception_on_memory_error`. exit is not good and it could be fixed in the future for C API as well, while it won't use exception.

cpp/util/SnowflakeCommon.cpp
sfc-gh-ext-simba-jz: bad_alloc is fine. By the way, what do you men inlinez?

sfc-gh-ext-simba-hx: I mean check the existing implementation first before making changes, try do the same/similar thing in the same/similar way, which would have lower risk to break anything.

include/snowflake/platform.h
```diff
@@ -144,6 +144,11 @@ void STDCALL sf_get_username(char * username, int bufLen);
 
 void STDCALL sf_delete_uniq_dir_if_exists(const char *tmpfile);
 
+void STDCALL sf_memory_error_handler();
+
+// this should be called by odbc driver before any calls of sfclient
```
sfc-gh-ext-simba-hx, line 149: make the comment more general about what's the functionality or no comment as other functions. This function could be called in other case as well if the application prefer to get exception on memory allocation error.

cpp/util/SnowflakeCommon.cpp
```diff
@@ -119,5 +122,19 @@ CURLcode set_curl_proxy(CURL *curl, const char* proxy, const char* no_proxy)
...
+void STDCALL sf_exception_on_memory_failure()
+{
+  // should be called by odbc driver snowflake_global_init
+  exception_on_memory_error = true;
+}
+
+void STDCALL sf_memory_error_handler()
+{
+  if (exception_on_memory_error)
+    throw std::bad_alloc();
```
sfc-gh-ext-simba-hx, line 134: Please make odbc side change and post your branch as well. I'd like to see the changes there to ensure this throw can be handled easily there and the relative change there is minor.

cpp/util/SnowflakeCommon.cpp
```diff
@@ -10,10 +10,13 @@
 #include "snowflake/platform.h"
 #include "snowflake/Proxy.hpp"
 #include "../logger/SFLogger.hpp"
+#include <stdexcept>
 
 using namespace Snowflake;
 using namespace Snowflake::Client;
 
+static bool exception_on_memory_error = false;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-hx: @sfc-gh-dprzybysz 
Sorry for the delay. @sfc-gh-ext-simba-jz is on vacation I'll take over.
Yes by default (for libsfc user other than odbc) it's false so no behavior change. For ODBC it would call sf_exception_on_memory_failure() to turn on exception on memory failures. ODBC is made with C++ and can handle exception well. The exception will cause ODBC API call to return SQL_ERROR instead of calling exit() to terminate customer application.

cpp/util/SnowflakeCommon.cpp
```diff
@@ -119,5 +122,19 @@ CURLcode set_curl_proxy(CURL *curl, const char* proxy, const char* no_proxy)
...
 
+void STDCALL sf_exception_on_memory_failure()
+{
+  // should be called by odbc driver snowflake_global_init
+  exception_on_memory_error = true;
+}
+
+void STDCALL sf_memory_error_handler()
+{
+  if (exception_on_memory_error)
```
sfc-gh-dprzybysz, line 133: let's add `{}` for if body

sfc-gh-dprzybysz (changes requested):

cpp/util/SnowflakeCommon.cpp
sfc-gh-ext-simba-hx: done

cpp/util/SnowflakeCommon.cpp
sfc-gh-ext-simba-hx: Checked implementation on ODBC side currently it would never trigger the functions defined in memory.c.
Will call sf_memory_error_handler() on ODBC side for possible further extension but for now no visible impact with that.

sfc-gh-dprzybysz (approved):

PR #656, opened 2024-01-11
SNOW-937196: make correction on arrow version for uploading

sfc-gh-ext-simba-hx:
Fix the arrow version for artifacts uploading since the package on S3 for debug version seems broken and we need to re-upload.


sfc-gh-dstempniak (approved):
LGTM

PR #658, opened 2024-01-26
SNOW-983104: build with vs2022

sfc-gh-ext-simba-hx:
sdk issue 783
- Add build/test with vs2022 and required build flags (CFG, Qspectre, SDL, CETCOMPAT) for all components.
- CETCOMPAT is linking option so won't apply to static libs. Add it when linking option is available in build script just in case.
- openssl doesn't support CFG https://github.com/openssl/openssl/issues/22554
- add /wd4996 for azure to suppress the build error of deprecated function calls hopefully we could get that fixed when updating azure sdk to higher version.
- remove deprecated function call in libsnowflakeclient


scripts/_init.bat
```diff
@@ -61,6 +61,10 @@ if "%target_name%"=="" (
 
 set cmake_generator=
 set vsdir=
+if /I "%vs_version%"=="VS17" (
```
sfc-gh-dprzybysz, line 64: let's update README also https://github.com/snowflakedb/libsnowflakeclient?tab=readme-ov-file#windows

sfc-gh-dprzybysz (approved):

scripts/_init.bat
sfc-gh-ext-simba-hx: done

lib/platform.c
```diff
@@ -627,21 +628,35 @@ void STDCALL sf_os_version(char *ret, size_t size) {
...
-    //     6.1  Windows  7.0/Server 2008 R2
-    //     6.0  Windows  Vist/Server 2008 R1 (Not supported)
-    OSVERSIONINFOEX info;
-    ZeroMemory(&info, sizeof(OSVERSIONINFOEX));
-    info.dwOSVersionInfoSize = sizeof(OSVERSIONINFOEX);
-    GetVersionEx((LPOSVERSIONINFO)&info);
+    int majorVersion = 0;
+    int minorVersion = 0;
+
+    if (IsWindows10OrGreater())
```
sfc-gh-dprzybysz, line 634: our jenkins builds report problem with this function on VS14:
```
2024-02-07 03:35:57 snowflakeclient.lib(platform.obj) : error LNK2019: unresolved external symbol _IsWindows10OrGreater referenced in function _sf_os_version@8
```

sfc-gh-dprzybysz (changes requested):

lib/platform.c
sfc-gh-ext-simba-hx: @sfc-gh-dprzybysz 
It should be fixed with my changes.
I can't reproduce it on my local though since I've had vs2022 installed and the windows sdk updated to 10.x.
Please rerun the build on jenkins to ensure the issue is solved. It would take some effort for me to get a clear environment that  only has Windows sdk 8.1 available.

sfc-gh-dprzybysz: The build was successful on jenkins

sfc-gh-dprzybysz (approved):

PR #659, opened 2024-02-06
SNOW-983100: replace banned functions

sfc-gh-ext-simba-hx:
sdk issue 782, replace banned function in security review
- refactoring on sb_ functions mainly adding meaningful replacement on non-Windows
- deprecate sf_getenv and sf_strerror and having replacement of sf_getenv_s and sf_strerror_s passing in buffer by caller.


cpp/FileTransferAgent.cpp
```diff
@@ -19,7 +19,7 @@
 #include "EncryptionProvider.hpp"
 #include "logger/SFLogger.hpp"
 #include "snowflake/platform.h"
-#include "snowflake/Simba_CRTFunctionSafe.h"
+#include "snowflake/sb_CRTFunctionSafe.h"
```
sfc-gh-dprzybysz, line 22: why do we rename this file? isn't it kind of our public API?

cpp/FileTransferAgent.cpp
```diff
@@ -847,20 +847,20 @@ RemoteStorageRequestOutcome Snowflake::Client::FileTransferAgent::downloadSingle
      }
      catch (...) {
        std::string err = "Could not open file " + fileMetadata->destPath + " to downoad";
-       char* str_error = sf_strerror(errno);
+       char strerr_buf[1024];
```
sfc-gh-dprzybysz, line 850: for the error buffer we set size to 1024 in many places - should we create a constant from it?

tests/test_unit_proxy.cpp
```diff
@@ -107,11 +95,7 @@ void test_httpproxy_fromenv(void **unused)
 
 void test_noproxy_fromenv(void **unused)
 {
-    if (sf_getenv("all_proxy") || sf_getenv("https_proxy") ||
-        sf_getenv("http_proxy")) {
-        // skip the test if the test environment uses proxy already
-        return;
-    }
+    SKIP_IF_PROXY_ENV_IS_SET;
```
sfc-gh-dprzybysz, line 98: sometimes we use this constant with `;` sometimes not - let's pick one format

include/snowflake/platform.h
```diff
@@ -56,21 +56,13 @@ void STDCALL sf_tzset(void);
 
 int STDCALL sf_setenv(const char *name, const char *value);
 
-/* on Windows, this function allocate new memory, caller should free it */
-char *STDCALL sf_getenv(const char *name);
+char *STDCALL sf_getenv(const char *name, char *outbuf, size_t bufsize);
```
sfc-gh-dprzybysz, line 59: we are changing the public API of libsnowflakeclient - should we do this?

cpp/FileTransferAgent.cpp
sfc-gh-ext-simba-hx: remove Simba from the file name. We can keep the existing name though please let me know which one you prefer.

include/snowflake/platform.h
sfc-gh-ext-simba-hx: This is utility function used in libsnowflakeclient internally.
I don't think external application would rely on that (like sb_ functions, it's available in standard c lib getenv, strcpy, strcat...)
Here we don't have perfect secure replacement for getenv cross all platforms with exactly the same interface.
Having buffer allocation internally would require free on application side and have thread-safe issue as well.
I think changing the interface to force the application take care of the buffer is a better option compare to commenting in header file to ask for free after calling this function.

cpp/FileTransferAgent.cpp
sfc-gh-ext-simba-hx: done

tests/test_unit_proxy.cpp
sfc-gh-ext-simba-hx: done

include/snowflake/sb_CRTFunctionSafe.h
```diff
@@ -1,13 +1,9 @@
-// =================================================================================================
-///  @file Simba_CRTFunctionSafe.h
-///
-///  C Run-Time library functions' definitions for cross platform use.
-///
-///  Copyright (C) 2019 Simba Technologies Incorporated.
-// =================================================================================================
+/*
+ * Copyright (c) 2019-2024 Snowflake Computing, Inc. All rights reserved.
```
sfc-gh-dprzybysz, line 2: is it right to change the copyright here? if so then we could name the file with `sf_` prefix not `sb_`?

changing the file is breaking compatibility and we should consider it as a breaking change and release next version with bumped major part

include/snowflake/sb_CRTFunctionSafe.h
```diff
@@ -21,186 +17,257 @@
...
+
+    /// @brief Copy bytes between buffers.
+    ///
+    /// @param out_dest         Destination buffer. (NOT OWN)
+    /// @param in_destSize      Size of the destination buffer.
+    /// @param in_src           Buffer to copy from. (NOT OWN)
+    /// @param in_sizeToCopy    Number of bytes to copy.
+    ///
+    /// @return A pointer to destination; NULL if an error occurred. (NOT OWN)
+    static inline void* sb_memcpy(
```
sfc-gh-dprzybysz, line 40: should it be `sb_` or `sf_`? and why this method declaration has additional indentation?

include/snowflake/sb_CRTFunctionSafe.h
```diff
@@ -21,186 +17,257 @@
...
+/// @brief Copy a string.
+    ///
     /// @param out_dest         Destination string buffer. (NOT OWN)
     /// @param in_destSize      Size of the destination string buffer.
     /// @param in_src           Null-terminated source string buffer. (NOT OWN)
-    /// 
+    ///
     /// @return The destination string; NULL if an error occurred. (NOT OWN)
-    inline char* sb_strcpy(
+    static inline char* sb_strcpy(
```
sfc-gh-dprzybysz, line 87: should it be `sf_` or `sb_`? it's also the question for other places in this file

sfc-gh-dprzybysz:
@sfc-gh-ext-simba-hx SNOW-983110 is not the right id, it should be SNOW-983100

cpp/FileTransferAgent.cpp
sfc-gh-ext-simba-hx: reverted. Will handle that with a separated issue and PR

include/snowflake/platform.h
sfc-gh-ext-simba-hx: keep the existing interface unchanged and mark it deprecated to keep the backward compatibility.

include/snowflake/sb_CRTFunctionSafe.h
sfc-gh-ext-simba-hx: No change on this for now handle it in separated issue/PR

include/snowflake/platform.h
sfc-gh-dprzybysz: there are no depractions in this file - the old methods are still missing

include/snowflake/Simba_CRTFunctionSafe.h
```diff
@@ -21,186 +17,269 @@
...
+#define MACRO_TO_STRING(x) MACRO_TO_STRING2(x)
 
-#if defined(WIN32) || defined(_WIN64)   // For Windows
+#if defined(_MSC_VER)
+#define SF_MACRO_DEPRECATED_WARNING(MSG) __pragma(message ( __FILE__ "(" MACRO_TO_STRING(__LINE__) ") : warning C4996: " MSG))
+#else
+#define SF_MACRO_DEPRECATED_WARNING(MSG)  _Pragma(MACRO_TO_STRING(GCC warning MSG))
+#endif
 
-    #define sf_free_s(x)        free(x)
```
sfc-gh-dprzybysz, line 38: we are removing this define - should it be also deprecated?

sfc-gh-ext-simba-hx: Will make it deprecated as well just in case.
This is added recently https://github.com/snowflakedb/libsnowflakeclient/pull/602 with the problematic implementation of sf_getenv and sf_strerror and should not be there in the first place.
I don't think external user could use it in proper way. It can only be used to free buffer returned from  sf_getenv/sf_strerror and can't be used as regular free (since it does nothing on non-Windows).

sfc-gh-ext-simba-hx: done

include/snowflake/platform.h
sfc-gh-ext-simba-hx: platform.h includes Simba_CRTFunctionSafe.h at the top so it can get deprecations from there.
https://github.com/snowflakedb/libsnowflakeclient/pull/659/files#diff-25951124eb05ec39b13d5c29ce2c708d01d9363367ee8e7c1a946e1774cddc6eL49

sfc-gh-dprzybysz (approved):

PR #665, opened 2024-02-12
SNOW-989449: porting custom changes in curl and update build scripts

sfc-gh-ext-simba-hx:
sdk issue 801
second PR to port custom changes


deps/curl-8.6.0/lib/curl_addrinfo.c
```diff
@@ -47,6 +47,10 @@
 #  include <inet.h>
 #endif
 
+#if defined(USE_THREADS_POSIX) && defined(HAVE_PTHREAD_H)
```
sfc-gh-dprzybysz, line 50: this if condition is repeated many times - can we create one define for it?

sfc-gh-dprzybysz (approved):

deps/curl-8.6.0/lib/curl_addrinfo.c
sfc-gh-ext-simba-hx: Such condition is used in many places in original curl code like here
https://github.com/snowflakedb/libsnowflakeclient/blob/master/deps/curl-8.6.0/lib/easy_lock.h#L96
so I think better to keep it unchanged.

PR #667, opened 2024-02-14
SNOW-1056426: rename CRT functions

sfc-gh-ext-simba-hx:
sdk issue 873
Having changes in #659 to pass build. Will rebase after that's merged.
The actual changes are in the last commit: https://github.com/snowflakedb/libsnowflakeclient/pull/667/commits/b8898f563479257269da81ac27a51ec92a3ab6c1


include/snowflake/platform.h
```diff
@@ -46,7 +46,7 @@ typedef pthread_mutex_t SF_MUTEX_HANDLE;
 
 #endif
 
-#include "Simba_CRTFunctionSafe.h"
+#include "SF_CRTFunctionSafe.h"
```
sfc-gh-dprzybysz, line 49: to be compatible we should still include here `Simba_CRTFunctionSafe.h` because when include only `platform.h` then `sb_*` operations are missing?

sfc-gh-dprzybysz (changes requested):

sfc-gh-dprzybysz (approved):

PR #675, opened 2024-03-05
SNOW-983098: add original source code for arrow 15.0.0 and boost 1.81.0

sfc-gh-ext-simba-hx:
teamwork issue 780
Add original source code for arrow 15.0.0 and boost 1.81.0 which is the default boost version used by arrow 15.0.0 as well.
Removed source code for all other programing language in arrow:
- c_glib
- csharp
- go
- java
- js
- matlab
- python
- r
- ruby
- swift

Will have following PRs to update build scripts and docker image.


sfc-gh-dprzybysz (approved):
Thanks @sfc-gh-ext-simba-hx 
do you think we can also remove from the repo:
- deps/arrow-15.0.0/dev
- unused libraries from boost?

sfc-gh-ext-simba-hx:
> Thanks @sfc-gh-ext-simba-hx do you think we can also remove from the repo:
> 
> * deps/arrow-15.0.0/dev
> * unused libraries from boost?

I think that's possible but we would need to do that each time updating arrow/boost. The main purpose of removing code for other programing languages from arrow is to prevent false alarm from there in security check that we had before.
If no special reason I would prefer to keep the updating process simple.

PR #677, opened 2024-03-08
SNOW-983098: update centos7 docker image with gcc8.3

sfc-gh-ext-simba-hx:
arrow 15.x needs to be built with c++17 standard and need gcc update.

### DIFF
--- a/ci/_init.sh
+++ b/ci/_init.sh
@@ -23,7 +23,14 @@ TEST_IMAGE_VERSION_X64=1
 TEST_IMAGE_VERSION_AARCH64=3
 
 PLATFORM_ARCH=$(uname -p)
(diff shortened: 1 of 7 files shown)


sfc-gh-dprzybysz (approved):

ci/_init.sh
```diff
@@ -15,15 +15,22 @@ fi
 export DRIVER_NAME=libsnowflakeclient
 
 # Build images
-BUILD_IMAGE_VERSION_X64=1
+BUILD_IMAGE_VERSION_X64=2
```
sfc-gh-dprzybysz, line 18: let's change also `Dockerfile.libsnowflakeclient-centos7-default-test` to point to the new image

PR #681, opened 2024-03-12
SNOW-983096: add original source code for aws-sdk-cpp 1.11.283

sfc-gh-ext-simba-hx:
Will make follow up PRs to build with awssdk 1.11.283
It's a big change from 1.3.50 to 1.11.283 would need changes on both build script and source code.
Removed following unused files, would need to do the same for each version up afterwards:
- tests folders under root, generated, crt/aws-crt-cpp, crt/aws-crt-cpp/crt/*
- crt/aws-crt-cpp/crt/aws-lc/fuzz
- all subfolders under generated besides aws-cpp-sdk-s3*


sfc-gh-dprzybysz:
@sfc-gh-ext-simba-hx The PR is so big that GH is not able to show the diff - are we able to get rid of unnecessary files (e.g. tests, docs, ci scripts) and/or modules we con't need?

sfc-gh-dprzybysz:
@sfc-gh-ext-simba-hx Please also take a look at semgrep action output - we should have it green to merge

sfc-gh-dprzybysz (approved):

PR #689, opened 2024-03-21
SNOW-1007447: fixes for building issues when moving to new jenkins node

sfc-gh-ext-simba-hx:
After moving to new jenkins node some third-party libs in default searching path breaks the build. Adding NO_DEFAULT_PATH to all find_library() to avoid such issue.


sfc-gh-dstempniak (approved):

ci/build_win.bat
```diff
@@ -13,8 +13,11 @@ set curdir=%cd%
 set utils_script="%scriptdir%..\scripts\utils.bat"
 call %utils_script% :init_git_variables
 if %ERRORLEVEL% NEQ 0 goto :error
-pip install --upgrade --user awscli
-pip install awscli
+echo === creating venv
+py -3.7 -m venv venv
+call venv\scripts\activate
+python -m pip install -U pip > nul 2>&1
+python -m pip install -U awscli > nul 2>&1
```
sfc-gh-aszczepankiewicz, line 20: Thanks, that's what we need here. 

nit: Is awscli the only pip dependency installed? It is available globally on the new nodes, so you can also try just not installing it.

sfc-gh-aszczepankiewicz:
@sfc-gh-ext-simba-hx Can we also remove this? https://github.com/snowflakedb/libsnowflakeclient/pull/689/files#diff-bd550d0e48092dbff35bd5743177dd853a2ecbde2d169f2b6131b282e67ee9c6R8
7z is available in the PATH (and in a different location). Python is installed in a different location, and you already use it correctly with py -3.7.

PR #708, opened 2024-04-17
SNOW-983098: redo arrow update 15.0.0

sfc-gh-ext-simba-hx:
teamwork issue 780
temporarily add build number in libsnowflakeclient version to avoid replacing artifacts on s3 built with gcc5. Will remove when ODBC also move to gcc8.


sfc-gh-ext-simba-hx:
The SEGFAULT with Mac debug build is related to static regex defined in https://github.com/snowflakedb/libsnowflakeclient/blob/master/cpp/logger/SecretDetector.cpp when calling ~mem_block_cache at the ending of test application. Will get that fixed before asking for review.

sfc-gh-ext-simba-hx:
Confirmed no performance issue with adding `-DBOOST_REGEX_MAX_CACHE_BLOCKS=0`
regex is used by arrow also for logging in libsfc directly.
Tested with ODBC driver using testapp in teamwork issue 697 (fetch a large result using SQLGetData) with both loglevel=6 (1GB log generated) and loglevel=0, compared with 3.3.0, no performance issue found.
new build vs 3.3.0 in seconds
loglevel=6:   94.043 / 112.014
loglevel=0:   7.344 / 7.401

include/snowflake/version.h
```diff
@@ -5,6 +5,6 @@
 #ifndef SNOWFLAKE_CLIENT_VERSION_H
 #define SNOWFLAKE_CLIENT_VERSION_H
 
-#define SF_API_VERSION "1.0.10"
+#define SF_API_VERSION "1.0.10.1"
```
sfc-gh-dprzybysz, line 8: why do we bump the version here? cannot we use 1.0.10?

sfc-gh-ext-simba-hx: changed back.

sfc-gh-dprzybysz (approved):

PR #709, opened 2024-04-24
force to use odbc image for debugging

sfc-gh-ext-simba-hx:
for debugging purpose. Never merge.


sfc-gh-ext-simba-hx:
Close as debugging finished.

PR #726, opened 2024-06-05
SNOW-1325227 login to nexus

sfc-gh-akolodziejczyk:


ci/test_linux.sh
```diff
@@ -36,6 +36,7 @@ else
     TEST_IMAGE_NAME="${TEST_IMAGE_NAMES[$TARGET_DOCKER_TEST_IMAGE]}"
 fi
 echo $TEST_IMAGE_NAME
+echo "$NEXUS_PASSWORD" | docker login --username jenkins --password-stdin
```
sfc-gh-dprzybysz, line 39: we have script to login to docker internally https://github.com/snowflakedb/libsnowflakeclient/blob/master/ci/scripts/login_internal_docker.sh
can we use it instead?

sfc-gh-akolodziejczyk: yes, this line was not working, I will try with the script

sfc-gh-dprzybysz (approved):

PR #732, opened 2024-07-16
SNOW-1526511: unhardcode top level domain

sfc-gh-ext-simba-hx:
teamwork issue 1046


sfc-gh-dprzybysz (approved):

lib/connection.c
```diff
@@ -594,34 +593,29 @@ char * STDCALL encode_url(CURL *curl,
     const char *amp = "&";
     size_t amp_size = strlen(amp);
 
+    if (host_empty)
+    {
+      SET_SNOWFLAKE_ERROR(error, SF_STATUS_ERROR_BAD_CONNECTION_PARAMS,
+                          "Invalid host trying to create encoded url",
+                          SF_SQLSTATE_UNABLE_TO_CONNECT);
+      goto cleanup;
+    }
```
sfc-gh-dstempniak, line 602: What about 6th point?
```
detected top level domain should be logged with a respective message at the INFO level:
Connecting to GLOBAL Snowflake domain
or
Connecting to CHINA Snowflake domain
respectively.
```

sfc-gh-ext-simba-hx: oops I missed that. Will add.

sfc-gh-ext-simba-hx: added

deps/curl/lib/vtls/sf_ocsp.c
```diff
@@ -2244,10 +2271,33 @@ void initOCSPCacheServer(struct Curl_easy *data)
...
+    else
+    {
+      // It's basically impossible not finding top domain in host.
+      // Use "com" as default just in case.
+      top_domain = "com";
+    }
+
+    if (strcasecmp(top_domain, "com") == 0)
+    {
+      infof(data, "Connecting to GLOBAL Snowflake domain");
```
sfc-gh-dstempniak, line 2288: What about private links? 
If `ocsp_cache_server_url_env != NULL` and `host` is specified there will be no info message at all, am I right?

lib/client.c
```diff
@@ -432,11 +437,23 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
...
+                sf_sprintf(buf, sizeof(buf), "%s.%s.snowflakecomputing.cn",
+                         sf->account, sf->region);
+                log_info("Connecting to CHINA Snowflake domain");
+            }
+            else
+            {
+                sf_sprintf(buf, sizeof(buf), "%s.%s.snowflakecomputing.com",
+                         sf->account, sf->region);
+                log_info("Connecting to GLOBAL Snowflake domain");
+            }
```
sfc-gh-dstempniak, line 452: Maybe we could move this check outside `if (!sf->host) {..}` condition and always check postfix of `sf->host`?
In such a case we will have only one `if .. else ...` condition and there will be no need to log similar information in `sf_ocsp.c`.

deps/curl/lib/vtls/sf_ocsp.c
sfc-gh-dprzybysz: We have a separate issue for private links support SNOW-1538117

lib/client.c
sfc-gh-ext-simba-hx: good idea. done.

lib/client.c
```diff
@@ -432,14 +438,45 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
...
+        top_domain++;
+    }
+    else
+    {
+        // It's basically impossible not finding top domain in host.
+        // Log the entire host just in case.
+        top_domain = sf->host;
+    }
+
+    if (strcasecmp(top_domain, "com") == 0)
```
sfc-gh-dprzybysz, line 471: please change the log to expose CHINA and GLOBAL only like in JDBC https://github.com/snowflakedb/snowflake-jdbc/blob/60e9e541794bbd80a33b00cf8039c8512d8c5361/src/main/java/net/snowflake/client/core/SFSession.java#L625-L627

sfc-gh-dprzybysz (changes requested):

lib/client.c
sfc-gh-ext-simba-hx: done

sfc-gh-dprzybysz (approved):

sfc-gh-dstempniak (approved):

PR #736, opened 2024-08-09
SNOW-1541757: remove LIBSFCLI_FOR_XP

sfc-gh-ext-simba-hx:
teamwork issue 1060


tests/test_error_handlings.c
```diff
@@ -31,11 +31,15 @@ void test_incorrect_password(void **unused) {
     SF_STATUS status = snowflake_connect(sf);
     assert_int_not_equal(status, SF_STATUS_SUCCESS); // must fail
 
+
+/* temporarily disable error code check as currently it's returning
+ * error of IP address not allowed due to server change on test account
```
sfc-gh-dstempniak, line 36: I think this explains my previous question :)

CMakeLists.txt
```diff
@@ -29,11 +29,6 @@ set(OPENSSL_VERSION_NUMBER  0x11100000L)
 # Generates compile_commands.json file for clangd to parse.
 set(CMAKE_EXPORT_COMPILE_COMMANDS ON)
 
-if ($ENV{XP_BUILD})
```
sfc-gh-dstempniak, line 32: Don't we want to remove `XP_BUILD` support as well?

sfc-gh-ext-simba-hx: hmm good question. Now XP build is using a separated repo so maybe we could remove `XP_BUILD` from all the build scripts? I think better to leave `XP_BUILD` unchanged for now to be safe, we could have a separated PR for that once confirmed.

PR #738, opened 2024-08-16
SNOW-1524269: support put/get for GCP

sfc-gh-ext-simba-hx:
teamwork issue 1035
part 1 for GCP support, will have separated PR to support PUT/GET natively.
Draft for now as temporary changes included to fake the client to be ODBC to pass test. Without that we would get error of "gcs downscoped token not supported" from server.
We would need server change to enable PUT/GET on GCS for libsnowflakeclient (maybe for PHP as well).


cpp/StatementPutGet.cpp
```diff
@@ -123,3 +170,124 @@ Util::Proxy* StatementPutGet::get_proxy()
...
+                         0, sf->retry_count, NULL, NULL, NULL, SF_BOOLEAN_FALSE,
+                         sf->proxy, sf->no_proxy, SF_BOOLEAN_FALSE, SF_BOOLEAN_FALSE);
+
+  free_curl_desc(curl_desc);
+  SF_FREE(urlbuf);
+  curl_slist_free_all(reqHeaders.header);
+  if (respHeaders)
+  {
+    responseHeaders = respHeaders;
+    SF_FREE(respHeaders);
```
sfc-gh-dprzybysz, line 223: should we free response headers when we pass them as a out parameter?

include/snowflake/client.h
```diff
@@ -18,7 +18,7 @@ extern "C" {
 /**
  * API Name
  */
-#define SF_API_NAME "C API"
+#define SF_API_NAME "ODBC"
```
sfc-gh-dprzybysz, line 21: it's temporal - please add todo comment to change before merging to master

lib/client.c
```diff
@@ -1527,6 +1527,7 @@ sf_put_get_response_deallocate(SF_PUT_GET_RESPONSE *put_get_response) {
 
     snowflake_cJSON_Delete((cJSON *) put_get_response->src_list);
     snowflake_cJSON_Delete((cJSON *) put_get_response->enc_mat_get);
+    snowflake_cJSON_Delete((cJSON*)put_get_response->presigned_urls);
```
sfc-gh-dprzybysz, line 1530: nitpicking: can we format the code as above

cpp/StatementPutGet.cpp
```diff
@@ -47,6 +71,17 @@ bool StatementPutGet::parsePutGetCommand(std::string *sql,
     putGetParseResponse->srcLocations.emplace_back(val->valuestring);
   }
 
+  cJSON* presignedUrls = (cJSON*)response->presigned_urls;
```
sfc-gh-dprzybysz, line 74: I will check if presigned urls are still necessary with GCP communication

sfc-gh-dprzybysz (commented):
why do we need to merge this feature to master 2.0? there are no incompatible changes with the current 1.x

lib/client.c
sfc-gh-ext-simba-hx: done

include/snowflake/client.h
sfc-gh-ext-simba-hx: done

cpp/StatementPutGet.cpp
sfc-gh-ext-simba-hx: responseHeaders is a std::string so basically here we are doing a copy. Changed the code though to make it clear.

sfc-gh-ext-simba-hx:
> why do we need to merge this feature to master 2.0? there are no incompatible changes with the current 1.x

Noted down in the ticket discussed about that with @sfc-gh-dstempniak before.

sfc-gh-dprzybysz (approved):

cpp/StatementPutGet.cpp
sfc-gh-dprzybysz: Presigned urls could be skipped in the added code

sfc-gh-dprzybysz (changes requested):
Let's remove presigned url related code

cpp/StatementPutGet.cpp
sfc-gh-ext-simba-hx: removed.

sfc-gh-dprzybysz (approved):

PR #740, opened 2024-08-20
SNOW-1526335: support OCSP FAIL_OPEN

sfc-gh-ext-simba-hx:
teamwork issue 1039

Add new boolean connection attribute `SF_CON_OCSP_FAIL_OPEN`. Can be set/get through existing APIs `snowflake_set_attribute()` and `snowflake_get_attribute()`.
Possible values are `SF_BOOLEAN_TRUE` and `SF_BOOLEAN_FALSE`. The default values is `SF_BOOLEAN_FALSE`. When it's set to `SF_BOOLEAN_TRUE` ocsp validation will take `fail-open` behavior and `fail-close` when it's `SF_BOOLEAN_FALSE`. The behavior details are documented here https://docs.snowflake.com/en/user-guide/ocsp#fail-open-or-fail-close-behavior


lib/http_perform.c
```diff
@@ -320,8 +321,6 @@ sf_bool STDCALL http_perform(CURL *curl,
             break;
         }
 
-#ifndef _WIN32
```
sfc-gh-dstempniak, line 323: Just checking  - do we want to enable it for Windows?

tests/test_error_handlings.c
```diff
@@ -30,12 +30,13 @@ void test_incorrect_password(void **unused) {
     snowflake_set_attribute(sf, SF_CON_PASSWORD, "HAHAHA");
     SF_STATUS status = snowflake_connect(sf);
     assert_int_not_equal(status, SF_STATUS_SUCCESS); // must fail
-
+/*
```
sfc-gh-dstempniak, line 33: https://github.com/snowflakedb/libsnowflakeclient/pull/741/files#r1731306175

lib/http_perform.c
sfc-gh-ext-simba-hx: That's consistent behavior cross drivers I think better to do the same in libsfclient as well
https://github.com/snowflakedb/snowflake-jdbc/blob/master/src/main/java/net/snowflake/client/core/SFBaseSession.java#L564
The application should not set it if they don't want to disable OCSP instead of having the setting ignored in libsfclient specifically on Windows only. This would be a BCR as well.
I think we can't avoid this PR to be a BCR change. Seems supporting OCSP is a BCR for all other drivers.
https://docs.snowflake.com/en/user-guide/ocsp#legacy-client-and-driver-versions

sfc-gh-dstempniak (approved):

sfc-gh-dprzybysz (approved):
If the problem is the default value then we can keep the old default FAIL_CLOSED and as part of the next major release just change the default value
it would allow merging the change even now to master

lib/client.c
```diff
@@ -525,6 +525,7 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
     log_debug("protocol: %s", sf->protocol);
     log_debug("autocommit: %s", sf->autocommit ? "true": "false");
     log_debug("insecure_mode: %s", sf->insecure_mode ? "true" : "false");
+    log_debug("ocsp_fail_open: %s", sf->ocsp_fail_open ? "true" : "false");
```
sfc-gh-dprzybysz, line 528: let's keep the default to false to not cause BCR and we can merge to master

sfc-gh-dprzybysz (changes requested):

lib/client.c
sfc-gh-ext-simba-hx: done https://github.com/snowflakedb/libsnowflakeclient/pull/740/files#diff-1573e024f58fe545f3b811283ce6ce75f7fa32eec02ef00794877e71941e9cf0R681

sfc-gh-dprzybysz (approved):

PR #747, opened 2024-10-02
Implement Easy Logging using a config file

sfc-gh-ext-simba-nl:
For SNOW-981602: Implement Easy Logging using a config file

Add several new global attributes:
SF_GLOBAL_CLIENT_CONFIG_FILE: To set the client config file so libsnowflakeclient knows where to look for the config file
SF_GLOBAL_LOG_LEVEL: To allow the user to get the log level since libsnowflakeclient handles log setup
SF_GLOBAL_LOG_PATH: To allow the user to get the log path


include/snowflake/ClientConfigParser.hpp
```diff
@@ -0,0 +1,104 @@
+/**
+ * Copyright (c) 2024 Snowflake Computing
+ */
+
+#ifndef PROJECT_CLIENTCONFIG_HPP
```
sfc-gh-ext-simba-hx, line 5: Better to follow the existing naming style. like `SNOWFAKECIENT_CONFIGPARSER_HTTP` ?

tests/test_unit_logger.c
```diff
@@ -25,6 +28,47 @@ void test_log_str_to_level(void **unused) {
 }
 
 #ifndef _WIN32
+/**
```
sfc-gh-ext-simba-hx, line 31: Please add a test case to cover the behavior in log_init(), e.g. having a config file in place, after calling snowflake_global_init() check the log settings are as expected.

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,315 @@
...
+    {
+      derivedConfigPath = binaryDirFilePath;
+      CXX_LOG_INFO("sf", "ClientConfigParser", "loadClientConfig",
+        "Using client configuration path from binary directory: %s", binaryDirFilePath.c_str());
+    }
+    else
+    {
+#if defined(WIN32) || defined(_WIN64)
+      // 4. Try user home dir
+      std::string homeDir = sf_getenv_s("USERPROFILE", envbuf, sizeof(envbuf));
```
sfc-gh-ext-simba-hx, line 103: sf_getenv_s could return NULL and cause runtime exception thrown.

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,315 @@
...
+  }
+  if (!derivedConfigPath.empty())
+  {
+    parseConfigFile(derivedConfigPath, out_clientConfig);
+  }
+}
+
+// Private =====================================================================
+////////////////////////////////////////////////////////////////////////////////
+bool ClientConfigParser::checkFileExists(const std::string& in_filePath)
```
sfc-gh-ext-simba-hx, line 144: We have boost in libsfclient I think better to use boost::filesystem::exists

lib/client.c
```diff
@@ -306,18 +309,34 @@ static sf_bool STDCALL log_init(const char *log_path, SF_LOG_LEVEL log_level) {
     SF_LOG_LEVEL sf_log_level = log_level;
     char strerror_buf[SF_ERROR_BUFSIZE];
 
+    char client_config_file[MAX_PATH] = {0};
+    snowflake_global_get_attribute(
+      SF_GLOBAL_CLIENT_CONFIG_FILE, client_config_file, sizeof(client_config_file));
+    client_config clientConfig = { .logLevel = "", .logPath = "" };
```
sfc-gh-ext-simba-hx, line 315: memory leak currently with the buffer allocated in load_client_config().
When you try to free the buffer, initialize with const string could be a problem.

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,315 @@
...
+namespace
+{
+  // constants
+  const std::string SF_CLIENT_CONFIG_FILE_NAME("sf_client_config.json");
+  const std::string SF_CLIENT_CONFIG_ENV_NAME("SF_CLIENT_CONFIG_FILE");
+  std::set<std::string> KnownCommonEntries = {"log_level", "log_path"};
+}
+
+////////////////////////////////////////////////////////////////////////////////
+void load_client_config(
```
sfc-gh-ext-simba-hx, line 33: When we wrapping c++ into c interface better to catch any possible exception since c interface doesn't expect exception.

lib/client.c
```diff
@@ -36,6 +37,8 @@ sf_bool DEBUG;
 sf_bool SF_OCSP_CHECK;
 char *SF_HEADER_USER_AGENT = NULL;
 
+static char *CLIENT_CONFIG_FILE = NULL;
+static char* LOG_LEVEL;
```
sfc-gh-ext-simba-hx, line 41: why not to set NULL as others?

lib/client.c
```diff
@@ -620,6 +644,15 @@ snowflake_global_set_attribute(SF_GLOBAL_ATTRIBUTE type, const void *value) {
         case SF_GLOBAL_OCSP_CHECK:
             SF_OCSP_CHECK = *(sf_bool *) value;
             break;
+        case SF_GLOBAL_CLIENT_CONFIG_FILE:
+            alloc_buffer_and_copy(&CLIENT_CONFIG_FILE, value);
+            break;
+        case SF_GLOBAL_LOG_LEVEL:
```
sfc-gh-ext-simba-hx, line 650: I would prefer SF_GLOBAL_LOG_PATH and SF_GLOBAL_LOG_LEVEL to be read only attributes since application already have enough ways to do log settings.

include/snowflake/client.h
```diff
@@ -277,7 +277,10 @@ typedef enum SF_GLOBAL_ATTRIBUTE {
     SF_GLOBAL_CA_BUNDLE_FILE,
     SF_GLOBAL_SSL_VERSION,
     SF_GLOBAL_DEBUG,
-    SF_GLOBAL_OCSP_CHECK
+    SF_GLOBAL_OCSP_CHECK,
+    SF_GLOBAL_CLIENT_CONFIG_FILE,
+    SF_GLOBAL_LOG_LEVEL,
```
sfc-gh-ext-simba-hx, line 282: remind me why we need SF_GLOBAL_LOG_LEVEL and SF_GLOBAL_LOG_PATH?

cpp/lib/ClientConfigParser.cpp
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Since the whole function is just to resolve the client config, I think it might make more sense to just keep it all there (I'll move all the resolving code to a separate function from loadClientConfig instead)

include/snowflake/client.h
sfc-gh-ext-simba-nl: Updated description

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,315 @@
...
+////////////////////////////////////////////////////////////////////////////////
+void ClientConfigParser::parseConfigFile(
+  const std::string& in_filePath,
+  ClientConfig& out_clientConfig)
+{
+  cJSON* jsonConfig;
+  FILE* configFile;
+  try 
+  {
+    configFile = sf_fopen(&configFile, in_filePath.c_str(), "r");
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Unfortunately std::filesystem isn't in osx10.14 yet so it can't be used and there seems to be a 32-bit windows debug issue with reading file with fstream

cpp/lib/ClientConfigParser.cpp
sfc-gh-ext-simba-nl: std::filesystem isn't available on osx10.14 and boost has a debug assertion on 32-bit windows with MT build

cpp/lib/ClientConfigParser.cpp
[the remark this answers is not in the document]
sfc-gh-jszczerbinski: That means we cannot use std::fstream at all? Can you show me the assertion?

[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Not exactly, it only happens for Windows 32-bit built with MT. All other configurations seem to work fine. If you prefer, I can ifdef it to not run when it's windows debug and it should work fine. The assertion says something like this: 
Debug Assertion Failed!

minkernel\crts\ucrt\src\appcrt\lowio\read.cpp

Expression: _osfile(fh) & FOPEN

include/snowflake/ClientConfigParser.hpp
```diff
@@ -0,0 +1,105 @@
...
+#define SNOWFLAKE_CONFIGPARSER_HPP
+
+#include <string>
+#include "client_config_parser.h"
+
+namespace Snowflake
+{
+namespace Client
+{
+  struct ClientConfigException : public std::exception
```
sfc-gh-jszczerbinski, line 15: Why do we need a class if it has no member variables?

sfc-gh-ext-simba-nl: Just using it as a wrapper for exception to pass along the error message. I saw other classes having a similar exception struct, but I can just make the function return false instead of throwing an exception if you prefer.

cpp/lib/ClientConfigParser.cpp
sfc-gh-ext-simba-nl: I think using boost filesystem was able to fix the issue

cpp/lib/ClientConfigParser.cpp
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: I'm still running into this issue with the tests run on jenkins. Basically the test run just hangs after it hits the assertion error. In the mean time, I've added an ifdef to to run easy logging when in 32-bit debug. I've locally tested running on /MDd and it runs fine, but currently I think we build with dynamic runtime set to OFF

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,292 @@
...
+  const std::string& in_configFilePath)
+{
+  char envbuf[MAX_PATH + 1];
+  if (!in_configFilePath.empty())
+  {
+    // 1. Try config file if it was passed in
+    CXX_LOG_INFO("Using client configuration path from a connection string: %s", in_configFilePath.c_str());
+    return in_configFilePath;
+  }
+  else if (const char* clientConfigEnv = sf_getenv_s(SF_CLIENT_CONFIG_ENV_NAME.c_str(), envbuf, sizeof(envbuf)))
```
sfc-gh-jszczerbinski, line 102: Else no longer needed when we do an early return.

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,292 @@
...
+  {
+    // 2. Try environment variable SF_CLIENT_CONFIG_ENV_NAME
+    CXX_LOG_INFO("Using client configuration path from an environment variable: %s", clientConfigEnv);
+    return clientConfigEnv;
+  }
+  else
```
sfc-gh-jszczerbinski, line 108: Else no longer needed when we do an early return.

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,292 @@
...
+    err = parse(jsonConfig, configFile);
+
+    if (!err.empty())
+    {
+      CXX_LOG_ERROR("Error in parsing JSON: %s, err: %s", in_filePath.c_str(), err.c_str());
+      std::string errMsg = "Error parsing client configuration file: " + in_filePath;
+      throw ClientConfigException(errMsg.c_str());
+    }
+  }
+  catch (std::exception& e)
```
sfc-gh-jszczerbinski, line 196: We don't need to explicitly close the file. It will close automatically when destructor of ifstream is called:
- https://en.cppreference.com/w/cpp/language/raii
- https://en.cppreference.com/w/cpp/io/basic_ifstream (destructor doc)

sfc-gh-jszczerbinski (changes requested):

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,288 @@
...
+  std::ifstream configFile;
+  try
+  {
+    configFile.open(in_filePath, std::fstream::in | std::ios::binary);
+    if (!configFile)
+    {
+      CXX_LOG_INFO("Could not open a file. The file may not exist: %s",
+        in_filePath.c_str());
+      std::string errMsg = "Error finding client configuration file: " + in_filePath;
+      throw ClientConfigException(errMsg.c_str());
```
sfc-gh-jszczerbinski, line 180: Please avoid using exceptions (https://google.github.io/styleguide/cppguide.html#Exceptions)

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,288 @@
...
+#if defined(WIN32) || defined(_WIN64)
+  std::string homeDirFilePath;
+  char* homeDir;
+  if ((homeDir = sf_getenv_s("USERPROFILE", envbuf, sizeof(envbuf))) && strlen(homeDir) != 0)
+  {
+    homeDirFilePath = homeDir + PATH_SEP + SF_CLIENT_CONFIG_FILE_NAME;
+  } else {
+    // USERPROFILE is empty, try HOMEDRIVE and HOMEPATH
+    char* homeDriveEnv = sf_getenv_s("HOMEDRIVE", envbuf, sizeof(envbuf));
+    char* homePathEnv = sf_getenv_s("HOMEPATH", envbuf, sizeof(envbuf));
```
sfc-gh-jszczerbinski, line 138: I think we overwrite homePathEnv here. `envbuf` is used twice.

sfc-gh-jszczerbinski: Please add tests for `resolveHomeDirConfigPath`

sfc-gh-jszczerbinski (changes requested):

sfc-gh-jszczerbinski (changes requested):
In general looks good to me. We need to log and return false if contents of of the config are malformed;

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,281 @@
...
+    }
+    if (commonProps.contains("log_path") && commonProps.get("log_path").is<std::string>())
+    {
+      const char* logPath = commonProps.get("log_path").get<std::string>().c_str();
+      size_t logPathSize = strlen(logPath) + 1;
+      out_clientConfig.logPath = (char*)SF_CALLOC(1, logPathSize);
+      sf_strcpy(out_clientConfig.logPath, logPathSize, logPath);
+    }
+  }
+  return true;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: The reason it still returns true is because if the user passes the log level but not log path to log_init() and the config file only has log path, we would still like the logger to log to the log path set in the config (and vice versa).

sfc-gh-jszczerbinski (commented):
In general looks good to me.

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,281 @@
...
+#endif
+  err = parse(jsonConfig, configFile);
+
+  if (!err.empty())
+  {
+    CXX_LOG_ERROR("Error in parsing JSON: %s, err: %s", in_filePath.c_str(), err.c_str());
+    return false;
+  }
+
+  value commonProps = jsonConfig.get("common");
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Added checks before and after

tests/test_unit_logger.c
```diff
@@ -1,149 +1,317 @@
...
+#else
+  assert_true(result);
+#endif
+}
+
+/**
+ * Tests log settings from client config file with invalid json
+ */
+void test_client_config_log_invalid_json(void** unused) {
+  char clientConfigJSON[] = "{{{\"invalid json\"}";
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Added several new test cases for this

tests/test_unit_logger.c
```diff
@@ -1,149 +1,431 @@
...
-        cmocka_unit_test(test_log_str_to_level),
-#ifndef _WIN32
-        cmocka_unit_test(test_log_creation),
-        cmocka_unit_test(test_mask_secret_log),
-#endif
-    };
-    return cmocka_run_group_tests(tests, NULL, NULL);
-}
+/*
+ * Copyright (c) 2018-2019 Snowflake Computing, Inc. All rights reserved.
```
sfc-gh-dprzybysz, line 2: 2024 or 2025 if it will be merged next year ;)

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,286 @@
...
+  std::set<std::string> KnownCommonEntries = {"log_level", "log_path"};
+}
+
+////////////////////////////////////////////////////////////////////////////////
+sf_bool load_client_config(
+  const char* in_configFilePath,
+  client_config* out_clientConfig)
+{
+// Disable easy logging for 32-bit windows debug build due to linking issues
+// with _osfile causing hanging/assertions until dynamic linking is available
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Correct, but only with 32-bit windows debug build, all other configurations/platforms/bitness work fine

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  // 2. Try environment variable SF_CLIENT_CONFIG_ENV_NAME
+  if (const char* clientConfigEnv = sf_getenv_s(SF_CLIENT_CONFIG_ENV_NAME.c_str(), envbuf, sizeof(envbuf)))
+  {
+    CXX_LOG_INFO("Using client configuration path from an environment variable: %s", clientConfigEnv);
+    return clientConfigEnv;
+  }
+
+  // 3. Try DLL binary dir
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: This was just ported from ODBC. Is this something we should change for libsnowflakeclient?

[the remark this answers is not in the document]
sfc-gh-dprzybysz: we can keep it if it will also work with static library - maybe then change the comment to say that the file will be read from the library dir?

cpp/lib/ClientConfigParser.cpp
```diff
@@ -0,0 +1,286 @@
...
+  }
+
+  // 2. Try environment variable SF_CLIENT_CONFIG_ENV_NAME
+  if (const char* clientConfigEnv = sf_getenv_s(SF_CLIENT_CONFIG_ENV_NAME.c_str(), envbuf, sizeof(envbuf)))
+  {
+    CXX_LOG_INFO("Using client configuration path from an environment variable: %s", clientConfigEnv);
+    return clientConfigEnv;
+  }
+
+  // 3. Try DLL library dir
```
sfc-gh-dprzybysz, line 108: Let's remove DLL (library seems to be enough) and rename functions and variables below

sfc-gh-dprzybysz (approved):

lib/client.c
```diff
@@ -315,18 +317,38 @@ static sf_bool STDCALL log_init(const char *log_path, SF_LOG_LEVEL log_level) {
     SF_LOG_LEVEL sf_log_level = log_level;
     char strerror_buf[SF_ERROR_BUFSIZE];
 
+    char client_config_file[MAX_PATH] = { 0 };
+    snowflake_global_get_attribute(
+      SF_GLOBAL_CLIENT_CONFIG_FILE, client_config_file, sizeof(client_config_file));
+    client_config clientConfig = { 0 };
```
sfc-gh-mmishchenko, line 323: Won't it be easier if we use preallocated buffers? taking into account we've already allocated them by this point? It would be then:
`client_config clientConfig = { log_level_buf, log_path_buf };`
Or alternatively we may adjust client_config definition (even better imho):
```
typedef struct client_config
  {
    // The log level
    char logLevel[64];

    // The log path
    char logPath[MAX_PATH];
  } client_config;
```
This way we'll use same buffers both for env-source attempt and config and get rid of heap allocations.

tests/test_unit_logger.c
sfc-gh-mmishchenko, line 1: Didn't look thoroughly into this relying on the test coverage report.

sfc-gh-mmishchenko (changes requested):

sfc-gh-mmishchenko (changes requested):
In essence I would suggest first:
- [ ] Check if we really require to allocate dynamically buffers for client_config
- [ ] Check if ClientConfigParser.hpp can be removed and ClientConfigParser.cpp be renamed to client_conifg_parser.cpp

lib/logger.c
```diff
@@ -216,6 +219,10 @@ SF_LOG_LEVEL log_from_str_to_level(const char *level_in_str) {
     return SF_LOG_FATAL;
 }
 
+const char* log_from_level_to_str(SF_LOG_LEVEL level) {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: It would return SF_LOG_FATAL, which is what we decided in the main thread of this issue would be the default

include/snowflake/logger.h
```diff
@@ -47,7 +47,8 @@ typedef enum SF_LOG_LEVEL {
     SF_LOG_INFO,
     SF_LOG_WARN,
     SF_LOG_ERROR,
-    SF_LOG_FATAL
+    SF_LOG_FATAL,
+    SF_LOG_DEFAULT
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Correct, it was discussed in [#773](https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/773)

cpp/lib/client_config_parser.cpp
```diff
@@ -0,0 +1,260 @@
...
+    // 1. Try config file if it was passed in
+    if (!configFilePath.empty())
+    {
+      CXX_LOG_INFO("Using client configuration path from a connection string: %s", configFilePath.c_str());
+      return configFilePath;
+    }
+
+    // 2. Try environment variable SF_CLIENT_CONFIG_ENV_NAME
+    char envbuf[MAX_PATH + 1];
+    if (const char* clientConfigEnv = sf_getenv_s(SF_CLIENT_CONFIG_ENV_NAME.c_str(), envbuf, sizeof(envbuf)))
```
sfc-gh-mkubik, line 128: could the `getEnvironmentVariableValue` function be used here?

lib/client.c
```diff
@@ -315,18 +317,36 @@ static sf_bool STDCALL log_init(const char *log_path, SF_LOG_LEVEL log_level) {
...
+       Specified parameters takes precedence over client config */
     sf_log_path = sf_getenv_s("SNOWFLAKE_LOG_PATH", log_path_buf, sizeof(log_path_buf));
-    if (sf_log_path == NULL && log_path) {
+    if (sf_log_path == NULL) {
+      if (log_path && strlen(log_path) != 0) {
         sf_log_path = log_path;
+      } else if (strlen(clientConfig.logPath) != 0) {
+        sf_log_path = clientConfig.logPath;
+      }
     }
```
sfc-gh-mmishchenko, line 338: Shall we extract this into a separate function and omit if log_path/log_level were provided by the caller?

sfc-gh-mmishchenko (changes requested):

sfc-gh-mmishchenko (changes requested):
- [ ] Check where `boost::filesystem::path` capability to concatenate paths can be utilised instead of `std::string` concatenation.
- [ ] The waterfall logic `function params -> env params -> config params` is a bit tough to follow from the current code, maybe there is some room for refactoring to make it more comprehensible

lib/client.c
```diff
@@ -315,18 +317,38 @@ static sf_bool STDCALL log_init(const char *log_path, SF_LOG_LEVEL log_level) {
...
+      load_client_config(client_config_file, &clientConfig);
+    }
+
     size_t log_path_size = 1; //Start with 1 to include null terminator
     log_path_size += strlen(time_str);
 
-    /* The environment variables takes precedence over the specified parameters */
+    /* The environment variables takes precedence over the specified parameters.
+       Specified parameters takes precedence over client config */
     sf_log_path = sf_getenv_s("SNOWFLAKE_LOG_PATH", log_path_buf, sizeof(log_path_buf));
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: With respect to precedence of parameters vs environment variable, if we change it to params > env (which I also think makes more sense), but then it'll be a behavioral change from what it used to do. Is that something we want to go through with?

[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Although, now that I think about it. I wonder if it was designed that way because log level is an enum which previously did not have SF_LOG_DEFAULT so the log level would always take precedence over everything else. So it might be ok to change to params > env, but would essentially break any backwards compatibility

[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Yes, correct.

sfc-gh-mmishchenko (changes requested):

sfc-gh-mmishchenko (approved):
LGTM

include/snowflake/client_config_parser.h
```diff
@@ -0,0 +1,53 @@
...
+  sf_bool load_client_config(
+    const char* configFilePath,
+    client_config* clientConfig);
+
+  /**
+    * Free client config memory
+    *
+    * @param clientConfig            The client_config object to be freed.
+    */
+  void free_client_config(client_config* clientConfig);
```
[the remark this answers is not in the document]
sfc-gh-jszczerbinski: Is it needed? It doesn't seem to be since we do not allocate anything on the heap

cpp/lib/client_config_parser.cpp
```diff
@@ -0,0 +1,278 @@
...
+          }
+        }
+        return true;
+      }
+    }
+    CXX_LOG_ERROR("Malformed client config file: %s", filePath.c_str());
+    return false;
+  }
+
+  ////////////////////////////////////////////////////////////////////////////////
```
sfc-gh-jszczerbinski, line 251: Nit: Remove lines like this

lib/platform.c
```diff
@@ -297,7 +297,7 @@ int STDCALL sf_mkdir(const char *path) {
 #ifdef _WIN32
     return _mkdir(path);
 #else
-    return mkdir(path, 0755);
+    return mkdir(path, 0700);
```
sfc-gh-jszczerbinski, line 300: Why is this change needed?

cpp/lib/client_config_parser.cpp
```diff
@@ -0,0 +1,278 @@
...
+#include <set>
+
+#undef snprintf
+#include <boost/filesystem.hpp>
+
+#ifndef _WIN32 
+#include <dlfcn.h>
+#endif
+
+using namespace Snowflake::Client;
```
sfc-gh-jszczerbinski, line 23: Move using namespace inside the anonymous namespace to limit the scope its effects

sfc-gh-jszczerbinski (approved):
Looks good to me. Please correct the minor comments I left.

include/snowflake/client_config_parser.h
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: No, it's no longer needed. I forgot to remove it from the header file. Thanks for catching it

lib/platform.c
sfc-gh-ext-simba-nl: When I was going through the security risk mitigation, it mentioned that the folder should only have 700 permissions

sfc-gh-jszczerbinski: We are not using the sf_mkdir in this PR

PR #750, opened 2024-10-08
SNOW-715551: Support authentication with OKTA

sfc-gh-ext-simba-jy:
- [x] Implemented the okta authencation
- [x] Added the disable saml URL check option


cpp/lib/Authenticator.cpp
```diff
@@ -112,9 +121,11 @@ extern "C" {
     }
     catch (...)
     {
-      SET_SNOWFLAKE_ERROR(&conn->error, SF_STATUS_ERROR_GENERAL,
-                          "authentication failed",
-                          SF_SQLSTATE_GENERAL_ERROR);
+        if (getAuthenticatorType(conn->authenticator) == AUTH_JWT) {
+            SET_SNOWFLAKE_ERROR(&conn->error, SF_STATUS_ERROR_GENERAL,
+                "authentication failed",
```
sfc-gh-ext-simba-hx, line 126: Better to append the error message from exception.

cpp/lib/Authenticator.cpp
```diff
@@ -337,12 +353,200 @@ namespace Client
...
+              retried_count, elapsed_time);
+          SET_SNOWFLAKE_ERROR(err, SF_STATUS_ERROR_REQUEST_TIMEOUT, "OktaConnectionFailed: timeout reached", SF_SQLSTATE_GENERAL_ERROR);
+          AUTH_THROW(err);
+      }
+
+      // 4. get SAML response
+      SFURL sso_url = SFURL::parse(sso_url_str);
+      sso_url.addQueryParam("onetimetoken", one_time_token);
+      resp = NULL;
+      if (!curl_get_call(m_connection, curl, (char*)sso_url.toString().c_str(), NULL, &resp, &m_connection->error))
```
sfc-gh-ext-simba-hx, line 456: We need renew one time token for each retry on get request. Check ODBC PR for details. https://github.com/snowflakedb/snowflake-odbc/pull/509

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: Updated the PR. Please take a look.

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: I applied it to the implementation.

CMakeLists.txt
```diff
@@ -155,6 +155,8 @@ set (SOURCE_FILES_PUT_GET
         cpp/StorageClientFactory.hpp
         cpp/StorageClientFactory.cpp
         cpp/RemoteStorageRequestOutcome.hpp
+        cpp/entities.cpp
+        cpp/entities.hpp
```
sfc-gh-ext-simba-hx, line 159: Since this is copied from ODBC I think better to make ODBC to reuse this later. Can we move entities.hpp to include/snowflake or move the function declaration into an existing header file there?

lib/http_perform.c
```diff
@@ -454,8 +454,8 @@ sf_bool STDCALL http_perform(CURL *curl,
 
         // When renew timeout is reached, stop retry and return to the caller
         // to renew request
-        if ((retry) && (renew_timeout > 0) &&
-            ((time(NULL) - elapsedRetryTime) >= renew_timeout)) {
+        if (((retry) && (renew_timeout > 0) &&
+            ((time(NULL) - elapsedRetryTime) >= renew_timeout)) || renew_timeout < 0) {
```
sfc-gh-ext-simba-hx, line 458: This is somehow misleading, please put `renew_timeout < 0` in a new line and put others in the same line.

lib/connection.c
```diff
@@ -470,7 +470,12 @@ sf_bool STDCALL curl_get_call(SF_CONNECT *sf,
                               char *url,
                               SF_HEADER *header,
                               cJSON **json,
-                              SF_ERROR_STRUCT *error) {
+                              SF_ERROR_STRUCT *error,
+                              int64 renew_timeout,
```
sfc-gh-jszczerbinski, line 474: Can we group those params into a struct?

cpp/lib/Authenticator.hpp
```diff
@@ -87,6 +88,28 @@ namespace Client
...
+  class AuthenticatorOKTA : public IAuthenticator
+  {
+  public:
+      AuthenticatorOKTA(SF_CONNECT* conn);
+
+      ~AuthenticatorOKTA();
+
+      void authenticate();
+
+      void updateDataMap(cJSON* dataMap);
```
sfc-gh-jszczerbinski, line 100: This is a copy of the code in ODBC, except it uses different json library. I don't like the duplication that's happening here Do we have a plan on how to share this functionality between two codebases? @sfc-gh-ext-simba-hx

sfc-gh-jszczerbinski (changes requested):

lib/http_perform.c
sfc-gh-ext-simba-jy: I applied it @sfc-gh-jszczerbinski. Thank you for your suggestion.

cpp/lib/Authenticator.hpp
sfc-gh-ext-simba-jy: Hello @sfc-gh-jszczerbinski, he is OOO today. As far as I know, we have implemented all the authentications in the libsnowflakeclient and will merge them into the master 2.0.0 libsnowflakeclient first. If we start using the libsnowflakeclient master 2.0, we will refactor the code in ODBC based on the changes in the libsnowflakeclient.

sfc-gh-jszczerbinski (commented):
In general looks good too me. But I would like to wait before merge to decide on a solution for the issues outlined here: https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/1045#issuecomment-2418987172

cpp/lib/Authenticator.hpp
sfc-gh-ext-simba-hx: @sfc-gh-jszczerbinski Sorry I missed this. Currently libsfclient and odbc use different implementation of rest request and json library and so far no plan to change that. Since authenticator implementation deeply rely on those parts I think we would need some refactoring there to make the authenticator implementation reusable.
It's doable but I think it might be too risky to do large refactoring on the very essential functionalities like rest request and json parser, especially no customer needs on that.

sfc-gh-ext-simba-hx: @sfc-gh-jszczerbinski Replied your comment in the ticket now I think it's a good idea to use picojson in C++ implementation.
For the rest request part, maybe we could have a virtual function for that to allow overriding on ODBC/libsfclient side, like what I'm doing for bulk updating? https://github.com/snowflakedb/libsnowflakeclient/compare/master...SNOW-710476-bulk-update-essential#diff-1e3b9047a718b31a2af5a7fa63d36f3f4333b7dfa3a0ddde58892d15e015c458R68
@sfc-gh-ext-simba-jy

cpp/lib/Authenticator.cpp
```diff
@@ -73,6 +80,11 @@ extern "C" {
         conn->auth_object = static_cast<Snowflake::Client::IAuthenticator*>(
                               new Snowflake::Client::AuthenticatorJWT(conn));
       }
+      if (AUTH_OKTA == auth_type)
```
sfc-gh-ext-simba-hx, line 83: switch or else if?

cpp/lib/Authenticator.cpp
```diff
@@ -116,29 +128,27 @@ extern "C" {
...
-        if (!data)
-        {
-            data = snowflake_cJSON_CreateObject();
-            snowflake_cJSON_AddItemToObject(body, "data", data);
-        }
+          data = snowflake_cJSON_CreateObject();
+          snowflake_cJSON_AddItemToObject(body, "data", data);
+    }
+
+    snowflake_cJSON_DeleteItemFromObject(data, "AUTHENTICATOR");
```
sfc-gh-ext-simba-hx, line 146: auth_update_json_body is a C wrapper for updateDataMap in C++. AUTH_OAUTH is an exception since it's quite simple so we don't have authenticator implementation for it in C++. For others we should fully rely on updateDataMap in case any common logic here might not apply.

cpp/lib/Authenticator.cpp
```diff
@@ -194,6 +212,10 @@ extern "C" {
       {
```
sfc-gh-ext-simba-hx, line 212: I think better to use
if (conn->auth_object)
{
    delete static_cast<Snowflake::Client::IAuthenticator*>(conn->auth_object);
}

sfc-gh-ext-simba-jy: Updated.

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: I moved this to connection.c

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: Updated.

cpp/lib/Authenticator.cpp
```diff
@@ -356,12 +356,197 @@ namespace Client
...
+          SET_SNOWFLAKE_ERROR(err, SF_STATUS_ERROR_REQUEST_TIMEOUT, "OktaConnectionFailed: timeout reached", SF_SQLSTATE_GENERAL_ERROR);
+          goto cleanup;
+      }
+
+      rawData = buf.buffer;
+  cleanup:
+      m_retryTimeout -= elapsedTime;
+      sf_header_destroy(httpExtraHeaders);
+      free_curl_desc(curl_desc);
+      SF_FREE(raw_resp);
```
sfc-gh-jszczerbinski, line 510: We allocate with new and deallocate with SF_FREE

cpp/lib/Authenticator.cpp
```diff
@@ -356,12 +356,197 @@ namespace Client
...
+
+          SET_SNOWFLAKE_ERROR(err, SF_STATUS_ERROR_REQUEST_TIMEOUT, "OktaConnectionFailed: timeout reached", SF_SQLSTATE_GENERAL_ERROR);
+          goto cleanup;
+      }
+
+      rawData = buf.buffer;
+  cleanup:
+      m_retryTimeout -= elapsedTime;
+      sf_header_destroy(httpExtraHeaders);
+      free_curl_desc(curl_desc);
```
sfc-gh-jszczerbinski, line 509: Why do we free curl_desc here?

cpp/lib/Authenticator.cpp
```diff
@@ -356,12 +356,197 @@ namespace Client
...
+      }
+
+      if (elapsedTime >= m_retryTimeout)
+      {
+          CXX_LOG_WARN("sf", "AuthenticatorOKTA", "get_curl_call",
+              "Fail to get SAML response, timeout reached: %d, elapsed time: %d",
+              m_retryTimeout, elapsedTime);
+
+          SET_SNOWFLAKE_ERROR(err, SF_STATUS_ERROR_REQUEST_TIMEOUT, "OktaConnectionFailed: timeout reached", SF_SQLSTATE_GENERAL_ERROR);
+          goto cleanup;
```
sfc-gh-jszczerbinski, line 502: Please don't use goto in C++ code. Use RAII and if you need to manage resource from C API use unique_ptr (https://en.cppreference.com/w/cpp/memory/unique_ptr/unique_ptr constructors (3) and (4))

cpp/lib/IAuth.cpp
```diff
@@ -0,0 +1,167 @@
...
+            getIDPInfo();
+
+            // 2. verify ssoUrl and tokenUrl contains same prefix
+            if (!SFURL::urlHasSamePrefix(tokenURLStr, m_authenticator))
+            {
+                CXX_LOG_ERROR("sf", "AuthenticatorOKTA", "authenticate",
+                    "The specified authenticator is not supported, "
+                    "authenticator=%s, token url=%s, sso url=%s",
+                    m_authenticator.c_str(), tokenURLStr.c_str(), ssoURLStr.c_str());
+                AUTH_THROW("SFAuthenticatorVerificationFailed: the token URL does not have the same prefix with the authenticator");
```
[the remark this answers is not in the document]
sfc-gh-jszczerbinski: Also please avoid using exceptions. Based on our internal discussions we should try to follow google c++ style guide: https://google.github.io/styleguide/cppguide.html#Exceptions

sfc-gh-jszczerbinski (changes requested):
This PR is moving in the right direction. We need to improve some things before the merge:
- Please rethink the inheritance structure. I don't think we need a complex one for this change
- Please avoid using exceptions and goto, it complicates the flow and makes it hard to implement reason about resource lifetimes.
- Consider moving the code which handles SAML auth to separate file. Currently it dominates files that are containing implementations of other auth methods.
- In C++ we use RAII to managed lifetimes of resources. When using resource from C API  consider using unique_ptr with custom deleter (https://stackoverflow.com/questions/19053351/how-do-i-use-a-custom-deleter-with-a-stdunique-ptr-member)

include/snowflake/IAuth.hpp
```diff
@@ -0,0 +1,125 @@
...
+        std::string m_account;
+        std::string m_appID;
+        std::string m_appVersion;
+        std::string m_user;
+        std::string m_port;
+        std::string m_host;
+        std::string m_protocol;
+    };
+
+    class IAuthenticatorOKTA : public IIDPAuthenticator, public IAuthenticator
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: IIDPAuthenticator is for the HTTP call to get IDP info for Okta and ExternalBrowser in the future. IAuthenticator is the original Interface class for all authenticators in the libsnowflakeclient.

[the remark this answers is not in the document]
sfc-gh-jszczerbinski: It makes sense, but also makes the code hard to read. Generally in C++ multiple inheritance is not advised and composition is preferred over implementation inheritance (https://google.github.io/styleguide/cppguide.html#Inheritance). Can we make IDPAuthenticator (perhaps  could be a better name) a member of AuthenticatorOKTA instead?

cpp/lib/IAuth.cpp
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Actually, this error throwing is to escape from the authentication block and go to auth_update_json_body or auth_renew_json_body. We have used this in JWT authentication as well.

include/snowflake/IAuth.hpp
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I will try to redesign this based on your feedback.

include/snowflake/SFURL.hpp
```diff
@@ -422,6 +422,15 @@ class SFURL
...
   }
 
+  /**
+   * Verify that if two urls has same prefix (protocl + host + port)
+   * @param url1
+   * @param url2
+   * 
+   * @return true if same prefix otherwise false
+   */
+  static bool urlHasSamePrefix(std::string url1, std::string url2);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: This function is from ODBC and I try to use the same function without any changes. Actually, this is not a good place for this function. I will try to move this function elsewhere.

cpp/lib/Authenticator.cpp
```diff
@@ -356,12 +356,197 @@ namespace Client
...
+      if (!::curl_post_call(m_connection, curl, (char*)destination.c_str(), httpExtraHeaders, (char*)s_body.c_str(),
+          &resp_data, err, renewTimeout, maxRetryCount, m_retryTimeout, &elapsedTime,
+          &m_retriedCount, NULL, SF_BOOLEAN_TRUE))
+      {
+          log_info("sf", "IDPAuthenticator", "post_curl_call",
+              "Fail to get authenticator info, response body=%s\n",
+              snowflake_cJSON_Print(snowflake_cJSON_GetObjectItem(resp_data, "data")));
+          SET_SNOWFLAKE_ERROR(err, SF_STATUS_ERROR_GENERAL, "SFConnectionFailed", SF_SQLSTATE_GENERAL_ERROR);
+          AUTH_THROW(err);
+          goto cleanup;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I will fix it.

cpp/lib/Authenticator.cpp
```diff
@@ -356,12 +356,197 @@ namespace Client
...
 
     return coded;
   }
+
+  void AuthenticatorOKTA::curl_post_call(SFURL& url, const jsonObject_t& obj, jsonObject_t& resp)
+  {
+      std::string destination = url.toString();
+      void* curl_desc;
+      CURL* curl;
+      curl_desc = get_curl_desc_from_pool(destination.c_str(), m_connection->proxy, m_connection->no_proxy);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Fixed.

sfc-gh-ext-simba-hx:
> * Please avoid using exceptions (https://google.github.io/styleguide/cppguide.html#Exceptions) and goto, it complicates the flow and makes it hard to reason about resource lifetimes.

The c++ implementation in libsnowflakeclient are mainly for reusing in ODBC, which is using exception for error handling. That's from the SimbaEngineSDK framework and unlikely that could be changed. Keep using exception would make it easier to reuse it in ODBC, especially when the implementation is originally from ODBC.
Or maybe we could consider not to reuse it in ODBC? That would make the logic (including the inheritance structure) a lot simpler.

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: I changed it to delete.

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: Sorry, it is unclear to me. I use free_curl_desc here because this is the end of the function. Don't you think that this is reasonable?

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: I use the exception instead of goto. In my opinion, using unique_ptr is inappropriate for this because I should move the pointer to call the function and get the pointer back from the HTTP perform.

include/snowflake/SFURL.hpp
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I moved this function to SNOWFLAKE_UTIL

sfc-gh-ext-simba-jy:
As we decide to follow Google c++ style, I will create a new PR instead of refactoring the code on the current PR. I will close this PR after I create a new one.

sfc-gh-ext-simba-jy:
Closed this PR. I made a new one. https://github.com/snowflakedb/libsnowflakeclient/pull/799

sfc-gh-dprzybysz:
Closing, @sfc-gh-ext-simba-jy please delete the branch if it is not necessary anymore

PR #781, opened 2024-11-25
SNOW-1825584: Cache deps to speed up the builds

sfc-gh-jszczerbinski:
In this PR:
- We cache deps built on master
- We use cached deps on pull requests to speed up build times

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -23,7 +23,7 @@ jobs:
             - uses: actions/checkout@v1
             - name: Restore cached deps
               id: cache-restore-deps
-              uses: actions/cache/restore@v3
+              uses: actions/cache/restore@v4
               with:
                 path: dep-cache
                 key: ${{ matrix.build_type }}-${{ github.event.pull_request.base.sha }}-Linux-dep-cache
@@ -35,7 +35,7 @@ jobs:
               run: ci/build_linux.sh
             - name: Cache deps
               id: cache-save-deps
-              uses: actions/cache/save@v3
+              uses: actions/cache/save@v4
               with:
                 path: dep-cache
                 key: ${{ matrix.build_type }}-${{ github.sha }}-Linux-dep-cache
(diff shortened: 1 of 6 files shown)


.github/workflows/build-test.yml
```diff
@@ -21,11 +21,25 @@ jobs:
                 cloud_provider: [ 'AWS', 'AZURE', 'GCP' ]
         steps:
             - uses: actions/checkout@v1
+            - name: Restore cached deps
+              id: cache-restore-deps
+              uses: actions/cache/restore@v3
```
sfc-gh-dprzybysz, line 26: there is version 4 available (also for save action) https://github.com/actions/cache

.github/workflows/build-test.yml
```diff
@@ -81,13 +110,27 @@ jobs:
...
+              with:
+                path: dep-cache
+                key: ${{ matrix.build_type }}-${{ matrix.platform }}-${{ matrix.vs_version }}-${{ github.event.pull_request.base.sha }}-Win-dep-cache
+              if: github.event_name == 'pull_request'
```
sfc-gh-dprzybysz, line 119: should we always restore or should we check explicitly if the entry is available?

.github/workflows/build-test.yml
```diff
@@ -114,11 +157,25 @@ jobs:
...
               env:
                 BUILD_TYPE: ${{ matrix.build_type }}
               run: ./ci/build_mac.sh
+            - name: Cache deps
+              id: cache-save-deps
+              uses: actions/cache/save@v3
+              with:
+                path: dep-cache
+                key: ${{ matrix.build_type }}-${{ github.sha }}-Mac-dep-cache
+              if: github.ref_name == github.event.repository.default_branch && matrix.cloud_provider == 'AWS'
```
sfc-gh-dprzybysz, line 178: we save to different place that we restore from - is it correct?

sfc-gh-jszczerbinski: We build and cache deps on master (commit sha = github.sha, but we restore on pull requests on master. On pull requests we will restore deps based on a base_ref - some commit on master (commit sha = github.event.pull_request.base.sha). So we restore and save from the same cache -> just not on the same workflow.

.github/workflows/build-test.yml
sfc-gh-jszczerbinski: I had issues with version 4 while testing. Maybe we can upgrade later on.

sfc-gh-dprzybysz (approved):

sfc-gh-mkubik (approved):

PR #782, opened 2024-11-26
SNOW-1524259: Reduce dependencies size in libsnowflakeclient

sfc-gh-ext-simba-jl:
Draft PR for reducing dependencies size in libsnowflakeclient


exports.def
```diff
@@ -0,0 +1,84 @@
...
+    snowflake_timestamp_get_year                   @68
+    snowflake_timestamp_get_tzoffset               @69
+    snowflake_timestamp_get_scale                  @70
+    sf_setenv                                      @71
+    sf_unsetenv                                    @72
+    sf_getenv_s                                    @73
+    sf_get_tmp_dir                                 @74
+    sf_calloc                                      @75
+    sf_free                                        @76
+    uuid4_generate                                 @77
```
sfc-gh-dprzybysz, line 84: we will need to update the exports whenever the public api is changed or we can make it dynamically created?

tests/CMakeLists.txt
```diff
@@ -310,6 +258,71 @@ else()
...
+
+    FOREACH (T ${TESTS_C_UNIT})
+        add_executable(${T} ${SOURCE_UTILS} ${T}.c)
+        if (WIN32)
+            target_include_directories(
+                    ${T} PUBLIC
+                    ../deps-build/${PLATFORM}/${VSDIR}/${CMAKE_BUILD_TYPE}/cmocka/include
+            )
+            set_target_properties(
+		    ${T} PROPERTIES LINK_FLAGS "/ignore:4099")
```
sfc-gh-dprzybysz, line 304: why do we need this ignore?

tests/test_get_query_result_response.c
```diff
@@ -26,12 +26,10 @@ void test_get_query_result_response(void **unused) {
     SF_QUERY_RESULT_CAPTURE *result_capture;
     snowflake_query_result_capture_init(&result_capture);
 
-    clear_snowflake_error(&sfstmt->error);
```
sfc-gh-dprzybysz, line 29: why do we need to clear the error?

tests/CMakeLists.txt
sfc-gh-ext-simba-jl: Removed. It's not needed here.

tests/test_get_query_result_response.c
sfc-gh-ext-simba-jl: Clearing the error seems unnecessary here and `clear_snowflake_error()` is an internal function that we don't want to expose in export.def.

exports.def
sfc-gh-ext-simba-jl: The exports are currently manually created (same as the ODBC driver) as we don't want to expose internal functions. There are tools that can generate the exports file but I'm not sure if they can build. I'll test it out and get back to you.

tests/CMakeLists.txt
sfc-gh-dprzybysz: it's not removed :(

sfc-gh-ext-simba-jl: Sorry missed this one. It's now removed.

lib/client_int.h
```diff
@@ -82,7 +82,7 @@
 #define TIMEZONE_OFFSET_RANGE  (int64)(24 * 60);
 
 int uuid4_generate_non_terminated(char *dst);
-int uuid4_generate(char *dst);
+SF_DLL_PUBLIC int uuid4_generate(char *dst);
```
sfc-gh-ext-simba-hx, line 85: Do we really want to expose this to customer? If so we need to move the declaration to public headers under include/snowflake. I prefer not to since no such needs and once we expose it any change there would be a BCR.
If we just need it for test purpose in the worst case we could copy the implementation in test case.

lib/memory.h
```diff
@@ -30,9 +30,9 @@ static SF_INTERNAL_MEM_HOOKS global_hooks = {malloc, free, realloc, calloc};
 void sf_memory_init();
 void sf_memory_term();
 void *sf_malloc(size_t size, const char *file, int line);
-void *sf_calloc(size_t num, size_t size, const char *file, int line);
+SF_DLL_PUBLIC void *sf_calloc(size_t num, size_t size, const char *file, int line);
 void *sf_realloc(void *ptr, size_t size, const char *file, int line);
-void sf_free(void *ptr, const char *file, int line);
+SF_DLL_PUBLIC void sf_free(void *ptr, const char *file, int line);
```
sfc-gh-ext-simba-hx, line 35: Move the declaration to public headers under include/snowflake? we do need to expose this if we want to allow customer use the same logic when migrate from static lib to dynamic, allocate memory in libsfclient and free on application side. Not a good practice for dynamic lib though, allocate and free on the same side would be better, which require adding new interfaces for dynamic lib though.

lib/client_int.h
sfc-gh-ext-simba-jl: It's used for creating database name for test purpose. I've changed the implementation to create a random string so that we don't need to expose this function.

lib/memory.h
sfc-gh-ext-simba-jl: Moved the declaration to platform.h

sfc-gh-dstempniak:
There are some problems on Windows:
```
2025-03-11T21:36:24.2952939Z test_async.obj : error LNK2019: unresolved external symbol _snowflake_init_async_query_result@8 referenced in function _test_new_connection [D:\a\libsnowflakeclient\libsnowflakeclient\cmake-build-win32-VS16-Debug\tests\test_async_dynamic.vcxproj]
2025-03-11T21:36:24.2956103Z test_async.obj : error LNK2019: unresolved external symbol _snowflake_get_query_status@4 referenced in function _test_long_query_get_query_status [D:\a\libsnowflakeclient\libsnowflakeclient\cmake-build-win32-VS16-Debug\tests\test_async_dynamic.vcxproj]
2025-03-11T21:36:24.2958777Z test_async.obj : error LNK2019: unresolved external symbol _snowflake_async_execute@4 referenced in function _test_normal_async_select_query [D:\a\libsnowflakeclient\libsnowflakeclient\cmake-build-win32-VS16-Debug\tests\test_async_dynamic.vcxproj]
2025-03-11T21:36:24.2961570Z test_async.obj : error LNK2019: unresolved external symbol _sf_sleep_ms referenced in function _test_long_query_get_query_status [D:\a\libsnowflakeclient\libsnowflakeclient\cmake-build-win32-VS16-Debug\tests\test_async_dynamic.vcxproj]
2025-03-11T21:36:24.4679560Z D:\a\libsnowflakeclient\libsnowflakeclient\cmake-build-win32-VS16-Debug\tests\Debug\test_async_dynamic.exe : fatal error LNK1120: 4 unresolved externals [D:\a\libsnowflakeclient\libsnowflakeclient\cmake-build-win32-VS16-Debug\tests\test_async_dynamic.vcxproj]
```

PR #787, opened 2024-11-28
SNOW-692968: Async queries support

sfc-gh-ext-simba-nl:
Added snowflake_async_execute to execute async queries and snowflake_async_stmt to get previously run async queries using query id.


include/snowflake/client.h
```diff
@@ -613,6 +632,16 @@ SF_STATUS STDCALL snowflake_get_attribute(
...
 
+/**
+ * Creates sf SNOWFLAKE_STMT context for async queries.
+ *
+ * @param sf The SF_CONNECT context.
+ * @param query_id the query id of the async query.
+ *
+ * @return sfstmt SNOWFLAKE_STMT context for async queries.
+ */
+SF_STMT* STDCALL snowflake_async_stmt(SF_CONNECT *sf, const char *query_id);
```
sfc-gh-ext-simba-hx, line 643: Why we need to create statement specifically for async query execution? I don't see similar approach in JDBC implementation. Adding a new statement attribute would make more sense to me.

lib/client.c
```diff
@@ -54,12 +56,173 @@ static SF_STATUS STDCALL
...
+  log_trace("Error getting query metadata.");
+  return NULL;
+}
+
+/**
+ * Get the status of the query
+ * @param sf the SF_CONNECT context
+ * @param query_id the query id
+ */
+SF_QUERY_STATUS get_query_status(SF_CONNECT *sf, const char *query_id) {
```
sfc-gh-ext-simba-hx, line 143: I think this would be something we expose to application? Please check JDBC async query usage to design what APIs should be expose to application. https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-using#performing-an-asynchronous-query
I think better to use SF_STMT instance for argument.

lib/client.c
```diff
@@ -54,12 +56,173 @@ static SF_STATUS STDCALL
...
+  }
+  return SF_QUERY_STATUS_NO_DATA;
+}
+
+/**
+ * Get the metadata of the query
+ * @param sf the SF_CONNECT context
+ * @param query_id the query id
+ */
+char *get_query_metadata(SF_CONNECT *sf, const char *query_id) {
```
sfc-gh-ext-simba-hx, line 104: I think better to use SF_STMT as argument, query id can be found there.

lib/client.c
```diff
@@ -54,12 +56,173 @@ static SF_STATUS STDCALL
...
+  char *status_query = (char*)SF_CALLOC(1, url_size);
+  sf_sprintf(status_query, url_size, QUERY_MONITOR_URL, query_id);
+
+  if (request(sf, &resp, status_query, NULL, 0, NULL, NULL,
+    GET_REQUEST_TYPE, &sf->error, SF_BOOLEAN_TRUE,
+    0, sf->retry_count, get_retry_timeout(sf),
+    NULL, NULL, NULL, SF_BOOLEAN_FALSE)) {
+
+    s_resp = snowflake_cJSON_Print(resp);
+    log_info("Here is JSON response:\n%s", s_resp);
```
sfc-gh-ext-simba-hx, line 120: trace log please.

lib/client.c
```diff
@@ -54,12 +56,173 @@ static SF_STATUS STDCALL
...
+    cJSON* query = snowflake_cJSON_GetArrayItem(queries, 0);
+
+    char *metadata = snowflake_cJSON_Print(query);
+    snowflake_cJSON_Delete(resp);
+    SF_FREE(s_resp);
+    SF_FREE(status_query);
+    return metadata;
+  }
+  SF_FREE(status_query);
+  log_trace("Error getting query metadata.");
```
sfc-gh-ext-simba-hx, line 134: error or warning log? Maybe add query id.

lib/client.c
```diff
@@ -54,12 +56,173 @@ static SF_STATUS STDCALL
...
+}
+
+/**
+ * Get the status of the query
+ * @param sf the SF_CONNECT context
+ * @param query_id the query id
+ */
+SF_QUERY_STATUS get_query_status(SF_CONNECT *sf, const char *query_id) {
+  SF_QUERY_STATUS ret = SF_QUERY_STATUS_NO_DATA;
+  char *metadata = get_query_metadata(sf, query_id);
```
sfc-gh-ext-simba-hx, line 145: Since this would be exposed to application we need to add error handling (call macro `SET_SNOWFLAKE_ERROR`) when failed to get query status.

lib/client.c
```diff
@@ -1576,6 +1739,33 @@ SF_STMT *STDCALL snowflake_stmt(SF_CONNECT *sf) {
...
+  }
+
+  SF_STMT *sfstmt = (SF_STMT *)SF_CALLOC(1, sizeof(SF_STMT));
+  if (sfstmt) {
+    _snowflake_stmt_reset(sfstmt);
+    sfstmt->connection = sf;
+    sf_strcpy(sfstmt->sfqid, SF_UUID4_LEN, query_id);
+  }
+
+  get_real_results(sfstmt);
```
sfc-gh-ext-simba-hx, line 1754: The main purpose for async query execution is to have non-blocking mode for query execution, so at least we should not call `get_real_results` (which would be blocked until query execution finished) when generating result set from query id. In JDBC `getRealResults` seems being called only when application trying to obtain result metadata or serializable query result.
Personally I would prefer to always return special error code (like `SQL_STILL_EXECUTING` in ODBC) for any API call on a result set (SF_STMT) with async query still running, but I'm OK if you want to keep the same approach as JDBC.

lib/client.c
```diff
@@ -131,29 +133,38 @@ char *get_query_metadata(SF_CONNECT *sf, const char *query_id) {
...
     cJSON* metadataJson = snowflake_cJSON_Parse(metadata);
 
     cJSON* status = snowflake_cJSON_GetObjectItem(metadataJson, "status");
-    if (snowflake_cJSON_IsString(status))
-    {
+    if (snowflake_cJSON_IsString(status)) {
       char* queryStatus = snowflake_cJSON_GetStringValue(status);
       ret = get_status_from_string(queryStatus);
     }
+    else {
```
sfc-gh-ext-simba-hx, line 152: After seeing the logic in `get_real_results` I'm not quite sure whether we should set error here. Seems no data is acceptable, the error log in `get_query_metadata` would be the same.

lib/client.c
```diff
@@ -199,28 +213,44 @@ void get_real_results(SF_STMT * sfstmt) {
         return;
       }
     }
-  }
-  int sleep_time = retry_pattern[retry] * 500;
+
+    int sleep_time = retry_pattern[retry] * 500;
 #ifdef _WIN32
-  Sleep(sleep_time);
```
sfc-gh-ext-simba-hx: define this as a common function in platform.c? I think same logic being used in http_perform.c as well.

tests/test_async.c
sfc-gh-ext-simba-hx, line 1: Please add test cases to cover typical use case, checking query status before calling any fetch/metadata API. Take JDBC doc as example, https://docs.snowflake.com/en/developer-guide/jdbc/jdbc-using#examples-of-asynchronous-queries
Add usage example in PR description as well (could be reference to test case if the code is clean enough).
Calling fetch/metadata function without checking query status also needed. Please test with queries would take some time.

lib/connection.c
```diff
@@ -414,41 +413,50 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
             break;
         }
 
-        while (strcmp(query_code, QUERY_IN_PROGRESS_CODE) == 0 ||
-               strcmp(query_code, QUERY_IN_PROGRESS_ASYNC_CODE) == 0) {
+        sf_bool isAsyncExec = SF_BOOLEAN_FALSE;
+        cJSON *json_body = snowflake_cJSON_Parse(body);
```
sfc-gh-jszczerbinski, line 417: Error handling?

sfc-gh-jszczerbinski (changes requested):
Please remove unnecessary formatting changes. It's hard to review with so much (unnecessary?) changes.

lib/connection.c
```diff
@@ -414,41 +413,50 @@ sf_bool STDCALL curl_post_call(SF_CONNECT *sf,
...
+
+        if (!isAsyncExec) {
+          while (strcmp(query_code, QUERY_IN_PROGRESS_CODE) == 0 ||
+            strcmp(query_code, QUERY_IN_PROGRESS_ASYNC_CODE) == 0) {
             // Remove old result URL and query code if this isn't our first rodeo
             SF_FREE(result_url);
             memset(query_code, 0, QUERYCODE_LEN);
             data = snowflake_cJSON_GetObjectItem(*json, "data");
             if (json_copy_string(&result_url, data, "getResultUrl") !=
-                SF_JSON_ERROR_NONE) {
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: I think VS automatically did that accidentally when I wrapped it in an if clause. I'll revert these

tests/test_async.c
```diff
@@ -0,0 +1,244 @@
...
+
+    /* get results */
+    int64 out = 0;
+    assert_int_equal(snowflake_num_rows(sfstmt), 1);
+
+    int counter = 0;
+    while ((status = snowflake_fetch(sfstmt)) == SF_STATUS_SUCCESS) {
+      snowflake_column_as_int64(sfstmt, 1, &out);
+      assert_int_equal(out, 1);
+      ++counter;
```
sfc-gh-mmishchenko, line 37: Looks `counter` isn't used anywhere

tests/test_async.c
```diff
@@ -0,0 +1,244 @@
...
+
+  int retries = 0;
+  while (query_status != SF_QUERY_STATUS_SUCCESS || retries > 5) {
+    query_status = snowflake_get_query_status(sfstmt);
+    sf_sleep_ms(2000);
+    retries++;
+  }
+
+  /* get results */
+  char *out = NULL;
```
sfc-gh-mmishchenko, line 79: `out` buffer is never released

tests/test_async.c
```diff
@@ -0,0 +1,244 @@
...
+  status = snowflake_prepare(sfstmt, "select system$wait(5);", 0);
+  assert_int_equal(status, SF_STATUS_SUCCESS);
+  status = snowflake_async_execute(sfstmt);
+  if (status != SF_STATUS_SUCCESS) {
+    dump_error(&(sfstmt->error));
+  }
+  assert_int_equal(status, SF_STATUS_SUCCESS);
+
+  /* get results */
+  char* out = NULL;
```
sfc-gh-mmishchenko, line 118: `out` buffer is never released

tests/test_async.c
```diff
@@ -0,0 +1,244 @@
...
+  SF_STMT* sfstmt = snowflake_stmt(sf);
+  status = snowflake_prepare(sfstmt, "select 1;", 0);
+  assert_int_equal(status, SF_STATUS_SUCCESS);
+  status = snowflake_async_execute(sfstmt);
+  if (status != SF_STATUS_SUCCESS) {
+    dump_error(&(sfstmt->error));
+  }
+  assert_int_equal(status, SF_STATUS_SUCCESS);
+
+  char* sfqid = (char*)SF_CALLOC(1, SF_UUID4_LEN);
```
sfc-gh-mmishchenko, line 156: `sfqid` is never released

sfc-gh-mmishchenko (changes requested):

include/snowflake/client.h
```diff
@@ -294,6 +294,25 @@ typedef enum SF_STMT_ATTRIBUTE {
     SF_STMT_USER_REALLOC_FUNC
 } SF_STMT_ATTRIBUTE;
 
+/**
+ * The query status
+ */
+typedef enum SF_QUERY_STATUS {
+    SF_QUERY_STATUS_RUNNING,
```
sfc-gh-dprzybysz, line 301: let's sort statuses alphabetically

lib/client.c
```diff
@@ -54,12 +56,193 @@ static SF_STATUS STDCALL
 _reset_connection_parameters(SF_CONNECT *sf, cJSON *parameters,
                              cJSON *session_info, sf_bool do_validate);
 
+static const char* query_status_names[] = {
+    "RUNNING",
```
sfc-gh-dprzybysz, line 60: can we keep the statuses sorted?

lib/client.c
```diff
@@ -54,12 +56,193 @@ static SF_STATUS STDCALL
...
  */
 sf_bool validate_application(const char *application);
 
+/**
+ * Helper function to get SF_QUERY_STATUS given the string representation
+ * @param query_status the string representation of the query status
+ */
+SF_QUERY_STATUS get_status_from_string(const char *query_status) {
+  if (query_status == NULL) {
+    return SF_QUERY_STATUS_NO_DATA;
```
sfc-gh-dprzybysz, line 87: is it no data or more UNKNOWN?

lib/client.c
```diff
@@ -54,12 +56,193 @@ static SF_STATUS STDCALL
...
+ * @param sfstmt The SF_STMT context.
+ * 
+ * The query metadata
+ */
+char *get_query_metadata(SF_STMT* sfstmt) {
+  cJSON *resp = NULL;
+  cJSON *data = NULL;
+  cJSON *queries = NULL;
+  char *s_resp = NULL;
+  size_t url_size = strlen(QUERY_MONITOR_URL) -2 + strlen(sfstmt->sfqid) + 1;
```
sfc-gh-dprzybysz, line 111: let's apply code formatter on changed files

tests/test_async.c
```diff
@@ -0,0 +1,245 @@
...
+
+int main(void) {
+    initialize_test(SF_BOOLEAN_FALSE); 
+    const struct CMUnitTest tests[] = {
+      cmocka_unit_test(test_select),
+      cmocka_unit_test(test_query_status),
+      cmocka_unit_test(test_premature_fetch),
+      cmocka_unit_test(test_new_connection),
+      cmocka_unit_test(test_fake_table),
+      cmocka_unit_test(test_invalid_query_id),
```
sfc-gh-dprzybysz, line 240: let's add a test that fetches chunks

lib/client.c
```diff
@@ -54,12 +56,194 @@ static SF_STATUS STDCALL
...
+    cJSON* query = snowflake_cJSON_GetArrayItem(queries, 0);
+
+    char *metadata = snowflake_cJSON_Print(query);
+    snowflake_cJSON_Delete(resp);
+    SF_FREE(s_resp);
+    SF_FREE(status_query);
+    return metadata;
+  }
+  SF_FREE(status_query);
+  log_info("No query metadata found. Query id: %s", sfstmt->sfqid);
```
sfc-gh-jszczerbinski, line 136: log_error? Seems this would happen if there is a bug in our code or api is misused.

lib/client.c
```diff
@@ -54,12 +56,194 @@ static SF_STATUS STDCALL
...
+  return NULL;
+}
+
+
+SF_QUERY_STATUS STDCALL snowflake_get_query_status(SF_STMT *sfstmt) {
+  SF_QUERY_STATUS ret = SF_QUERY_STATUS_NO_DATA;
+  char *metadata = get_query_metadata(sfstmt);
```
sfc-gh-jszczerbinski, line 143: We do not free the metadata

lib/client.c
```diff
@@ -54,12 +56,194 @@ static SF_STATUS STDCALL
...
+  char* query_template = "select * from table(result_scan('%s'))";
+  sf_sprintf(query, strlen(query_template) - 2 + strlen(sfstmt->sfqid) + 1, query_template, sfstmt->sfqid);
+  SF_STATUS ret = snowflake_query(sfstmt, query, strlen(query));
+  if (ret != SF_STATUS_SUCCESS) {
+    snowflake_propagate_error(sfstmt->connection, sfstmt);
+    return SF_BOOLEAN_FALSE;
+  }
+
+  // Get query stats
+  char* metadata_str = get_query_metadata(sfstmt);
```
sfc-gh-jszczerbinski, line 227: We do not free metadata_str?

lib/client.c
```diff
@@ -54,12 +56,194 @@ static SF_STATUS STDCALL
...
+}
+
+/**
+ * Get the metadata of the query
+ * 
+ * @param sfstmt The SF_STMT context.
+ * 
+ * The query metadata
+ */
+char *get_query_metadata(SF_STMT* sfstmt) {
```
sfc-gh-jszczerbinski, line 107: Maybe this function should return a striuct with parsed metadata instead of char array containing JSON. Consider moving parsing and validation into this function

lib/client.c
```diff
@@ -54,12 +56,194 @@ static SF_STATUS STDCALL
...
+  char *status_query = (char*)SF_CALLOC(1, url_size);
+  sf_sprintf(status_query, url_size, QUERY_MONITOR_URL, sfstmt->sfqid);
+
+  if (request(sfstmt->connection, &resp, status_query, NULL, 0, NULL, NULL,
+    GET_REQUEST_TYPE, &sfstmt->error, SF_BOOLEAN_TRUE,
+    0, sfstmt->connection->retry_count, get_retry_timeout(sfstmt->connection),
+    NULL, NULL, NULL, SF_BOOLEAN_FALSE)) {
+
+    s_resp = snowflake_cJSON_Print(resp);
+    log_trace("Here is JSON response:\n%s", s_resp);
```
sfc-gh-jszczerbinski, line 122: Please be more descriptive in logs, maybe: log_trace("GET %s returned response: %s", status_query, s_resp)

lib/client.c
sfc-gh-jszczerbinski: Especially since stringify it in this function and parse it again when it's used

sfc-gh-jszczerbinski (changes requested):
Looks good. Still needs some work.

include/snowflake/platform.h
```diff
@@ -142,6 +142,9 @@ void STDCALL sf_memory_error_handler();
 // this should be called by application before any calls of sfclient
 void STDCALL sf_exception_on_memory_failure();
 
+void STDCALL sf_sleep_ms(int sleep_ms);
```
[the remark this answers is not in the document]
sfc-gh-dprzybysz: The method is still defined in platform.h - do we need it here or could the declaration be moved to lib directory?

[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: I can move it to util instead, that way it stays in the lib directory

lib/client.c
```diff
@@ -2684,6 +2907,9 @@ static SF_STATUS _snowflake_execute_with_binds_ex(SF_STMT* sfstmt,
                                   is_string_empty(sfstmt->connection->directURL) ?
                                   NULL : sfstmt->request_id, is_describe_only,
                                   sfstmt->multi_stmt_count);
+								  
+	  snowflake_cJSON_AddBoolToObject(body, "asyncExec", is_async_exec);
```
sfc-gh-mkubik, line 2911: formatting

lib/client.c
```diff
@@ -55,12 +57,197 @@ static SF_STATUS STDCALL
...
+      log_error(
+        "Cannot retrieve data on the status of this query. Max retries hit with queryID=%s", sfstmt->sfqid);
+      break;
+    }
+    query_status = snowflake_get_query_status(sfstmt);
+  }
+
+  // Get query results
+  char query[1024];
+  char* query_template = "select * from table(result_scan('%s'))";
```
sfc-gh-mmishchenko, line 226: This approach with use RESULT_SCAN to retrieve results has a significant flaw - [the original order of the results isn't guaranteed](https://docs.snowflake.com/en/sql-reference/functions/result_scan). Probably we should use /queries/{sfqid}/result endpoint instead.

sfc-gh-mmishchenko (changes requested):

lib/client.c
```diff
@@ -55,12 +57,194 @@ static SF_STATUS STDCALL
...
+  }
+  SF_FREE(status_query);
+  log_error("No query metadata found. Query id: %s", sfstmt->sfqid);
+  query_metadata->status = SF_QUERY_STATUS_UNKNOWN;
+  return query_metadata;
+}
+
+
+SF_QUERY_STATUS STDCALL snowflake_get_query_status(SF_STMT *sfstmt) {
+  SF_QUERY_STATUS ret = SF_QUERY_STATUS_NO_DATA;
```
sfc-gh-dprzybysz, line 160: do we need the default value set or we can join this line with  extracteing the exact status?

lib/client.c
```diff
@@ -55,12 +57,194 @@ static SF_STATUS STDCALL
...
+    (query_status == SF_QUERY_STATUS_RESUMING_WAREHOUSE) ||
+    (query_status == SF_QUERY_STATUS_QUEUED_REPAIRING_WAREHOUSE) ||
+    (query_status == SF_QUERY_STATUS_NO_DATA);
+}
+
+/**
+ * Get the results of the async query
+ * @param sfstmt The SF_STMT context
+ */
+sf_bool get_real_results(SF_STMT *sfstmt) {
```
sfc-gh-dprzybysz, line 183: I am not sure if we need a method that effectively converts async query into sync query

lib/client.c
```diff
@@ -55,12 +57,194 @@ static SF_STATUS STDCALL
...
+  }
+
+  // Get query results
+  if (snowflake_next_result(sfstmt) != SF_STATUS_SUCCESS) {
+    snowflake_propagate_error(sfstmt->connection, sfstmt);
+    return SF_BOOLEAN_FALSE;
+  }
+
+  // Get query stats
+  SF_QUERY_METADATA *metadata = get_query_metadata(sfstmt);
```
sfc-gh-dprzybysz, line 232: we are fetching query metadata above to get status - do we need to get metadata again here or we can reuse the last result?

include/snowflake/client.h
```diff
@@ -543,6 +563,8 @@ typedef struct SF_STMT {
     int64 paramset_size;
     sf_bool array_bind_supported;
     int64 affected_rows;
+    sf_bool is_async;
```
sfc-gh-dprzybysz, line 566: could we add brief description - is_async means that the query is async

include/snowflake/client.h
```diff
@@ -543,6 +563,8 @@ typedef struct SF_STMT {
...
+    sf_bool is_async_initialized;
```
sfc-gh-dprzybysz, line 567: does it mean that the async is initialized? is query running now, or we fetched data when true?

lib/client.c
```diff
@@ -2263,6 +2454,23 @@ SF_STMT *STDCALL snowflake_stmt(SF_CONNECT *sf) {
     return sfstmt;
 }
 
+SF_STMT *STDCALL snowflake_create_async_query_result(SF_CONNECT *sf, const char *query_id) {
```
sfc-gh-dprzybysz, line 2457: create or more init?

lib/client.c
```diff
@@ -3081,13 +3313,32 @@ int64 STDCALL snowflake_num_rows(SF_STMT *sfstmt) {
         return -1;
     }
 
+    if (sfstmt->is_async && !sfstmt->is_async_initialized) {
+      if (get_real_results(sfstmt)) {
+        sfstmt->is_async_initialized = SF_BOOLEAN_TRUE;
+      }
+      else {
```
sfc-gh-dprzybysz, line 3320: please format files according to other files formatting

lib/client.c
```diff
@@ -3081,13 +3313,32 @@ int64 STDCALL snowflake_num_rows(SF_STMT *sfstmt) {
...
     if (!sfstmt) {
         return -1;
     }
+
+    if (sfstmt->is_async && !sfstmt->is_async_initialized) {
+      if (get_real_results(sfstmt)) {
+        sfstmt->is_async_initialized = SF_BOOLEAN_TRUE;
+      }
+      else {
+        return -1;
```
sfc-gh-dprzybysz, line 3338: even when the query is not finished we should now the metadata and number of fields?

lib/client.c
```diff
@@ -3096,6 +3347,16 @@ uint64 STDCALL snowflake_num_params(SF_STMT *sfstmt) {
         // TODO change to -1?
         return 0;
     }
+
+    if (sfstmt->is_async && !sfstmt->is_async_initialized) {
+      if (get_real_results(sfstmt)) {
+        sfstmt->is_async_initialized = SF_BOOLEAN_TRUE;
+      }
+      else {
+        return 0;
```
sfc-gh-dprzybysz, line 3356: even when the query is not finished we should now the metadata and number of parameters?

tests/test_async.c
```diff
@@ -0,0 +1,362 @@
+/*
+ * Copyright (c) 2024 Snowflake Computing, Inc. All rights reserved.
```
sfc-gh-dprzybysz, line 2: we have 2025 :)

tests/test_async.c
```diff
@@ -0,0 +1,362 @@
...
+ */
+#include <string.h>
+#include "memory.h"
+#include "utils/test_setup.h"
+#include "util.h"
+
+ /**
+  * Test normal query flow with async
```
sfc-gh-dprzybysz, line 10: let's rename the test function to express the same what comment says and remove the comment, also in other cases

tests/test_async.c
```diff
@@ -0,0 +1,362 @@
...
+    }
+    assert_int_equal(status, SF_STATUS_SUCCESS);
+
+    /* get results */
+    int64 out = 0;
+    assert_int_equal(snowflake_num_fields(sfstmt), 1);
+    assert_int_equal(snowflake_num_rows(sfstmt), 1);
+
+    while ((status = snowflake_fetch(sfstmt)) == SF_STATUS_SUCCESS) {
+      snowflake_column_as_int64(sfstmt, 1, &out);
```
sfc-gh-dprzybysz, line 36: let's add count and assert after while that it is equal to 1 - now empty result set passes makes the while loop not making any assertion
also in other tests

tests/test_async.c
```diff
@@ -0,0 +1,362 @@
...
+    if (status != SF_STATUS_EOF) {
+      dump_error(&(sfstmt->error));
+    }
+    assert_int_equal(status, SF_STATUS_EOF);
+    snowflake_stmt_term(sfstmt);
+    snowflake_term(sf);
+}
+
+/**
+ * Test normal getting query status
```
sfc-gh-dprzybysz, line 48: system$wait is not normal query - let's say long running query

tests/test_async.c
```diff
@@ -0,0 +1,362 @@
...
+  assert_int_equal(status, SF_STATUS_EOF);
+  snowflake_stmt_term(sfstmt);
+  snowflake_term(sf);
+  SF_FREE(out);
+}
+
+/**
+ * Test premature fetch
+ */
+void test_premature_fetch() {
```
sfc-gh-dprzybysz, line 100: where exactly the premature fetch is, can we make it fail?

tests/test_async.c
```diff
@@ -0,0 +1,362 @@
...
+  /* query */
+  SF_STMT* sfstmt = snowflake_stmt(sf);
+  status = snowflake_prepare(sfstmt, "select * from my_fake_table;", 0);
+  assert_int_equal(status, SF_STATUS_SUCCESS);
+  status = snowflake_async_execute(sfstmt);
+  assert_int_equal(status, SF_STATUS_SUCCESS);
+
+  /* get results */
+  status = snowflake_fetch(sfstmt);
+  assert_int_equal(status, SF_STATUS_ERROR_GENERAL);
```
sfc-gh-dprzybysz, line 209: are we able to get there exact message why the query cannot be performed?

lib/client.c
sfc-gh-ext-simba-nl: I wasn't able to find the value for this in the monitoring/queries/{qid} call

lib/client.c
sfc-gh-ext-simba-nl: I wasn't able to find the value for this in the monitoring/queries/{qid} call

lib/client.c
sfc-gh-dprzybysz: but don't we get metadata when async query is started?

lib/client.c
```diff
@@ -55,12 +57,197 @@ static SF_STATUS STDCALL
...
+        SET_SNOWFLAKE_STMT_ERROR(&sfstmt->error,
+          SF_STATUS_ERROR_GENERAL,
+          "Cannot retrieve data on the status of this query.",
+          NULL,
+          sfstmt->sfqid);
+        break;
+      }
+    }
+
+    int sleep_time = retry_pattern[retry] * 500;
```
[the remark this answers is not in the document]
sfc-gh-mkubik: Sorry, was this addressed?

sfc-gh-mkubik (commented):
We fixed the code quality check so now it can be turned green again so please take a look as well

lib/client.c
sfc-gh-ext-simba-nl: Removed

lib/client.c
sfc-gh-ext-simba-nl: Removed and reused last result

lib/client.c
sfc-gh-ext-simba-nl: Nope, when I execute the async query, all I get back is this:
`{
	"data":	{
		"queryId":	"01ba594e-0a0e-efe0-0000-53b10a6c842e",
		"getResultUrl":	"/queries/01ba594e-0a0e-efe0-0000-53b10a6c842e/result",
		"queryAbortsAfterSecs":	300,
		"progressDesc":	null
	},
	"code":	"333334",
	"message":	"Asynchronous execution in progress. Use provided query id to perform query monitoring and management.",
	"success":	true
}`
I only get proper metadata when I call the getResultUrl, which I do during get_real_results

lib/client.c
sfc-gh-ext-simba-nl: I saw a similar implementation in JDBC and used it here. It's only a helper function that gets called when the user calls fetch/num_params/etc. Essentially when it's time to get the results

sfc-gh-ext-simba-nl:
Hi @sfc-gh-mkubik,
I looked at the code quality and the logs still says "Baseline file does not exist."

sfc-gh-mkubik (approved):

cpp/util/SnowflakeCommon.cpp
```diff
@@ -12,6 +12,7 @@
 #include "../logger/SFLogger.hpp"
 #include <chrono>
 #include "SnowflakeCommon.hpp"
+#include <chrono>
```
sfc-gh-jszczerbinski, line 15: Please remove it. I added the missing header in a separate PR

include/snowflake/client.h
```diff
@@ -606,6 +628,15 @@ typedef struct SF_TIMESTAMP {
     SF_DB_TYPE ts_type;
 } SF_TIMESTAMP;
 
+/**
+ * Query metadata
+ */
+typedef struct SF_QUERY_METADATA {
+  SF_QUERY_STATUS status;
+  char *qid;
+  char *stats;
```
sfc-gh-jszczerbinski, line 637: change type to SF_STATS*

lib/client.c
```diff
@@ -57,12 +59,204 @@ static SF_STATUS STDCALL
...
+      sf_sprintf(msg, sizeof(msg),
+        "Cannot retrieve data on the status of this query. Max retries hit. Query status: %d", query_status);
+      SET_SNOWFLAKE_STMT_ERROR(&sfstmt->error,
+        SF_STATUS_ERROR_GENERAL,
+        msg,
+        NULL,
+        sfstmt->sfqid);
+      return SF_BOOLEAN_FALSE;
+    }
+    metadata = get_query_metadata(sfstmt);
```
sfc-gh-jszczerbinski, line 233: We are leaking here, since metadata already points to allocated object

lib/client.c
```diff
@@ -57,12 +59,204 @@ static SF_STATUS STDCALL
...
+  }
+
+  // Get query stats
+  if (metadata->stats) {
+    cJSON* stats = snowflake_cJSON_Parse(metadata->stats);
+    if (snowflake_cJSON_IsObject(stats)) {
+      if (sfstmt->stats) {
+        SF_FREE(sfstmt->stats);
+      }
+      sfstmt->stats = set_stats(stats);
```
sfc-gh-jszczerbinski, line 250: Are we setting stats both here and in get_query_metadata? Is the stats field even necessary?

sfc-gh-jszczerbinski (changes requested):

lib/client.c
sfc-gh-ext-simba-nl: No, not anymore. I can remove it actually

include/snowflake/client.h
sfc-gh-ext-simba-nl: Will remove entirely

sfc-gh-jszczerbinski (approved):
LGTM

PR #789, opened 2024-12-02
SNOW-1821508: OCSP deprecation plan steps

sfc-gh-ext-simba-hx:
teamwork issue 1179
- targeting 2.0.0
- default behavior from FAIL_CLOSE to FAIL_OPEN
- FAIL_OPEN log changed (log level unchanged since curl only have info and error log, and the log is info level already)
- log error for responder URL missing only when it's FAIL_CLOSE
- FAIL_CLOSE retry from 3 to 2.  FAIL_OPEN retry and cache server retry are 1 already unchanged.
- Confirmed timeout is correct already.
- No change for `insecureMode` since libsfclient doesn't have such attribute and is using `SF_GLOBAL_OCSP_CHECK` to enable/disable OCSP check.


deps/curl/lib/vtls/sf_ocsp.c
```diff
@@ -1718,20 +1718,7 @@ OCSP_RESPONSE * getOCSPResponse(X509 *cert, X509 *issuer,
...
-              "responder. Details:%s",ocsp_log_data);
-  if (ocsp_log_data)
-  {
-    if (oob_enable)
-    {
-      sendOOBevent(ocsp_log_data);
-    }
-    if(ocsp_log_data) sf_curl_cJSON_free(ocsp_log_data);
-  }
+  infof(data, "OCSP responder didn't respond correctly. Assuming certificate is not revoked.");
```
sfc-gh-pfus, line 1721: Can we preserve details as it was before, please?

sfc-gh-ext-simba-hx: done

sfc-gh-pfus (approved):

sfc-gh-jszczerbinski (approved):
LGTM

PR #791, opened 2024-12-11
SNOW-1852178 Remove the deprecated part of API

sfc-gh-mmishchenko:
- [ ] Removed CRT function wrappers
- [ ] Removed max LOB size as a macro
- [ ] Removed deprecating macros as non-used anymore


include/snowflake/SF_CRTFunctionSafe.h
```diff
@@ -26,40 +26,6 @@ extern "C" {
     #endif
 #endif
 
-// Defined for #pragma warning messages.
-#define MACRO_TO_STRING2(x) #x
-#define MACRO_TO_STRING(x) MACRO_TO_STRING2(x)
-
-#if defined(_MSC_VER)
-#define SF_MACRO_DEPRECATED_WARNING(MSG) __pragma(message ( __FILE__ "(" MACRO_TO_STRING(__LINE__) ") : warning C4996: " MSG))
```
sfc-gh-dprzybysz, line 34: Please move the macro to a different place - I think there will be more places we would like to mark in the future

sfc-gh-dprzybysz (changes requested):
1. Please add Jira ID to the PR name
2. The PR should be to master-2.0.0, not to the master branch

include/snowflake/SF_CRTFunctionSafe.h
sfc-gh-mmishchenko: Makes sense. Though I don't see currently a proper header to keep these macros - shall we create new `common.h`?

sfc-gh-dprzybysz: common.h could be created or we can add it to include/snowflake/version.h - deprecating is related to the versioning

sfc-gh-dprzybysz: platform.h is fine I think.

sfc-gh-dprzybysz (approved):

PR #813, opened 2025-01-23
SNOW-715619: Support authentication with external browser

sfc-gh-ext-simba-jy:


CMakeLists.txt
```diff
@@ -248,6 +248,7 @@ endif ()
 
 if (APPLE)
     # macOS
+    find_library(CORESERVICES_LIBRARY CoreServices)
```
sfc-gh-ext-simba-hx, line 251: For static library we should not add extra linking libs here but should be in the link options for test cases:
https://github.com/snowflakedb/libsnowflakeclient/blob/master/tests/CMakeLists.txt#L235
@sfc-gh-ext-simba-jl Just a heads up you might need to adjust build options for dynamic lib when merging your PR with the changes here.
Also CoreServices seems to be a larger one FYI ODBC are using following build options on MacOS maybe we could use the same?
```
if (APPLE)
    target_link_libraries(snowflakeclient INTERFACE
        "-framework CoreFoundation"
        "-framework SystemConfiguration"
        "-framework ApplicationServices"
        "-framework Security"
        )
endif ()
```

include/snowflake/IAuth.hpp
```diff
@@ -122,6 +135,101 @@ namespace IAuth
         std::string oneTimeToken;
         std::string m_samlResponse;
     };
+
+    class IAuthWebServer : public AuthErrorHandler
```
sfc-gh-ext-simba-hx, line 139: Why we need IAuthWebServer defined in external header file? Does it include driver specific logic that needs to be overridden on ODBC or any other driver side?
For my understanding in ODBC IAuthWebServer is defined as pure interface without any implementation, for test purpose only so it can be overridden in unit test cases.
If we want to keep test cases on both ODBC and libsfclient side, keep IAuthWebServer as a pure interface would be better. Or if test it on libsfclient side is enough, we could move the definition to internal header files.

sfc-gh-ext-simba-jy:
Because of many commit lists, I moved this to https://github.com/snowflakedb/libsnowflakeclient/pull/834

PR #824, opened 2025-02-03
SNOW-1865263 Remove VS14 support in libsnowflakeclient

sfc-gh-mmishchenko:
- [x] Drop VS14 build support
- [x] Build Apache Arrow library from source by default
- [x] Get rid of the environment variable which sets how the dependency on Apache Arrow is resolved (it's a cmake parameter now)
- [x] Boost is explicitly built as a separate dependency but not as a part of Arrow
- [x] Fix scripts/build_dependencies -t Debug


scripts/_init.bat
```diff
@@ -84,13 +83,8 @@ if /I "%vs_version%"=="VS15" (
     set cmake_generator=Visual Studio 15 2017
     set vsdir=vs15
 )
-if /I "%vs_version%"=="VS14" (
```
sfc-gh-dprzybysz, line 87: let's remove also the directory https://github.com/snowflakedb/libsnowflakeclient/tree/master/deps-build

scripts/build_arrow_source.sh
```diff
@@ -11,6 +11,11 @@ function usage() {
 }
 set -o pipefail
 
+export ARROW_SRC_VERSION=15.0.0
```
sfc-gh-dprzybysz, line 14: can we rename the file to just `build_arrow.sh`?

sfc-gh-dprzybysz, line 14: don't we need changes in https://github.com/snowflakedb/libsnowflakeclient/blob/master/scripts/build_arrow_source.bat ?

deps-build/.gitignore
```diff
@@ -1,4 +0,0 @@
-darwin/
```
sfc-gh-jszczerbinski, line 1: Why remove .gitignore? Maybe add deps-build to toplevel .gitignore

CMakeLists.txt
```diff
@@ -4,6 +4,11 @@
 #
 cmake_minimum_required(VERSION 3.17)
 project(snowflakeclient)
+
+option(BUILD_TESTS "True if build tests" on)
+option(MOCK "True if mock should be used" off)
+option(ARROW_FROM_SOURCE "Build Apache Arrow library from source" on)
```
sfc-gh-jszczerbinski, line 10: Why not remove it altogether?

sfc-gh-jszczerbinski (approved):

sfc-gh-mkubik (approved):

deps-build/.gitignore
sfc-gh-mmishchenko: Indeed, I was in the assumption it was there

scripts/build_arrow.sh
```diff
@@ -16,31 +16,91 @@ export ARROW_BUILD_VERSION=6
 #The full version number for dependency packaging/uploading/downloading
```
sfc-gh-jszczerbinski, line 16: Bump build version since we change build script

sfc-gh-jszczerbinski (changes requested):
We need to bump BUILD versions in all of the modified scripts

sfc-gh-jszczerbinski (approved):

PR #825, opened 2025-02-03
SNOW-1908931: Print warnings report to stdout and filter warnings from deps

sfc-gh-jszczerbinski:


ci/scripts/warning_report.sh
```diff
@@ -21,6 +21,7 @@ then
     exit 0
   else
     echo "Warnings check failed. Check workflow summary to find out the details."
```
sfc-gh-dprzybysz, line 23: do we need this echo when result is printed?

sfc-gh-mkubik (approved):

ci/scripts/warning_report.sh
sfc-gh-jszczerbinski: No, but I would leave it as hint that you can look to workflow summary for rendered markdown.

sfc-gh-dprzybysz (approved):

ci/scripts/warning_report.sh
sfc-gh-dprzybysz: Maybe rephrase then - but I can live with it

PR #836, opened 2025-02-28
SNOW-1739603: Implement toml config file parsing

sfc-gh-ext-simba-nl:
SNOW-1739603: Implement toml config file parsing in libsnowflakeclient to be used in other connectors.
Added toml_config_parser file with load_toml_config, returns a `key=value;` style string with parsed parameters.
Behaviour: If `SNOWFLAKE_HOME` exists, searches for toml config file there, otherwise defaults to `~/.snowflake`.
Within toml config file, if `SNOWFLAKE_DEFAULT_CONNECTION_NAME` defined, use that as configuration name, otherwise defaults to `default`.


sfc-gh-ext-simba-nl:
Hi,
For some reason, build_tomlplusplus.sh isn't updating to the latest commit. Even if I remove the file and add it back, it's still running the version I first committed. Can someone help me look into why it's doing that?
Thanks

include/snowflake/toml_config_parser.h
```diff
@@ -0,0 +1,22 @@
...
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+  /**
+    * Load toml configuration file.
+    * @param connectionParams       The connection parameters parsed from the toml configuration file in k=v; format
+    */
+  void load_toml_config(char** connectionParams);
```
sfc-gh-jszczerbinski, line 16: I think we should expose the parameters via C++ API, for example sth like this?
`std::map<std::string, std::string> loadTomlConfig()`

sfc-gh-jszczerbinski (commented):
Changed permissions for build scripts on linux and now they fail: please fix it.
I have some issues with general design of this feature - passing the parameters as connection string, I will re-review the code after the changes

tests/test_toml_config.cpp
```diff
@@ -0,0 +1,200 @@
...
+#include "utils/test_setup.h"
+#include "utils/TestSetup.hpp"
+
+void test_valid_toml_file(void** unused) {
+  SF_UNUSED(unused);
+  // Create toml file
+  char tomlConfig[] = "[default]\nkey1 = \"value1\"\nkey2 = \"value2\"";
+  char tomlFilePath[] = "./connections.toml";
+  FILE* file;
+  file = fopen(tomlFilePath, "w");
```
sfc-gh-jszczerbinski, line 15: Use c++ fstream API instead. Same for other tests

tests/test_toml_config.cpp
```diff
@@ -0,0 +1,200 @@
...
+  fprintf(file, "%s", tomlConfig);
+  fclose(file);
+
+  char envbuf[MAX_PATH + 1];
+  char* snowflakeHome = sf_getenv_s("SNOWFLAKE_HOME", envbuf, sizeof(envbuf));
+  sf_setenv("SNOWFLAKE_HOME", "./");
```
sfc-gh-jszczerbinski, line 21: Please extract `EnvOverride` class from `test_unit_secure_storage.cpp` to a new file `tests/utils/EnvOverride.hpp`. It will make sure we cleanup env after test finishes without being too verbose. Same for other tests

tests/test_toml_config.cpp
```diff
@@ -0,0 +1,200 @@
...
+  remove(tomlFilePath);
+}
+
+int main(void) {
+  initialize_test(SF_BOOLEAN_FALSE);
+  const struct CMUnitTest tests[] = {
+      cmocka_unit_test(test_missing_toml_file),
+      cmocka_unit_test(test_invalid_toml_file),
+      cmocka_unit_test(test_client_config_log_invalid_config_name),
+#if (!defined(_WIN32) && !defined(_DEBUG)) || defined(_WIN64)
```
sfc-gh-jszczerbinski, line 193: Why do we need this ifdef?

tests/test_toml_config.cpp
```diff
@@ -0,0 +1,200 @@
...
+ */
+
+#include <snowflake/TomlConfigParser.hpp>
+#include "utils/test_setup.h"
+#include "utils/TestSetup.hpp"
+
+void test_valid_toml_file(void** unused) {
+  SF_UNUSED(unused);
+  // Create toml file
+  char tomlConfig[] = "[default]\nkey1 = \"value1\"\nkey2 = \"value2\"";
```
sfc-gh-jszczerbinski, line 12: Prefer std::string over char[], same in other tests

cpp/util/SnowflakeCommon.cpp
```diff
@@ -12,6 +12,7 @@
 #include "../logger/SFLogger.hpp"
 #include <chrono>
 #include "SnowflakeCommon.hpp"
+#include <chrono>
```
sfc-gh-jszczerbinski, line 15: Remove the include, it's not needed. The same file is included in 13th line of this file

sfc-gh-jszczerbinski (changes requested):

tests/test_toml_config.cpp
sfc-gh-ext-simba-nl: on 32-bit windows debug build, the toml::parse_file call causes a "(_osfile(fh) & FOPEN)" debug assertion causing test cases to hang indefinitely. It gets bypassed with the ifdef in load_toml_config, but the rest of the test case is irrelevant anyways because of that. So I can either ifdef each test case separately to check for nothing or just not run it entirely since it doesn't actually test for anything

sfc-gh-ext-simba-nl:
Thanks @sfc-gh-dstempniak , I'll take a look

cpp/lib/TomlConfigParser.cpp
```diff
@@ -0,0 +1,111 @@
...
+    return connectionParams;
+  }
+}
+
+std::map<std::string, std::string> load_toml_config()
+{
+  std::map<std::string, std::string> params;
+// Disable toml config parsing for 32-bit windows debug build due to linking issues
+// with _osfile causing hanging/assertions until dynamic linking is available
+#if (!defined(_WIN32) && !defined(_DEBUG)) || defined(_WIN64)
```
sfc-gh-jszczerbinski, line 103: I think we should support win32

cpp/lib/TomlConfigParser.cpp
```diff
@@ -0,0 +1,111 @@
...
+      }
+#endif
+    }
+    return tomlFilePath;
+  }
+
+  std::map<std::string, std::string> parseTomlFile(
+    const boost::filesystem::path& filePath) {
+    std::map<std::string, std::string> connectionParams;
+    toml::parse_result result = toml::parse_file(filePath.c_str());
```
sfc-gh-jszczerbinski, line 75: Can you load file into to the string and support win 32 this way?

cpp/lib/TomlConfigParser.cpp
```diff
@@ -0,0 +1,111 @@
...
+    if (!result)
+    {
+      CXX_LOG_ERROR("Failed to parse toml file: %s. Error: %s", filePath.c_str(), result.error().description().data());
+      return connectionParams;
+    }
+    toml::table table = std::move(result).table();
```
sfc-gh-jszczerbinski, line 81: Why std::move here?

sfc-gh-jszczerbinski (changes requested):

cpp/lib/TomlConfigParser.cpp
sfc-gh-ext-simba-nl: Actually, I realized CXX tests don't get run on 32-bit debug so I can actually remove it since it won't hang anymore. Although anyone running this on 32-bit debug windows will hit an assertion if they ever tried

cpp/lib/TomlConfigParser.cpp
sfc-gh-ext-simba-nl: I tried it previously and it didn't work, because the issue originates from one of the debug crt libraries and not the toml library.

cpp/lib/TomlConfigParser.cpp
sfc-gh-ext-simba-nl: I saw it in their example and did the same. I tried it without the move and it was fine, so I'll remove it

cpp/lib/TomlConfigParser.cpp
sfc-gh-jszczerbinski: Still please remove the ifdef. 
- I think it's ok that it won't run properly on win32 debug as long as it runs well on release version.
- Having ifdef like this doesn't bring us closer to functionality working in win32 debug, but makes code harder to read and maintain

sfc-gh-jszczerbinski (approved):
Please just remove the ifdef and LGTM

cpp/lib/TomlConfigParser.cpp
```diff
@@ -0,0 +1,102 @@
...
+      return connectionParams;
+    }
+    for (auto [key, val] : *config.as_table()) {
+      connectionParams[key.data()] = val.as_string()->get();
+    }
+    return connectionParams;
+  }
+}
+
+std::map<std::string, std::string> load_toml_config()
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: The original ticket was for ODBC. There will be a part 2 to this PR where ODBC makes use of this

cpp/lib/TomlConfigParser.cpp
```diff
@@ -0,0 +1,102 @@
+#include "snowflake/TomlConfigParser.hpp"
+#include "../logger/SFLogger.hpp"
+#include "memory.h"
+
+#define TOML_EXCEPTIONS 0
+#include <toml++/toml.hpp>
+
+#undef snprintf
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-nl: Removed

cpp/lib/TomlConfigParser.cpp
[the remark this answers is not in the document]
sfc-gh-jszczerbinski: @sfc-gh-dstempniak As @sfc-gh-ext-simba-nl mentioned this implemented in order to read config in ODBC. Currently we have no request to support this feature in LibSF or PHP, but I thought it's better to have the non-driver specific part of implementation in libsnowflakeclient in case there is such a request in the future.

sfc-gh-dstempniak (approved):
LGTM

PR #841, opened 2025-03-08
SNOW-1737940: Support keepalive parameters in LibSfClient

sfc-gh-ext-simba-jy:


cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,282 @@
...
+            {
+                // gained the lock first
+                MutexGuard m_guard(m_lock);
+
+                m_connections[connection->session_id] = connection;
+                if (m_worker == NULL)
+                {
+                    this->m_master_token_validation_time = connection->master_token_validation_time;
+                    this->m_heart_beat_interval = connection->client_session_keep_alive_heartbeat_frequency;
+                    if (connection->is_heart_beat_debug_mode)
```
sfc-gh-ext-simba-hx, line 85: I don't think it's a good idea to expose debugging option to customer. Maybe try using #ifdef _DEBUG and make the test case only available with debug build. In the worst case we could make the test case manual only.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,282 @@
...
+        }
+        _mutex_unlock(&sf->mutex_heart_beat);
+    }
+
+    void stop_heart_beat_for_this_session(SF_CONNECT* sf)
+    {
+        _mutex_lock(&sf->mutex_heart_beat);
+        if (sf->is_heart_beat_on)
+        {
+            CXX_LOG_TRACE("sf::start_heart_beat_for_this_session::Add the connection to heartbeatSync list");
```
sfc-gh-ext-simba-hx, line 35: typo: start->stop

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,282 @@
...
+                        CXX_LOG_TRACE("sf::HeartbeatBackground::heartBeatAll::queue heartbeat requests...");
+                        for (std::map<std::string, SF_CONNECT*>::const_iterator itr = m_connections.begin(); itr != m_connections.end(); itr++)
+                        {
+                            HeartBeatQueue.emplace_back(itr->second);
+                        }
+                    }
+                }
+
+                // For debug purpose only sleep before sending heartbeat.
+                {
```
sfc-gh-ext-simba-hx, line 228: only do so when it's debug.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,282 @@
...
+        }
+
+        void HeartbeatBackground::removeConnection(SF_CONNECT* connection)
+        {
+            MutexGuard m_guard(m_lock);
+            m_connections.erase(connection->session_id);
+        }
+
+        void HeartbeatBackground::sendQueuedHeartBeatReq(
+            std::vector<SF_CONNECT*>& HeartBeatQueue,
```
sfc-gh-ext-simba-hx, line 117: I think we are copying the behavior from ODBC. If so please mention any different approach compare to ODBC in PR description.
Here ODBC is intentionally having heartbeatReq struct to store necessary information for heart beat request. Sending req is time consuming and not locked for performance reason, therefore it's possible the connection instance is destroyed already when sending the heart beat req.

include/snowflake/SFURL.hpp
```diff
@@ -11,6 +11,7 @@
 #include <list>
 #include "snowflake/BaseClasses.hpp"
 #include "snowflake/Proxy.hpp"
+#include "client.h"
```
sfc-gh-ext-simba-hx, line 14: why having this here?

sfc-gh-ext-simba-jy: Removed.

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: Done.

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: Fixed.

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: I changed it to ifdef _DEBUG.

sfc-gh-ext-simba-jy: "Unfortunately, I could not find a way to enable this option only in test_heartbeat, so I added a CMake condition that developers can enable manually like MOCK

include/snowflake/SFURL.hpp
```diff
@@ -422,6 +422,11 @@ class SFURL
     return m_proxy;
   }
 
+  inline static SFURL getServerURLSync(std::string& protocol, std::string& host, std::string& port)
```
sfc-gh-ext-simba-hx, line 425: As a member function of SFURL the function name sounds weird. Better to be a constructor?

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,317 @@
...
+            bool needToNotify = false;
+            {
+                // gained the lock first
+                MutexGuard m_guard(m_lock);
+
+                m_connections[connection->session_id] = connection;
+                if (m_worker == NULL)
+                {
+                    this->m_master_token_validation_time = connection->master_token_validation_time;
+                    this->m_heart_beat_interval = connection->client_session_keep_alive_heartbeat_frequency;
```
sfc-gh-ext-simba-hx, line 94: This logic seems different from ODBC. In ODBC the interval is calculated from master_token_validation_time/4. Have you checked the logic in other drivers like jdbc?

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,317 @@
...
+            sf->is_heart_beat_on = SF_BOOLEAN_FALSE;
+        }
+        else
+        {
+            CXX_LOG_TRACE("sf::HeartbeatBackrgound::startHeartBeatForThisSessionSync::Heartbeat already enabled for this session");
+        }
+        _mutex_unlock(&sf->mutex_heart_beat);
+    }
+
+    void renew_session_sync(SF_CONNECT* sf)
```
sfc-gh-ext-simba-hx, line 47: Should use existing `renew_session` function in connection.c.

tests/test_heartbeat.c
```diff
@@ -0,0 +1,145 @@
+#include "utils/test_setup.h"
+#include "../lib/snowflake_util.h"
+#include "../include/snowflake/client.h"
+#include "connection.h"
+#include "memory.h"
+
+void test_connect_with_client_session_keep_alive_disable(void** unused)
```
sfc-gh-ext-simba-hx, line 7: Please add test case for concurrency like `ConnectionLatestTest` in ODBC, to ensure no deadlock or invalid memory access in any case.

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: Yes. In ODBC, the value is fixed. However, this value can be adjustable in Node.js. Please have a look and let me know if we can use this logic instead of the fixed value. https://github.com/snowflakedb/snowflake-connector-nodejs/blob/master/lib/util.js#L471

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: According to the ODBC implementation, when the client calls renew_session, it should be locked using a recursive mutex. Since using a recursive mutex in C can be challenging, I believe we need a wrapper function for this. What do you think?

tests/test_heartbeat.cpp
```diff
@@ -0,0 +1,133 @@
...
+    }
+    protocol = getenv("SNOWFLAKE_TEST_PROTOCOL");
+    if (protocol) {
+        snowflake_set_attribute(sf, SF_CON_PROTOCOL, protocol);
+    }
+    sf_bool client_session_keep_alive = SF_BOOLEAN_TRUE;
+    snowflake_set_attribute(sf, SF_CON_CLIENT_SESSION_KEEP_ALIVE, &client_session_keep_alive);
+
+
+    snowflake_connect(sf);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I forgot to add the comment that they need to enable HEARTBEAT_DEBUG before testing the heartbeat concurrency. If HEARTBEAT_DEBUG is enabled, the heartbeat frequency is only 1 second. I also make the threads sleep for 20 seconds.

lib/connection.c
```diff
@@ -1308,4 +1309,20 @@ size_t non_json_resp_write_callback(char* ptr, size_t size, size_t nmemb, void*
 sf_bool is_password_required(AuthenticatorType auth)
 {
     return (AUTH_JWT != auth) && (AUTH_OAUTH != auth) && (AUTH_PAT != auth) && (AUTH_EXTERNALBROWSER != auth);
+}
+
+uint64 validate_client_session_keep_alive_heart_beat_frequency(int64 heart_beat_frequency)
+{
+    int64 max = SF_DEFAULT_CLIENT_SESSION_ALIVE_HEARTBEAT_FREQUENCY;
```
sfc-gh-dstempniak, line 1316: I think we shouldn't use the default value to limit the heartbeat frequency. `masterValidityInSeconds` can be changed on server like [here](https://github.com/snowflakedb/snowflake-odbc/blob/0fb98e696a479bd7b6e8ce17ac6cd6bf23cacd2e/Tests/EndToEndTests/LongRunningQueryLatestTest/LongRunningQueryLatestTest.cpp#L21). If we get 60 from the server, we will set it to 900, which seems not good.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,349 @@
...
+            {
+                // gained the lock first
+                MutexGuard m_guard(m_lock);
+
+                m_connections[connection->session_id] = connection;
+
+                if (m_worker == NULL)
+                {
+                    this->m_master_token_validation_time = connection->master_token_validation_time;
+                    this->m_heart_beat_interval = connection->client_session_keep_alive_heartbeat_frequency;
```
sfc-gh-dstempniak, line 106: I think `m_master_token_validation_time` is never used.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,349 @@
...
+                    CXX_LOG_TRACE("sf::HeartbeatBackground::addConnection:: start a new thread for heartbeatSync");
+                    m_worker = new std::thread(&HeartbeatBackground::heartBeatAll, this);
+                }
+                else
+                {
+                    if (connection->master_token_validation_time < this->m_master_token_validation_time)
```
sfc-gh-dstempniak, line 112: If we use `m_heart_beat_interval` for sleeping, we should recalculate it as well.

tests/test_heartbeat.cpp
```diff
@@ -0,0 +1,225 @@
...
+
+int main(void) {
+  initialize_test(SF_BOOLEAN_FALSE);
+  const struct CMUnitTest tests[] = {
+    //cmocka_unit_test(test_connect_with_client_session_keep_alive_disable),
+    //cmocka_unit_test(test_connect_with_client_session_keep_alive_current),
+    //cmocka_unit_test(test_token_renew),
+    cmocka_unit_test(test_heartbeat_manually),
+    //cmocka_unit_test(test_heartbeat),
+  };
```
sfc-gh-dstempniak, line 221: "manual" test enabled by default, and others disabled? What do we want to be run on CI?

sfc-gh-dstempniak (changes requested):

lib/client.c
```diff
@@ -819,6 +829,8 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
     log_debug("get_fastfail: %s", sf->get_fastfail ? "true" : "false");
     log_debug("get_maxretries: %d", sf->get_maxretries);
     log_debug("get_threshold: %d", sf->get_threshold);
+    log_debug("client_session_keep_alive: %s", sf->client_session_keep_alive ? "true" : "false");
+    log_debug("client_session_keep_alive: %d", sf->client_session_keep_alive_heartbeat_frequency);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I've updated it.

lib/connection.c
sfc-gh-ext-simba-jy: I removed this code to match the behavior of ODBC.

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: Since I removed validate_client_session_alive_heartbeat_frequency, I also removed m_heartbeat_interval and updated libsfclient to use m_master_token_validation_time to calculate the frequency.

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: The same as above

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,349 @@
...
+        void* curl_desc = get_curl_desc_from_pool(RENEW_SESSION_URL, sf->proxy, sf->no_proxy);
+        CURL* curl = get_curl_from_desc(curl_desc);
+        SF_ERROR_STRUCT* err = &sf->error;
+
+        RecursiveMutexGuard guard(*static_cast<Snowflake::Client::RecursiveMutex*>(sf->mutex_tokens));
+        if (!renew_session(curl, sf, err))
+        {
+            CXX_LOG_TRACE("sf::HeartbeatBackrgound::renew_session_sync::Failed to renew session");
+            stop_heart_beat_for_this_session(sf);
+            return SF_BOOLEAN_FALSE;
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Hello @sfc-gh-dstempniak, If you don’t mind, could you please elaborate on this? As far as I remember, the original return type of this function was void, and I only changed it to sf_bool for testing purposes. I’m not quite sure why we need to return the curl object here — could you clarify that?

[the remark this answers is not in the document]
sfc-gh-ddas:
```suggestion
            free_curl_desc(curl_desc);
            return SF_BOOLEAN_FALSE;
```

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,353 @@
...
+        HeartbeatBackground& bg = HeartbeatBackground::getInstance();
+        bg.mockHeartBeat(sf);
+    }
+
+} // extern "C"
+
+namespace Snowflake
+{
+    namespace Client
+    {
```
sfc-gh-ddas:
```suggestion
namespace Snowflake::Client {
```

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,353 @@
...
+                CXX_LOG_TRACE("sf::HeartbeatBackground::heartBeatAll::%d connections need retry with session renew", renewQueue.size());
+                freeHeartBeatReqQueue(heartBeatQueue);
+                // get lock during renew. the reason is that:
+                // 1. session renew needs more connection related implementation and it
+                //    would be too heavy to bring it here
+                // 2. hopefully there won't be too many connections need renew since
+                //    by default it's per 4 hours
+                // 3. We need to update session token in the connection which needs lock
+                //    anyway
+                MutexUnique guard(m_lock);
```
sfc-gh-ddas: I think this is not a good justification and we should lock more granularly. having a lock during the network request for multiple requests will halt everything for a while.

cpp/lib/Mutex.hpp
```diff
@@ -0,0 +1,125 @@
...
+#include <mutex>
+
+namespace Snowflake
+{
+    namespace Client
+    {
+        /**
+         * Thin wrapper on regular mutex
+         */
+        class Mutex : public std::mutex
```
sfc-gh-ddas: `std::mutex` has a non-virtual destructor and no virtual functions. This is bad practice and can lead to unpredictable behaviour. Prefer composition over inheritance.

cpp/lib/Mutex.cpp
```diff
@@ -0,0 +1,58 @@
...
+
+extern "C" {
+    void create_recursive_mutex(void** mutex, uint64_t id)
+    {
+        *mutex = (void*) new Snowflake::Client::RecursiveMutex(id);
+    }
+
+    void free_recursive_mutex(void** mutex)
+    {
+        delete static_cast<Snowflake::Client::RecursiveMutex*>(*mutex);
```
sfc-gh-ddas:
```suggestion
        delete static_cast<Snowflake::Client::RecursiveMutex*>(*mutex);
        *mutex = NULL;
```

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,353 @@
+#include "HeartbeatBackground.hpp"
+#include "../lib/heart_beat_background.h"
+#include "../lib/client_int.h"
+#include "../lib/connection.h"
+#include "../logger/SFLogger.hpp"
+#include "curl_desc_pool.h"
+#include "../include/snowflake/SFURL.hpp"
+#include "../lib/snowflake_util.h"
+
+extern "C" {
```
sfc-gh-ddas: Wrap C++ code inside extern C in try catch blocks to avoid throwing exceptions outside C++ and return C style error codes.

cpp/lib/Mutex.cpp
```diff
@@ -0,0 +1,58 @@
+#include "Mutex.hpp"
+#include "../lib/mutex.h"
+#include "../logger/SFLogger.hpp"
+
+extern "C" {
```
sfc-gh-ddas: Wrap C++ code inside extern C in try catch blocks avoid throwing exceptions outside C++ and return C style error codes.

cpp/lib/HeartbeatBackground.cpp
sfc-gh-ext-simba-jy: Done.

cpp/lib/HeartbeatBackground.hpp
```diff
@@ -0,0 +1,114 @@
...
+
+namespace Snowflake
+{
+    namespace Client
+    {
+        typedef struct _heartbeatreq_s
+        {
+            _heartbeatreq_s(SF_CONNECT* sf, const std::string& url,
+                SF_HEADER* header)
+                : sessionId(sf->session_id), heartBeatURL(url), httpExtraHeaders(header),
```
sfc-gh-ddas: Here, a raw `SF_HEADER*` allocated by sf_header_create() is passed into the std::shared_ptr<SF_HEADER> via implicit conversion. The default shared_ptr deleter calls delete, but SF_HEADER is allocated with `sf_header_create()` and should be freed with `sf_header_destroy()`. `delete` will leak all the HTTP header strings every heartbeat cycle and the struct itself is allocated with `calloc` that won't be handled by `delete`.
```suggestion
                : sessionId(sf->session_id), heartBeatURL(url), httpExtraHeaders(header, sf_header_destroy),
```

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,360 @@
...
+            {
+                log_trace("sf::HeartbeatBackrgound::startHeartBeatForThisSessionSync::Heartbeat already enabled for this session");
+            }
+            _mutex_unlock(&sf->mutex_heart_beat);
+        }
+        catch (...)
+        {
+            log_error("sf::HeartbeatBackrgound::start_heart_beat_for_this_session::Exception occurred when starting heartbeat for this session");
+        }
+    }
```
sfc-gh-ddas: If any exception is thrown between `_mutex_lock` and `_mutex_unlock`, execution jumps to the catch block, leaving `mutex_heart_beat` permanently locked. All subsequent calls will deadlock.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,360 @@
...
+            {
+                log_trace("sf::HeartbeatBackrgound::startHeartBeatForThisSessionSync::Heartbeat already disabled for this session");
+            }
+            _mutex_unlock(&sf->mutex_heart_beat);
+        }
+        catch (...)
+        {
+            log_error("sf::HeartbeatBackrgound::stop_heart_beat_for_this_session::Exception occurred when stopping heartbeat for this session");
+        }
+    }
```
sfc-gh-ddas: Same pattern as in `start_heart_beat_for_this_session`

sfc-gh-ext-simba-jy: I moved the try catch block. Thanks.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,358 @@
...
+                    cJSON* codeItem = snowflake_cJSON_GetObjectItem(resp_data, "code");
+                    const char* code = codeItem ? codeItem->valuestring : "";
+                    if ((renewQueue) && strcmp(code, SESSION_TOKEN_EXPIRED_CODE) == 0)
+                    {
+                        CXX_LOG_TRACE("sf::HeartbeatBackground::sendQueuedHeartBeatReq::Session token expired will retry later sessionId: %s", sid.c_str());
+                        renewQueue->emplace_back(HeartBeatQueue[i]);
+                    }
+                    else
+                    {
+                        CXX_LOG_TRACE("sf::HeartbeatBackground::sendQueuedHeartBeatReq::Sending heartbeat failed with Session Id: %s and errorcode: %d", sid.c_str(), code);
```
sfc-gh-pbulawa: Implicit conversion of the `const char*code` into the integer `%d`.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,358 @@
...
+                sf->is_heart_beat_on = SF_BOOLEAN_FALSE;
+            }
+            catch (...)
+            {
+                log_error("sf::HeartbeatBackrgound::stop_heart_beat_for_this_session::Exception occurred when stopping heartbeat for this session");
+            }
+        }
+        else
+        {
+            log_trace("sf::HeartbeatBackrgound::startHeartBeatForThisSessionSync::Heartbeat already disabled for this session");
```
sfc-gh-pbulawa:
```suggestion
            log_trace("sf::HeartbeatBackground::startHeartBeatForThisSessionSync::Heartbeat already disabled for this session");
```

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,358 @@
...
+    }
+
+    void stop_heart_beat_for_this_session(SF_CONNECT* sf)
+    {
+        _mutex_lock(&sf->mutex_heart_beat);
+        if (sf->is_heart_beat_on)
+        {
+            try 
+            {
+                log_trace("sf::HeartbeatBackrgound::stop_heart_beat_for_this_session::Add the connection to heartbeatSync list");
```
sfc-gh-pbulawa:
```suggestion
                log_trace("sf::HeartbeatBackground::stop_heart_beat_for_this_session::Add the connection to heartbeatSync list");
```

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,358 @@
...
+        sf_bool ret = SF_BOOLEAN_TRUE;
+        void* curl_desc = get_curl_desc_from_pool(RENEW_SESSION_URL, sf->proxy, sf->no_proxy);
+        CURL* curl = get_curl_from_desc(curl_desc);
+        SF_ERROR_STRUCT* err = &sf->error;
+
+        RecursiveMutexGuard guard(*static_cast<Snowflake::Client::RecursiveMutex*>(sf->mutex_tokens));
+        if (!renew_session(curl, sf, err))
+        {
+            CXX_LOG_TRACE("sf::HeartbeatBackrgound::renew_session_sync::Failed to renew session");
+            stop_heart_beat_for_this_session(sf);
```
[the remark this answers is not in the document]
sfc-gh-pbulawa: There is a call chain on the same thread:

1. `HeartbeatBackground::renewSession()` acquires `MutexUnique guard(m_lock)`
2. While holding `m_lock`, it calls `renew_session_sync(itr->second)`
3. `renew_session_sync()`, if `renew_session()` fails, calls `stop_heart_beat_for_this_session(sf)`
4. `stop_heart_beat_for_this_session()` calls `bg.removeConnection(sf)`
5. `removeConnection()` tries to acquire `MutexGuard m_guard(m_lock)` - **deadlock**

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,358 @@
...
+            renewQueue.clear();
+            renewQueue.insert(renewQueue.begin(), HeartBeatQueue.begin(), HeartBeatQueue.end());
+#endif
+
+            // renew session
+            renewSession(HeartBeatQueue, renewQueue);
+        }
+    }
+
+    heartbeatReq HeartbeatBackground::genHeartBeatReq(SF_CONNECT* connection)
```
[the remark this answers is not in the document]
sfc-gh-pbulawa: `genHeartBeatReq` calls `create_header`:
https://github.com/snowflakedb/libsnowflakeclient/blob/3d1974b672732844323e3457ec4b63594f56dc7f/cpp/lib/HeartbeatBackground.cpp#L291

And `create_header` in `connection.c` does read the token without any lock:
https://github.com/snowflakedb/libsnowflakeclient/blob/3d1974b672732844323e3457ec4b63594f56dc7f/lib/connection.c#L182-L188

It reads `sf->token (or sf->master_token)`, which is exactly the kind of field protected by `sf->mutex_tokens`. Another thread performing query execution or session renewal can update these tokens concurrently.

cpp/lib/HeartbeatBackground.cpp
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Hello @sfc-gh-pbulawa, thank you for your feedback. After investigating the issue, I confirmed that this workflow can indeed cause a deadlock. Upon comparing the code flow with the ODBC implementation, I realized that stop_heartbeat is not required in renew_session_sync. Therefore, I have removed the stop_heartbeat_for_this_session call from renew_session_sync, which should resolve this issue.

tests/test_heartbeat.cpp
```diff
@@ -0,0 +1,224 @@
...
+    assert_true(sf->is_heart_beat_on);
+
+    char* previous_sessiontoken = (char*)SF_MALLOC(strlen(sf->token) + 1);
+    char* previous_masterToken = (char*)SF_MALLOC(strlen(sf->master_token) + 1);;
+    strcpy(previous_sessiontoken, sf->token);
+    strcpy(previous_masterToken, sf->master_token);
+    test_heartbeat(sf);
+
+    assert_false(sf_strncasecmp(previous_sessiontoken, sf->token, strlen(sf->token)) == 0);
+    assert_false(sf_strncasecmp(previous_masterToken, sf->master_token, strlen(sf->token)) == 0);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Done.

cpp/lib/Mutex.cpp
```diff
@@ -0,0 +1,77 @@
...
+    void create_recursive_mutex(void** mutex, uint64_t id)
+    {
+        try {
+            *mutex = (void*) new Snowflake::Client::RecursiveMutex(id);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: I changed the function type to sf_bool and set the validation in start_hearbeat and stop_heartbeat.

cpp/lib/HeartbeatBackground.cpp
```diff
@@ -0,0 +1,374 @@
...
+        {
+            MutexGuard m_guard(m_lock);
+            m_workerEnded = true;
+        }
+        // wake up worker thread and let it ended
+        m_cv.notify_all();
+        // join
+        if (m_worker != NULL)
+        {
+            m_worker->join();
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-jy: Done.

sfc-gh-ext-simba-jy:
This PR was created a few months ago and have updated the code with the different workflows and reviewers. I opened a new PR [here](https://github.com/snowflakedb/libsnowflakeclient/pull/999) and close this PR.

PR #843, opened 2025-03-10
Snow 1906112 - upgrade curl 8.12.1

sfc-gh-ext-simba-nl:
Updated windows build to use cmake instead of nmake since libcurl is deprecating nmake

Procedure:
1. Downloaded old version of curl (8.10.1) and created a diff
2. Removed old version and commited
3. Downloaded new version of curl (8.12.1) and commited
4. Patched new version with diff, fixed mismatches, commited changes
5. Updated build script versions for unix. Updated build script on Windows to use cmake/msbuild rather than nmake, commited changes


sfc-gh-jszczerbinski (changes requested):
Please rewrite branch history so it includes three commits:
- Remove old library
- Add new library
- Apply curl library modifications

It's hard to review the PR in the current state

PR #850, opened 2025-03-25
Implement abort (cancel) query

sfc-gh-ext-simba-nl:


include/snowflake/client.h
```diff
@@ -837,6 +838,14 @@ SF_STATUS STDCALL snowflake_propagate_error(SF_CONNECT *sf, SF_STMT *sfstmt);
 SF_STATUS STDCALL
 snowflake_query(SF_STMT *sfstmt, const char *command, size_t command_size);
 
+/**
+ * Cancels a query given the statement.
+ *
+ * @param sf SNOWFLAKE_STMT context.
+ * @return 0 if success, otherwise an errno is returned.
+ */
+SF_STATUS STDCALL snowflake_cancel_query(SF_STMT* sfstmt);
```
sfc-gh-ext-simba-hx, line 847: For async query we allow application to retrieve the result through a different connection using query id. I think better to support query cancellation with that as well. Calling `snowflake_init_async_query_result` before you can cancel the query sounds a bit inconvenient to me.

include/snowflake/client.h
```diff
@@ -643,6 +643,7 @@ typedef struct SF_TIMESTAMP {
 typedef struct SF_QUERY_METADATA {
   SF_QUERY_STATUS status;
   char *qid;
+  SF_ERROR_STRUCT error;
```
sfc-gh-ext-simba-hx, line 646: `SF_QUERY_METADATA` seems to be internal use only please move the definition to client_int.h. Sorry didn't find that in the PR for async query.
Also I don't see the usage of `error` added here. Set in `get_query_metadata` only but anywhere using it?

lib/client.c
```diff
@@ -2710,6 +2718,74 @@ SF_STATUS STDCALL snowflake_query(
...
+    clear_snowflake_error(&sfstmt->error);
+    if (strlen(sfstmt->sfqid) != 0) {
+        SF_QUERY_METADATA *metadata = get_query_metadata(sfstmt);
+        if (!is_query_still_running(metadata->status)) {
+            log_trace("Query is no longer running.");
+            return SF_STATUS_SUCCESS;
+        }
+    }
+
+    char urlbuf[sizeof(ABORT_REQUEST_URL)];
```
sfc-gh-ext-simba-hx, line 2738: What if sfqid is empty string?

lib/client.c
```diff
@@ -2710,6 +2718,74 @@ SF_STATUS STDCALL snowflake_query(
...
+    URL_KEY_VALUE url_params[] = {
+        {.key = URL_PARAM_REQUEST_ID, .value = request_id, .formatted_key = NULL, .formatted_value = NULL, .key_size = 0, .value_size = 0},
+        {.key = URL_PARAM_REQEST_GUID, .value = request_guid, .formatted_key = NULL, .formatted_value = NULL, .key_size = 0, .value_size = 0}
+    };
+
+    cJSON *body = NULL;
+    cJSON *data = NULL;
+    char *s_body = NULL;
+    char *s_resp = NULL;
+    body = create_query_json_body(sfstmt->sql_text, sfstmt->sequence_counter,
```
sfc-gh-ext-simba-hx, line 2759: This would cause extra fields in json body, might not break anything but unnecessary. ODBC only have `sqlText` and `requestId` in json body for cancel request please do the same unless you find same approach in other drivers.

lib/client.c
```diff
@@ -2710,6 +2718,74 @@ SF_STATUS STDCALL snowflake_query(
...
+    cJSON *resp = NULL;
+    if (request(sfstmt->connection, &resp, urlbuf, url_params, 0, s_body, NULL,
+        POST_REQUEST_TYPE, &sfstmt->error, SF_BOOLEAN_TRUE,
+        0, sfstmt->connection->retry_count, get_retry_timeout(sfstmt->connection),
+        NULL, NULL, NULL, SF_BOOLEAN_FALSE)) {
+        s_resp = snowflake_cJSON_Print(resp);
+        log_trace("Here is JSON response:\n%s", s_resp);
+
+        SF_FREE(s_resp);
+        SF_FREE(s_body);
```
sfc-gh-ext-simba-hx, line 2771: free s_body only when http request succeeded?

lib/client.c
```diff
@@ -2710,6 +2718,74 @@ SF_STATUS STDCALL snowflake_query(
...
+    };
+
+    cJSON *body = NULL;
+    cJSON *data = NULL;
+    char *s_body = NULL;
+    char *s_resp = NULL;
+    body = create_query_json_body(sfstmt->sql_text, sfstmt->sequence_counter,
+        sfstmt->request_id, SF_BOOLEAN_FALSE, sfstmt->multi_stmt_count);
+    s_body = snowflake_cJSON_Print(body);
+    cJSON *resp = NULL;
```
sfc-gh-ext-simba-hx, line 2762: I think resp also needs to be freed.

tests/test_cancel.c
```diff
@@ -0,0 +1,258 @@
...
+  // Give time for query to init
+  sf_sleep_ms(1000);
+  _thread_init(&cancel_thread, (void *)snowflake_cancel_query, (void *)sfstmt);
+
+  _thread_join(execute_thread);
+  _thread_join(cancel_thread);
+  // Give time for query to cancel
+  sf_sleep_ms(1000);
+  SF_QUERY_STATUS query_status = snowflake_get_query_status(sfstmt);
+  assert_int_equal(query_status, SF_QUERY_STATUS_FAILED_WITH_ERROR);
```
sfc-gh-ext-simba-hx, line 35: Server returns specific error code (604) for query cancellation better to return specific error to application. See `Statement::executeInternal()` in ODBC Statement.cpp for your reference.

tests/test_cancel.c
```diff
@@ -0,0 +1,258 @@
...
+
+  snowflake_stmt_term(sfstmt);
+  snowflake_term(sf);
+}
+
+void test_bind_params() {
+  int array_size = 100000;
```
sfc-gh-ext-simba-hx, line 154: Great to have this test case stage binding is quite tricky for query cancellation as there is time consuming stage on driver side (binding data uploading). In that stage nothing could be cancelled on server side so we might need to check the cancel flag and stop the process. I don't see such logic though maybe the cancellation is not triggered on the right timing. Please debug into it and make sure the test covers all cases.

tests/test_cancel.c
```diff
@@ -0,0 +1,258 @@
...
+int main(void) {
+    initialize_test(SF_BOOLEAN_FALSE);
+    const struct CMUnitTest tests[] = {
+      cmocka_unit_test(test_basic_cancel),
+      cmocka_unit_test(test_no_stmt),
+      cmocka_unit_test(test_no_query),
+      cmocka_unit_test(test_async),
+      cmocka_unit_test(test_finished_query),
+      cmocka_unit_test(test_multiple_statements),
+      cmocka_unit_test(test_bind_params),
```
sfc-gh-ext-simba-hx, line 253: Please add one more test case for parameter binding, batch execution, see `test_array_binding_supported_false_*` (one case would be enough for cancellation) in test_bind_params.c for your reference. In such case the driver execute the query in a loop for each parameter set there is nothing to cancel on server side if the timing is just between two execution. Checking a cancel flag would work we need such logic to be added and test case to cover it.

lib/client.c
sfc-gh-ext-simba-nl: sfqid isn't used to cancel so it should be fine

include/snowflake/client.h
sfc-gh-ext-simba-nl: I thought I was going to use it but didn't. Removed.

sfc-gh-jszczerbinski (approved):

lib/client.c
```diff
@@ -2710,6 +2722,86 @@ SF_STATUS STDCALL snowflake_query(
     return SF_STATUS_SUCCESS;
 }
 
+SF_STATUS STDCALL snowflake_cancel_query(SF_STMT *sfstmt) {
+    if (!sfstmt) {
+        return SF_STATUS_ERROR_STATEMENT_NOT_EXIST;
+    }
+    if (!sfstmt->sql_text) {
```
sfc-gh-dstempniak, line 2729: Shouldn't we check `sfstmt->request_id` instead? A statement can be prepared without being executed, in such case we shouldn't try to cancel it.

lib/client.c
```diff
@@ -2777,18 +2779,39 @@ SF_STATUS STDCALL snowflake_cancel_query(SF_STMT *sfstmt) {
...
+            "No message field in JSON response", SF_SQLSTATE_GENERAL_ERROR, sfstmt->sfqid);
+        }
+        if (!(data = snowflake_cJSON_GetObjectItem(resp, "data"))) {
+          log_error("Missing data field in response");
+          SET_SNOWFLAKE_STMT_ERROR(&sfstmt->error, SF_STATUS_ERROR_BAD_JSON,
+            "No data object in JSON response", SF_SQLSTATE_GENERAL_ERROR, sfstmt->sfqid);
+        }
         if (!snowflake_cJSON_IsNull(data)) {
             cJSON *sql_state = snowflake_cJSON_GetObjectItem(data, "sqlState");
             SET_SNOWFLAKE_STMT_ERROR(&sfstmt->error, (int64)strtol(code->valuestring, NULL, 10),
```
sfc-gh-dstempniak, line 2817: `SET_SNOWFLAKE_STMT_ERROR` doesn't throw an exception, so In an extreme case `code` and `msg` can be NULL here.

sfc-gh-dstempniak (approved):

PR #853, opened 2025-03-25
SNOW-1906112 upgrade to curl 8.12.1

sfc-gh-ext-simba-nl:
Downloaded old version of curl (8.10.1) and created a diff
Removed old version and commited
Downloaded new version of curl (8.12.1) and commited
Patched new version with diff, fixed mismatches, commited changes
Updated build script versions for unix. Updated build script on Windows to use cmake/msbuild rather than nmake, commited changes


sfc-gh-jszczerbinski (approved):

sfc-gh-dstempniak:
What about `build_dependencies.bat `and `build_dependencies.sh`? Don't we want to use it and test it in libsnowflakeclient?

sfc-gh-ext-simba-nl:
Hi @sfc-gh-dstempniak 
I don't think build_dependencies.* need to be updated with a thirdparty upgrade, only if a new library has been added. They will still run build_curl.bat/sh with the same arguments

sfc-gh-pbulawa (approved):

sfc-gh-dstempniak:
@sfc-gh-ext-simba-nl You are right, I confused it with odbc. There are errors, I'll attach logs to the ticket.

PR #866, opened 2025-04-02
Revert "SNOW-1906112 upgrade to curl 8.12.1 (#853)"

sfc-gh-ext-simba-nl:
This reverts commit c2f6b168b8b84c5b7a112d5def626f1b2102f43e.


sfc-gh-jszczerbinski (commented):
I think there is no reason to revert curl upgrade, the issues on main were caused by windows runner upgrade (CMake got upgraded to 4.0.0) which dropped support for earlier CMake versions (2.8 ... 3.5) and caused build failures in aws sdk

PR #869, opened 2025-04-04
change hardcoded thread number to session parameters

sfc-gh-ext-simba-jy:


sfc-gh-ext-simba-jy:
Close this PR as we do not need this anymore

PR #878, opened 2025-04-17
SNOW-1737939 unicode support POC

sfc-gh-ext-simba-hx:
This is POC only never merge.


sfc-gh-ext-simba-hx:
Not needed anymore. Close

PR #885, opened 2025-05-07
SNOW-2032704 ODBC/libsfclient driver support virtual-style domains

sfc-gh-ext-simba-jy:
- [x] Added the virtual style end points
- [x] Added unit testing.


sfc-gh-ext-simba-hx:
@sfc-gh-ext-simba-jy missing changes in `StatementPutGet.cpp` ?

cpp/SnowflakeGCSClient.cpp
```diff
@@ -59,6 +59,14 @@ SnowflakeGCSClient::SnowflakeGCSClient(StageInfo *stageInfo, unsigned int parall
   {
     m_stageEndpoint = m_stageInfo->endPoint;
   }
+  else if (m_stageInfo->useVirtualUrl)
+  {
+      std::string bucket;
+      std::string object;
+      extractBucketAndObject(m_stageInfo->location, bucket, object);
+      m_stageEndpoint = bucket + ".storage.googleapis.com";
```
sfc-gh-ext-simba-hx, line 67: we could use GCS_ENDPOINT here

cpp/SnowflakeGCSClient.hpp
```diff
@@ -110,6 +110,8 @@ class SnowflakeGCSClient : public Snowflake::Client::IStorageClient
   std::string m_gcsAccessToken;
 
   std::string m_stageEndpoint;
+
+  bool m_useVirtualUrl = false;
```
sfc-gh-ext-simba-hx, line 114: Seems unnecessary as we have this in StageInfo already.

cpp/SnowflakeGCSClient.cpp
sfc-gh-ext-simba-jy: Done!

cpp/SnowflakeGCSClient.hpp
sfc-gh-ext-simba-jy: Done!

cpp/StatementPutGet.cpp
```diff
@@ -126,12 +126,11 @@ bool StatementPutGet::parsePutGetCommand(std::string *sql,
     putGetParseResponse->stageInfo.credentials = {
             {"AZURE_SAS_KEY",     response->stage_info->stage_cred->azure_sas_token}
       };
-    putGetParseResponse->stageInfo.endPoint = response->stage_info->endPoint;
```
sfc-gh-ext-simba-jy, line 129: This is a duplicated code: https://github.com/snowflakedb/libsnowflakeclient/pull/885/files#diff-be11d2d18afd4d1aab1ce41131170ce611ab3915a7ef1b887ca1978b78f8759eL110

cpp/SnowflakeGCSClient.cpp
```diff
@@ -299,7 +306,10 @@ void SnowflakeGCSClient::buildGcsRequest(const std::string& filePathFull,
   object = encodeUrlName(object);
 
   // https://storage.googleapis.com//BUCKET_NAME/OBJECT_NAME
-  url = "https://" + m_stageEndpoint + "/" + bucket + "/" + object;
+  // When use the virtual endPoint: https://BUCKET_NAME.storage.googleapis.com/OBJECT_NAME
+  std::string queryString = (!(m_stageInfo->endPoint.empty() && m_stageInfo->useVirtualUrl) ? bucket + "/" : "") + object;
```
sfc-gh-ext-simba-hx, line 310: So this means use virtual URL only when useVirtualUrl=true and m_stageInfo->endPoint is empty, right?
I think following could be easier to understand.
std::string queryString = ((m_stageInfo->endPoint.empty() && m_stageInfo->useVirtualUrl) ? "" :  bucket + "/" ) + object;

sfc-gh-ext-simba-jy: Done!

cpp/SnowflakeGCSClient.cpp
```diff
@@ -299,7 +306,10 @@ void SnowflakeGCSClient::buildGcsRequest(const std::string& filePathFull,
   object = encodeUrlName(object);
 
   // https://storage.googleapis.com//BUCKET_NAME/OBJECT_NAME
-  url = "https://" + m_stageEndpoint + "/" + bucket + "/" + object;
+  // When use the virtual endPoint: https://BUCKET_NAME.storage.googleapis.com/OBJECT_NAME
+  std::string queryString = ((m_stageInfo->endPoint.empty() && m_stageInfo->useVirtualUrl) ? "" : bucket + "/") + object;
```
sfc-gh-dstempniak, line 310: I think `m_stageInfo->endPoint.empty()` should be removed. If `useVirtualUrl` is true, we shouldn't add `bucket` to the URL. Take a look at [these tests](https://github.com/snowflakedb/snowflake-connector-python/blob/fe9547b1674be1bf646615f20908693bcccf61a4/test/unit/test_gcs_client.py#L424). If `use_virtual_url` is true, there is no `location` in the URL regardless `endpoint` is specified.

tests/test_unit_put_get_gcs.cpp
```diff
@@ -277,22 +288,38 @@ void test_simple_put_gcs_with_presignedurl(void ** unused)
...
 {
-  gcs_regional_url_test_core(false, "testregion", "testendpoint.googleapis.com", "testendpoint.googleapis.com");
+  SF_UNUSED(unused);
+  gcs_regional_url_test_core(false, "testregion", "testendpoint.googleapis.com", "testendpoint.googleapis.com", false);
+}
+
+void test_gcs_all_endpoint_fields_enabled(void** unused)
+{
+  SF_UNUSED(unused);
+  gcs_regional_url_test_core(false, "testregion", "testendpoint.googleapis.com", "testendpoint.googleapis.com", true);
```
sfc-gh-dstempniak, line 316: Can you add verification of the whole URL? Something like [here](https://github.com/snowflakedb/snowflake-connector-net/blob/1f4aab31a0532bece2c30ec8fb4819a07c5d8c79/Snowflake.Data.Tests/UnitTests/SFGCSClientTest.cs#L375)

cpp/SnowflakeGCSClient.cpp
sfc-gh-ext-simba-jy: Hello @sfc-gh-dstempniak, thank you for pointing this out. 

In my opinion, the Python driver may have missed this condition [here](https://github.com/snowflakedb/snowflake-connector-python/blob/fe9547b1674be1bf646615f20908693bcccf61a4/src/snowflake/connector/gcs_storage_client.py#L471).

Based on the get_location function [here](https://github.com/snowflakedb/snowflake-connector-python/blob/fe9547b1674be1bf646615f20908693bcccf61a4/src/snowflake/connector/gcs_storage_client.py#L438), the driver currently checks the following three conditions:

1. endPoint is not empty

2. use_virtual_url is True

3. use_regional_url is True

As far as I understand, when the endPoint is specified, we should include the bucket name in the query string. The reason why not including the bucket name in the query string when the virtual endPoint  == true is that the bucket name is already included at the domain like 'https://location.storage..."). Here is the Google document [here](https://cloud.google.com/storage/docs/request-endpoints#xml-api). That’s the reason I submitted the second PR for both the Node.js [here](https://github.com/snowflakedb/snowflake-connector-nodejs/pull/1065) and Go drivers [here](https://github.com/snowflakedb/gosnowflake/pull/1402). I believe the Python driver needs to be updated accordingly. 

What do you think @sfc-gh-pbulawa ? If this condition is not actually required, please let me know. I’ll be happy to remove it from this PR as well as the ones for Node.js and Go.

tests/test_unit_put_get_gcs.cpp
sfc-gh-ext-simba-jy: I think this testing already did it. At https://github.com/snowflakedb/libsnowflakeclient/pull/885/files#diff-f0d50d0640fcef879eee03b7fc86e83a1cb16ea9992cd670efe1397c933713d4R112, this validates that the URL should be the same as the expected url

cpp/SnowflakeGCSClient.cpp
sfc-gh-pbulawa: Hi @sfc-gh-ext-simba-jy,

If the endpoint is provided, we should use it as a host for the query. Based on the `use_virtual_url`, driver can define if the location path should be included in the rest of the query or not.

The location that we want to achieve may or may not be a virtual-style endpoint. Thanks to the use of `use_virtual_url,` the driver can be flexible, and if the use of the path-style endpoint is required, it can force such behavior, e.g.
`{ endPoint: "https://overriddenurl.com", use_virtual_url: true }` -> `https://overrideenurl.com/filename`
`{ endPoint: "https://overriddenurl.com", use_virtual_url: false }` -> `https://overrideenurl.com/location/filename`

sfc-gh-ext-simba-jy: I see. I updated the PR and made the PRs for this for both Go and Node.js

sfc-gh-pbulawa (approved):

sfc-gh-dstempniak (approved):
LGTM

PR #886, opened 2025-05-12
SNOW-1736045: update cJSON 1.7.18

sfc-gh-ext-simba-hx:
teamwork issue 1134


sfc-gh-jszczerbinski (changes requested):
Please download the cJSON in build script as dependency and remove the copied files

sfc-gh-dstempniak (approved):

sfc-gh-dstempniak:
Hi @sfc-gh-ext-simba-hx , I think it's not a top priority, let's do the integration in a separate ticket. 
@sfc-gh-jszczerbinski please create a ticket in case you want it.

sfc-gh-ext-simba-hx:
> Please download the cJSON in build script as dependency and remove the copied files

Hi @sfc-gh-jszczerbinski I've made changes to download the cJSON in build script, removed the copied files and use patch to generate them from the downloaded one. Please take a look if that looks good I might apply the same approach when updating other dependencies later (such as aws sdk).

PR #900, opened 2025-06-17
SNOW-2041981: add curl header customization hook in aws sdk

sfc-gh-ext-simba-hx:
teamwork issue 1273
add header customization hook for kerberos proxy auth support.
Draft for now, will be ready for review when implementation completed in https://github.com/snowflakedb/libsnowflakeclient/tree/SNOW-2041981-header-customization-hook


sfc-gh-ext-simba-hx:
Ticket closed we don't need this anymore.

PR #902, opened 2025-06-26
SNOW-2133572: Permission denied error for "/root/sf_client_config.json"

sfc-gh-ext-simba-gc:
teamwork issue #1276

The exception that the customer reported was triggered by lack of read permission on home folder, which lead to crashing the program.

Code changes:
Implemented try-catch block to catch the exception generated from client config file so that the program can continue running without interruption


cpp/lib/client_config_parser.cpp
```diff
@@ -238,13 +243,24 @@ namespace
     const boost::filesystem::path& configFilePath,
     client_config& clientConfig)
   {
-    boost::filesystem::path derivedConfigPath = resolveClientConfigPath(configFilePath);
-
-    if (!derivedConfigPath.empty())
-    {
-      return parseConfigFile(derivedConfigPath, clientConfig);
+    try {
+      boost::filesystem::path derivedConfigPath = resolveClientConfigPath(configFilePath);
```
sfc-gh-ext-simba-hx, line 247: There might be something we could do for automated testing.
Set `HOME` environment variable to `/root` before running the test might be able to trigger the permission exception. Ignore the test if the current user is root already.

cpp/lib/client_config_parser.cpp
```diff
@@ -238,13 +243,24 @@ namespace
...
+    } catch (const std::ios_base::failure &e) {
+      // catch exception from fstream
+      CXX_LOG_ERROR("file operation exception caught in loadClientConfig(): %s", e.what());
+    } catch (const std::exception &e) {
+      CXX_LOG_ERROR("Caught a general excpetion in loadClientConfig(): %s", e.what());
+    } catch (...) {
+      CXX_LOG_ERROR("Caught unknown exception in loadClientConfig()");
     }
-    return false;
+      return false;
```
sfc-gh-ext-simba-hx, line 263: neat: please remove the unnecessary indent and tailing space.

cpp/lib/client_config_parser.cpp
```diff
@@ -11,6 +11,10 @@
 #undef snprintf
 #include <boost/filesystem.hpp>
 
+#include <string>
+#include <iostream>
+#include <exception>
```
sfc-gh-ext-simba-nl, line 16: Only include them if they're necessary, ie. not already included elsewhere

sfc-gh-ext-simba-nl: I don't think we need exception to build, it should already be included from snowflake/Exceptions.hpp

tests/test_unit_logger.c
```diff
@@ -3,6 +3,10 @@
 #include "memory.h"
 #include <stdio.h>
 
+#include <string.h>
+#include <pwd.h>        // pwd structure
```
sfc-gh-ext-simba-nl, line 7: Don't need to add comments as to why its needed

tests/test_unit_logger.c
```diff
@@ -408,6 +413,31 @@ void test_client_config_stdout() {
   remove(configFilePath);
 }
 
+void test_log_creation_no_permission_to_home_folder(){
+  // check if current user is root. If so, exit test
+  char *name;
+  struct passwd *pwd;
+  pwd = getpwuid(getuid());
+  name = pwd->pw_name;
+  assert_int_not_equal(strcmp(name, "root"), 0);
```
sfc-gh-ext-simba-nl, line 422: Does this mean if it's root, the test will fail? We'd want it to still pass, so the test should just skip instead

tests/test_unit_logger.c
```diff
@@ -567,6 +597,10 @@ int main(void) {
         cmocka_unit_test(test_client_config_log_no_level),
         cmocka_unit_test(test_client_config_log_no_path),
         cmocka_unit_test(test_client_config_stdout),
+#endif
+#if !defined(_WIN32) || !defined(_WIN64)
```
sfc-gh-ext-simba-nl, line 601: There's already an #ifndef _WIN32 below which is essentially the same as what you have here, so you can just reuse it

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+ */
+void test_log_creation_no_permission_to_home_folder(){
+ 
+  // check if current user is root. If so, exit test
+  char *name;
+  struct passwd *pwd;
+  pwd = getpwuid(getuid());
+  name = pwd->pw_name;
+
+  if(strcmp(name, "root") != 0){
```
sfc-gh-ext-simba-hx, line 452: return directly if it's root user. That would be better than having a long if clause.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    struct stat fileBuffer;
+    int rc;
+
+    if(stat("/tmp/SF_client_config_folder/sf_client_config.json", &fileBuffer) != 0){
```
sfc-gh-ext-simba-hx, line 456: creating json file is unnecessary. The purpose of this test case is checking no exception thrown when no permission on the folder. It doesn't matter whether having json file in the folder, no permission anyway.

sfc-gh-ext-simba-hx: Please use existing function `sf_create_directory_if_not_exists` to create the temporary folder, that would be a lot easier.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+      
+      // /tmp/SF_Client_config folder exists and needs to be cleaned up
+      chmod("/tmp/SF_client_config_folder", 777);
+      rc = remove("/tmp/SF_client_config_folder/sf_client_config.json");
+
+      if(rc == 0){
+        // successful in removing sf_client_config.json
+        // removing SF_client_config_folder
+        rc = remove("/tmp/SF_client_config_folder/");
```
sfc-gh-ext-simba-hx, line 465: Please use existing function `sf_create_directory_if_not_exists` to create the temporary folder, that would be a lot easier.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+      }
+
+      if(rc != 0){
+        // not successful in removing sf_client_config_folder
+        log_debug("Not successful in ensuring /tmp/sf_client_config_folder does not exist");
+      }
+    }    
+
+    // creating SF_client_config_folder
+    rc = mkdir("/tmp/SF_client_config_folder", S_IRWXU | S_IRGRP | S_IROTH | S_IXOTH);
```
sfc-gh-ext-simba-hx, line 477: Use `sf_create_directory_if_not_exists` to create the folder, then you could use `chmod` to set the proper permission afterwards.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    if(rc != 0){
+      log_debug("NOT successful in creating sf client config folder inside of tmp folder");
+    }
+
+    // obtaining a record of tmp folder permission
+    struct stat folder_stat;
+    mode_t folderOrigPerm = 0755;
```
sfc-gh-ext-simba-hx, line 484: Why keeping the original permission while the temporary folder will be deleted at the end anyway?

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    if(stat("/tmp/SF_client_config_folder", &folder_stat) == 0){
+      folderOrigPerm = folder_stat.st_mode & 0777;
+    } else {
+      log_debug("Error: unable to get /tmp/sf_client_config folder permission");
+    }
+    
+    char configFilePath[] = "/tmp/SF_client_config_folder/sf_client_config.json";
```
sfc-gh-ext-simba-hx, line 491: again, json file is unnecessary for this case.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    char clientConfigJSON[] = "{\"common\":{\"log_level\":\"warn\",\"log_path\":\"./test/\"}}";
+    FILE *file;
+    file = fopen(configFilePath, "w");   
+    fprintf(file, "%s", clientConfigJSON);
+    fclose(file);
+
+    // setting $HOME to point at /tmp/sf_client_config_folder
+    setenv("HOME", "/tmp/SF_client_config_folder/", 1);
```
sfc-gh-ext-simba-hx, line 499: Please use macro for the path of temporary folder so it would be easier if we need to change it in the future.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    
+    // setting SF_CLIENT_CONFIG dir to be inaccessible to all
+    mode_t newPermission = 0000;
```
sfc-gh-ext-simba-hx, line 502: 0 is enough. If you want to represent flag of each permission should use 0x0000.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    // parse client config for log details - exception should be thrown here and caught
+    client_config clientConfig;
+    sf_bool result = load_client_config("", &clientConfig);
+    assert_false(result);
+
+    // changing permission of tmp folder back to what it was 
+    chmod("/tmp/SF_client_config_folder", folderOrigPerm);
+    
+    // clean up /tmp/SF_client_config_folder
+    rc = remove("/tmp/SF_client_config_folder/sf_client_config.json");
```
sfc-gh-ext-simba-hx, line 514: handling on json file is unnecessary.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+
+    if(rc == 0){
+      // Successful in removing sf_client_config.json
+      // removing SF_client_config folder
+      rc = remove("/tmp/SF_client_config_folder/");
```
sfc-gh-ext-simba-hx, line 519: I think you could use existing function `sf_delete_directory_if_exists` to delete the folder. It can delete the folder content (if there is any) as well.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    } else {
+      log_debug("Not succesfful in cleaning up sf_client_config.json");
+    }
+
+    if(rc != 0){
+      log_debug("Not succesfful in removing /tmp/sf_client_config_folder");
```
sfc-gh-ext-simba-hx, line 525: Please remove logs for debugging purpose.

tests/test_unit_logger.c
sfc-gh-ext-simba-hx, line 499: Please keep the original value of `HOME` and set it back at the end, otherwise it could break test cases afterwards if they rely on `HOME` folder to work.

tests/test_unit_logger.c
```diff
@@ -433,6 +437,96 @@ void test_log_creation() {
...
+    // setting $HOME to point at /tmp/sf_client_config_folder
+    setenv("HOME", "/tmp/SF_client_config_folder/", 1);
+    
+    // setting SF_CLIENT_CONFIG dir to be inaccessible to all
+    mode_t newPermission = 0000;
+    chmod("/tmp/SF_client_config_folder", newPermission);
+    
+    // parse client config for log details - exception should be thrown here and caught
+    client_config clientConfig;
+    sf_bool result = load_client_config("", &clientConfig);
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-hx: Sorry nvm. Didn't realize this test case is made in C, not C++.

sfc-gh-ext-simba-hx (approved):
LGTM

PR #903, opened 2025-06-26
SNOW-2111927: fix thread handle leak from event loop

sfc-gh-ext-simba-hx:
teamwork issue 1265
The handle leak is from event loop having a duplicated thread handle opened in thread itself.

### DIFF
--- a/crt/aws-crt-cpp/crt/aws-c-io/source/event_loop.c
+++ b/crt/aws-crt-cpp/crt/aws-c-io/source/event_loop.c
@@ -120,6 +120,8 @@ static struct aws_event_loop_group *s_event_loop_group_new(
         /* Don't pin to hyper-threads if a user cared enough to specify a NUMA node */
         if (!pin_threads || (i < group_cpu_count && !usable_cpus[i].suspected_hyper_thread)) {
             struct aws_thread_options thread_options = *aws_default_thread_options();
+            // SNOW-2111927 custom changes to keep, fixing thread handle leak from event loop.
+            thread_options.join_strategy = AWS_TJS_MANAGED;
 
             struct aws_event_loop_options options = {
                 .clock = clock,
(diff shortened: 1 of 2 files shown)


crt/aws-crt-cpp/crt/aws-c-io/source/event_loop.c
```diff
@@ -120,6 +120,8 @@ static struct aws_event_loop_group *s_event_loop_group_new(
         /* Don't pin to hyper-threads if a user cared enough to specify a NUMA node */
         if (!pin_threads || (i < group_cpu_count && !usable_cpus[i].suspected_hyper_thread)) {
             struct aws_thread_options thread_options = *aws_default_thread_options();
+            /* SNOW-2111927 custom changes to keep, fixing thread handle leak from event loop. */
+            thread_options.join_strategy = AWS_TJS_MANAGED;
```
sfc-gh-dstempniak, line 124: Are you sure this won't break anything?
In the sources I see:
```
 * Currently, only event loop group async cleanup and host resolver threads participate in the managed thread system.
 * Additionally, event loop threads will increment and decrement the pending join count (they are manually joined
 * internally) in order to have an accurate view of internal thread usage and also to prevent failure to release
 * an event loop group fully from allowing aws_thread_join_all_managed() from running to completion when its
 * intent is such that it should block instead.
```
Or maybe it should be reported to AWS?

sfc-gh-ext-simba-hx: The join strategy is not the root cause. reverted that change and made fix on the actual one.

sfc-gh-ext-simba-hx:
Hi @sfc-gh-dstempniak @sfc-gh-jszczerbinski 
I'm not authorized to push to branch deps/aws-sdk-1.11.283 so can't merge this PR.
Please go ahead to merge it once you think it's all good.
On the other hand, in the [PR](https://github.com/snowflakedb/libsnowflakeclient/pull/890) for updating aws sdk I'll download the source from original repo and apply a patch which would be made from all custom changes in deps/aws-sdk-1.11.283 branch.

sfc-gh-ext-simba-hx:
Close this as the change has been merged with aws sdk update https://github.com/snowflakedb/libsnowflakeclient/pull/890

PR #904, opened 2025-06-26
SNOW-2111927: fix thread handle leak in from aws sdk

sfc-gh-ext-simba-hx:
teamwork issue 1165
Combine with https://github.com/snowflakedb/libsnowflakeclient/pull/903 fix the handle leak from aws sdk.

Manually tested with [handle](https://learn.microsoft.com/en-us/sysinternals/downloads/handle) after re-initialize aws sdk multiple times no handle leak found.

Ready for review but draft for now will update aws sdk build scripts to use the changes in https://github.com/snowflakedb/libsnowflakeclient/pull/903 once it's merged.

### DIFF
--- a/cpp/AWSUtils.cpp
+++ b/cpp/AWSUtils.cpp
@@ -17,6 +17,16 @@ namespace Snowflake {
       public:
         AwsSdkInitialized() : options{} {
           CXX_LOG_INFO("Initializing AWS SDK");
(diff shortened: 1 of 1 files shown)


sfc-gh-ext-simba-hx:
Close as actual root cause found and made fix in #903. The workaround here is not needed anymore.

PR #909, opened 2025-07-17
SNOW-1017608 Improve logging in libsnowflakeclient

sfc-gh-ext-simba-nl:
Teamwork issue 846
Update logging based on design doc to align with other connectors

### DIFF
--- a/cpp/FileMetadataInitializer.cpp
+++ b/cpp/FileMetadataInitializer.cpp
@@ -163,27 +163,27 @@ void Snowflake::Client::FileMetadataInitializer::initCompressionMetadata(
                   sizeof(COMPRESSION_AUTO)))
   {
     // guess
-    CXX_LOG_INFO("Auto detect on compression type");
+    CXX_LOG_DEBUG("Auto detect on compression type");
     fileMetadata.sourceCompression = FileCompressionType::guessCompressionType(
       fileMetadata.srcFileName);
   }
   else if (!sf_strncasecmp(m_sourceCompression, COMPRESSION_NONE, 
                         sizeof(COMPRESSION_NONE)))
   {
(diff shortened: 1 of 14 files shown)


sfc-gh-pbulawa (changes requested):

sfc-gh-pbulawa (approved):

tests/test_unit_logger.c
```diff
@@ -438,6 +439,32 @@ void test_log_creation() {
...
+
+  log_set_lock(NULL);
+  log_set_level(SF_LOG_WARN);
+  log_set_quiet(1);
+  log_set_path(logname);
+
+  log_file_usage(non_existent_file, "TestContext", true);
+
+  // warning log will trigger the log file creation
+  log_warn("dummy warning log");
```
sfc-gh-dstempniak, line 461: What is being tested here? A warning should be created by `log_file_usage`.

cpp/util/LogFileUtil.cpp
sfc-gh-dstempniak, line 1: It is rather a C file, can we rename it and move to lib folder?

cpp/util/Stopwatch.cpp
sfc-gh-dstempniak, line 1: Same here. Do we need to expose Snowflake.h outside the library? Can we move to lib as well?

tests/test_unit_logger.c
sfc-gh-ext-simba-nl: Updated test case

cpp/util/LogFileUtil.cpp
sfc-gh-ext-simba-nl: Sure, I'll use stat instead of boost then

sfc-gh-ddas (approved):

PR #910, opened 2025-07-23
SNOW-1528817: deprecate region parameter

sfc-gh-ext-simba-hx:
teamwork issue 1110
mark SF_CON_REGION as deprecated in both header file and log.


lib/client.c
```diff
@@ -713,6 +724,21 @@ _snowflake_check_connection_parameters(SF_CONNECT *sf) {
         alloc_buffer_and_copy(&sf->host, buf);
     }
 
+    // continue on split account removing exteral ID
+    if (dot_ptr) {
```
sfc-gh-pbulawa, line 728: If the comparison at L:696 succeeds, then this will never happen, as the `dot_ptr` will be cleaned. Seems that is not expected?

sfc-gh-pbulawa (approved):

sfc-gh-dstempniak (approved):

PR #918, opened 2025-08-22
SNOW-2295402: Fix NO_PROXY character limit

sfc-gh-pbulawa:
SNOW-2285315: Fix NO_PROXY character limit


cpp/util/Proxy.cpp
```diff
@@ -78,7 +78,7 @@ void Snowflake::Client::Util::Proxy::clearPwd() {
 }
 
 void Snowflake::Client::Util::Proxy::setProxyFromEnv() {
-    char valbuf[1024];
+    char valbuf[4096];
```
sfc-gh-snoonan, line 81: I think there's a max of 32k for the environment on most versions of Windows, which is where we see this. Any reason not to just go to that value?

PR #928, opened 2025-09-16
NO-SNOW: Refresh build caches by running actions each week

sfc-gh-jszczerbinski:
Build caches get wiped after a week of not being used. When development is slow the caches can get wiped - this causes:
- slow builds
- code quality fails since Extra Warnings compares its result with master


sfc-gh-akolodziejczyk (approved):
LGTM

PR #995, opened 2026-03-16
SNOW-3121512: update openssl to 3.5.7

sfc-gh-ext-simba-hx:


sfc-gh-pbulawa (changes requested):
Shouldn't we rebuild curl and dependencies?

sfc-gh-pbulawa (approved):

PR #1014, opened 2026-05-29
SNOW-3411977: AWS WIF outbound token

sfc-gh-pbulawa:


cpp/AwsAttestation.cpp
```diff
@@ -5,12 +5,32 @@
...
+#include <string>
 
 namespace Snowflake::Client {
+  namespace {
+    // opt-in env var that switches AWS WIF from the legacy
+    // base64(GetCallerIdentity-presigned-URL) format to a JWT obtained via
+    // STS:GetWebIdentityToken. Will be removed once the new flow is GA.
+    constexpr const char* AWS_OUTBOUND_TOKEN_ENV_VAR =
+        "SNOWFLAKE_ENABLE_AWS_WIF_OUTBOUND_TOKEN";
+    constexpr const char* SNOWFLAKE_WIF_AUDIENCE = "snowflakecomputing.com";
```
sfc-gh-pczajka, line 13: Let's use the same constant in line 154:
```
request->SetHeaderValue("X-Snowflake-Audience", SNOWFLAKE_WIF_AUDIENCE);                                                                                                                               
```

cpp/AWSUtils.cpp
```diff
@@ -80,6 +103,145 @@ namespace Snowflake {
...
+          Aws::Client::AWSAuthV4Signer signer(credProvider, "sts", region);
+          if (!signer.SignRequest(*request)) {
+            CXX_LOG_ERROR("Failed to sign STS GetWebIdentityToken request");
+            return boost::none;
+          }
+
+          Aws::Client::ClientConfiguration clientConfig;
+          clientConfig.region = region;
+          auto httpClient = Aws::Http::CreateHttpClient(clientConfig);
+          auto response = httpClient->MakeRequest(request);
```
sfc-gh-pczajka, line 210: No explicit HTTP timeout on the new STS call. getWebIdentityToken constructs Aws::Client::ClientConfiguration clientConfig; clientConfig.region = region; and otherwise relies on defaults. If STS  
  is unreachable / the network blackholes, this is the symptom users will hit as a hung connect. Setting connectTimeoutMs / requestTimeoutMs (or whatever value the rest of the driver standardizes on)  
  would make the failure mode bounded.

cpp/AWSUtils.cpp
```diff
@@ -80,6 +103,145 @@ namespace Snowflake {
...
+          if (!response) {
+            CXX_LOG_ERROR("STS GetWebIdentityToken: no HTTP response");
+            return boost::none;
+          }
+
+          const auto status = response->GetResponseCode();
+          // Drain the response body once into a string so we can both inspect
+          // it on failure and parse it on success.
+          std::ostringstream bodyBuf;
+          bodyBuf << response->GetResponseBody().rdbuf();
```
sfc-gh-pczajka: NIT (feel free to resolve):
Response body is materialized unbounded.                                                                                                                                                       `truncateForLog` caps the log line, not the buffer itself. STS responses are small in practice, so this is defense-in-depth, but adding a size cap would close the door on a misbehaving intermediary returning a large body.

sfc-gh-pczajka (approved):
The logic matches Python implementation. Left a few comments though

cpp/AwsAttestation.cpp
sfc-gh-pbulawa: Done

cpp/AWSUtils.cpp
sfc-gh-pbulawa: Done

cpp/AWSUtils.cpp
sfc-gh-pbulawa: Done

PR #1017, opened 2026-06-11
add toml connection support

sfc-gh-ext-simba-jy:
- [x] Added Support for the toml file connection in libsnowflakeclient.
- [x] Implemented the SF_ENV_SKIP_TOKEN_FILE_PERM_VERIFICATION overrides ENV_SKIP_WARNING_FOR_READ_PERM


lib/connection.c
```diff
@@ -1409,4 +1410,421 @@ sf_bool is_secure_storage_auth(AuthenticatorType auth)
...
+    }
+    else if (sf_strncasecmp(key, "WAREHOUSE", strlen(key)) == 0)
+    {
+        snowflake_set_attribute(sf, SF_CON_WAREHOUSE, value);
+    }
+    else if (sf_strncasecmp(key, "ROLE", strlen(key)) == 0)
+    {
+        snowflake_set_attribute(sf, SF_CON_ROLE, value);
+    }
+    else if (sf_strncasecmp(key, "HOST", strlen(key)) == 0 || sf_strncasecmp(key, "SERVER", strlen(key) == 0))
```
sfc-gh-makowalski: Misplaced paren: the `== 0` is *inside* `strlen()`'s argument on the `SERVER` clause, so this calls `sf_strncasecmp(key, "SERVER", 0)` (compares 0 chars → returns 0 → falsy) and never compares the real result. `SERVER` will never map to `SF_CON_HOST`. Should be `sf_strncasecmp(key, "SERVER", strlen(key)) == 0`.

lib/connection.c
```diff
@@ -1409,4 +1410,421 @@ sf_bool is_secure_storage_auth(AuthenticatorType auth)
...
-}
\ No newline at end of file
+}
+
+void handle_single_param(SF_CONNECT* sf, const char* key,const char* value)
+{
+    sf_bool v = SF_BOOLEAN_FALSE;
+    int64 i = 0;
+    int8 c = 0;
+    if (sf_strncasecmp(key, "ACCOUNT", strlen(key)) == 0)
```
sfc-gh-makowalski: This whole key table uses `sf_strncasecmp(key, "TARGET", strlen(key))` — length = the *input* key length — which makes every comparison a **prefix** match: an abbreviated key silently binds to the first target it prefixes (e.g. `PASS` → `PASSWORD`, a 1-char `A` → `ACCOUNT`). Full canonical keys mostly work because the NUL terminator forces a mismatch on longer targets, but it is fragile. Recommend an exact comparison across this function (e.g. `strcasecmp(key, "TARGET") == 0`, or a `strlen` equality check).

cpp/util/SnowflakeCommon.cpp
```diff
@@ -217,7 +220,70 @@ void append_spcs_token(cJSON* data, const char* spcs_token_path)
...
+    {
+        return;
+    }
+
+    if (dsn.back() == ';')
+    {
+        dsn.pop_back();
+    }
+
+    if (sf == NULL)
```
sfc-gh-makowalski, line 267: `sf = snowflake_init();` here reassigns the *local* parameter, so the caller never receives the new handle and the freshly-init'd `SF_CONNECT` leaks. Either take `SF_CONNECT**`, or drop this dead branch and require a non-NULL `sf` (the only caller, `snowflake_load_toml_config`, already passes an initialized handle).

cpp/util/SnowflakeCommon.cpp
```diff
@@ -217,7 +220,70 @@ void append_spcs_token(cJSON* data, const char* spcs_token_path)
...
+
+    SF_CONNECT* sf = snowflake_init();
+    snowflake_parse_dsn(sf, dsn);
+    return sf;
+}
+
+char* STDCALL snowflake_load_toml_as_dsn()
+{
+    std::string dsn = load_toml_config_as_dsn();
+    char* dsn_cstr = (char*)SF_MALLOC(dsn.size() + 1);
```
sfc-gh-makowalski, line 245: No NULL check after `SF_MALLOC` before the `std::memcpy` on the next line — an allocation failure would `memcpy` into NULL and crash. Separately: for an empty config this returns a malloc'd empty string, whereas `snowflake_load_toml_config` returns `NULL` for the same case — worth making the empty-config contract consistent between the two.

sfc-gh-makowalski (changes requested):

lib/connection.c
sfc-gh-pbulawa:
```suggestion
    else if (sf_strncasecmp(key, "HOST", strlen(key)) == 0 || sf_strncasecmp(key, "SERVER", strlen(key)) == 0)

```

lib/client.c
```diff
@@ -1286,6 +1286,11 @@ SF_STATUS STDCALL snowflake_term(SF_CONNECT *sf) {
 }
 
 SF_STATUS STDCALL snowflake_connect(SF_CONNECT* sf) {
+    if (sf == NULL)
+    {
+        sf = snowflake_load_toml_config();
```
sfc-gh-pbulawa: `sf` is passed by value. Assigning it inside snowflake_connect does not propagate to the caller, who passed `NULL` and holds no other reference. On success the caller can neither run queries on nor `snowflake_term the connection` → guaranteed leak and an unusable handle.

lib/connection.c
```diff
@@ -1409,4 +1410,421 @@ sf_bool is_secure_storage_auth(AuthenticatorType auth)
       default:
         return 0;
     }
-}
\ No newline at end of file
+}
+
+void handle_single_param(SF_CONNECT* sf, const char* key,const char* value)
```
sfc-gh-pbulawa, line 1420: `sf_strncasecmp(key, "USER", strlen(key))` compares only up to the input's length, so any short prefix matches — "R"→REGION, "APP"→APPLICATION, "t"/"tru"→true in parse_bool

cpp/util/SnowflakeCommon.cpp
```diff
@@ -217,7 +220,70 @@ void append_spcs_token(cJSON* data, const char* spcs_token_path)
...
+    std::memcpy(dsn_cstr, dsn.c_str(), dsn.size() + 1);
```
sfc-gh-pbulawa, line 250: No NULL check on the allocation before memcpy

cpp/util/SnowflakeCommon.cpp
```diff
@@ -217,7 +220,70 @@ void append_spcs_token(cJSON* data, const char* spcs_token_path)
...
+    return dsn_cstr;
+}
+
+void snowflake_parse_dsn(SF_CONNECT* sf, std::string& dsn)
```
sfc-gh-pbulawa: snowflake_parse_dsn mutates the caller's dsn (pop_back) as a side effect. Taking std::string& and stripping the trailing ; mutates the caller's buffer. Take it by const& and work on a copy/index, or document the mutation.

lib/connection.c
sfc-gh-pbulawa, line 1420: Let's add logging for any unknown key

sfc-gh-pbulawa (dismissed):

lib/connection.c
sfc-gh-ext-simba-jy: I refactored it due to the sf_strncasecmp issue and fixed it at that time.

lib/connection.c
sfc-gh-ext-simba-jy: I replaced sf_strncasecmp with strcasecmp and added a test case to verify the issue you raised.

cpp/util/SnowflakeCommon.cpp
sfc-gh-ext-simba-jy: I added an early return if sf is null.

cpp/util/SnowflakeCommon.cpp
sfc-gh-ext-simba-jy: I added the null check.

lib/client.c
sfc-gh-ext-simba-jy: I've posted my thoughts on your question here. Please take a look. CC @sfc-gh-makowalski : https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/1395#issuecomment-4896862374

sfc-gh-makowalski: Bringing the recommendation from the linked issue (#1395) back onto the PR:

I'd go with **option 2** — the separate `snowflake_connect_with_toml()` that returns `SF_CONNECT*`, which is effectively what's already implemented here. cc @sfc-gh-ext-simba-jy @sfc-gh-pbulawa

- Returning a handle is idiomatic here: `snowflake_init()` already returns `SF_CONNECT*`, so the caller owns and `snowflake_term`s the returned handle exactly as today — no new ownership model.
- It leaves `snowflake_connect(SF_CONNECT*)` untouched — no extra field, no file/env I/O as a side effect of the core connect, and no precedence question between explicitly-set attributes and TOML values. Option 1 would push all three onto the primary connect path.
- It's the closest C analog to the Go/Node "connect returns the connection" design that motivated this, and it removes the by-value-doesn't-propagate problem at the root — which resolves the concern in this thread.

Two small, non-blocking tightenings on the current option-2 code:
1. `snowflake_load_toml_config()` ignores the `snowflake_parse_dsn` return — on malformed TOML it would hand back a partially-populated handle. Suggest checking it and doing `snowflake_term(sf)` + `return NULL` on failure.
2. Document in the header that the caller owns the returned handle and must `snowflake_term` it (mirroring `snowflake_init`), and that it returns `NULL` for empty/missing config.

The leak-free failure path in `snowflake_connect_with_toml` (term on connect failure) looks good.

sfc-gh-ext-simba-jy: I've updated both points. Could you please take a look? @sfc-gh-makowalski

sfc-gh-makowalski (approved):

PR #1021, opened 2026-06-23
SNOW-3684659: [Libsfc] Make WIF audience configurable

sfc-gh-ext-simba-jy:
- [x] Create connection param for the WIF audience
- [x] Update the AWS and GCP Attestation to use the audience field with a custom audience.
- [x] Add test cases for the custom audience.


sfc-gh-makowalski (commented):
Reviewed the WIF-audience change — the AWS wiring and the attribute plumbing (`SF_CON_WIF_AUDIENCE` get/set, default macro, `configureWIFAttestation` refactor) look correct. This is purely additive: the audience was hard-coded to `snowflakecomputing.com` for every provider before this PR, so nothing regresses.

A few non-blocking notes inline. The only one worth an explicit decision is whether `SF_CON_WIF_AUDIENCE` is meant to be provider-agnostic (then GCP needs wiring) or is intentionally AWS-scoped for now.

cpp/AwsAttestation.cpp
```diff
@@ -101,7 +100,7 @@ namespace Snowflake::Client {
         "Requesting AWS WIF JWT (STS:GetWebIdentityToken) in region %s",
         region.c_str());
     auto jwtOpt = config.awsSdkWrapper->getWebIdentityToken(
-        creds, region, SNOWFLAKE_WIF_AUDIENCE, AWS_WIF_SIGNING_ALGORITHM);
+        creds, region, config.audience.get(), AWS_WIF_SIGNING_ALGORITHM);
```
sfc-gh-makowalski: AWS now correctly honors `config.audience` here. Two related notes:

**GCP isn't wired to it.** `cpp/GcpAttestation.cpp` still uses its own hard-coded `SNOWFLAKE_AUDIENCE` (lines 94 and 196) and never reads `config.audience`, even though `config` is already in scope in `createGcpAttestation`. This isn't a regression — GCP was hard-coded before this PR too — but since `SF_CON_WIF_AUDIENCE` is provider-agnostic, a user who sets it will have it silently ignored on GCP. Intentionally AWS-scoped, or worth wiring GCP (one line per site: `config.audience.value_or(SF_SNOWFLAKE_WIF_AUDIENCE)`) / a follow-up?

**Nit:** `config.audience.get()` throws if the optional is ever empty. It can't be today (default member-initializer + `configureWIFAttestation` never clears it), but `.value_or(SF_SNOWFLAKE_WIF_AUDIENCE)` is cheap insurance against future construction paths.

include/snowflake/client.h
```diff
@@ -127,6 +127,11 @@ extern "C" {
 */
 #define SF_SPCS_ENV_VAR "SNOWFLAKE_RUNNING_INSIDE_SPCS"
 
+/**
+* DEFAULT WIF AUDIENCE
+*/
+#define SF_SNOWFLAKE_WIF_AUDIENCE "snowflakecomputing.com"
```
sfc-gh-makowalski, line 133: The default audience now lives in two places: this `SF_SNOWFLAKE_WIF_AUDIENCE` macro and `cpp/GcpAttestation.cpp:10`'s `constexpr SNOWFLAKE_AUDIENCE`, both `"snowflakecomputing.com"`. AWS dropped its local copy in this PR (good) — if GCP gets wired to `config.audience`, folding GCP's constant into this macro too would leave a single source of truth.

tests/test_create_wif_attestation.cpp
```diff
@@ -1032,6 +1032,50 @@ void test_unit_oidc_attestation_missing_token(void **) {
...
+
+    assert_false(config.snowflakeEntraResource.has_value());
+
+    snowflake_set_attribute(conn, SF_CON_WIF_PROVIDER, "AWS");
+    config.configureWIFAttestation(conn);
+
+    assert_true(config.type.has_value());
+    assert_int_equal(config.type.get(), AttestationType::AWS);
+
+    assert_true(config.audience.has_value());
```
sfc-gh-makowalski: This verifies `configureWIFAttestation` *populates* `config.audience`, but nothing asserts a provider actually *sends* a custom audience. The GCP tests (lines 379, 505) assert against `GCP_TEST_AUDIENCE = "snowflakecomputing.com"` (the default), so they'd pass even if a user-set audience were ignored. A case that sets a non-default `wif_audience` and asserts it reaches the AWS `getWebIdentityToken` call (and the GCP request, if wired) would lock down the actual feature.

sfc-gh-makowalski (changes requested):
Requesting changes on one item (the other two inline notes are non-blocking).

**Blocking — GCP ignores the new audience.** `SF_CON_WIF_AUDIENCE` is a provider-agnostic attribute, but `createGcpAttestation` never reads `config.audience`; both GCP audience sites (`cpp/GcpAttestation.cpp:94` and `:196`) still send the hard-coded `SNOWFLAKE_AUDIENCE`. So a user who sets the audience gets it honored on AWS but silently dropped on GCP. `config` is already in scope in `createGcpAttestation`, so this is a one-line-per-site change (`config.audience.value_or(SF_SNOWFLAKE_WIF_AUDIENCE)`).

Please either wire GCP to `config.audience`, or — if this is intentionally AWS-scoped for now — make that explicit (attribute docs + a follow-up ticket) so the generic name doesn't set a false expectation. Details in the inline comment on `cpp/AwsAttestation.cpp:103`.

Note: this is not a regression — the audience was hard-coded for all providers before this PR; the request is to finish the feature for the path it's named after.

cpp/AwsAttestation.cpp
sfc-gh-ext-simba-jy: Initially, I set the default value in the struct to SF_SNOWFLAKE_WIF_AUDIENCE, so I didn't expect the value to be empty. Taking your suggestion, instead of using a default value, I created a getAudience method in the struct that calls .value_or(SF_SNOWFLAKE_WIF_AUDIENCE) instead of config.audience.get(), so that it won't throw an exception when the value is empty, and replaced config.audience.get() with config.getAudience().

include/snowflake/client.h
sfc-gh-ext-simba-jy: I removed the macro in the GcpAttestation.cpp.

tests/test_create_wif_attestation.cpp
sfc-gh-ext-simba-jy: I've added two test cases for custom audiences for AWS and GCP. Please take a look.

lib/client.c
```diff
@@ -1904,6 +1905,9 @@ SF_STATUS STDCALL snowflake_set_attribute(
         case SF_CON_WIF_AZURE_RESOURCE:
             alloc_buffer_and_copy(&sf->wif_azure_resource, value);
             break;
+        case SF_CON_WIF_AUDIENCE:
+            alloc_buffer_and_copy(&sf->wif_audience, value);
```
sfc-gh-makowalski, line 1910: `wif_audience` is allocated here via `alloc_buffer_and_copy`, but I don't see a matching `SF_FREE(sf->wif_audience)` in `snowflake_term` — the sibling `wif_azure_resource` is freed there, so this leaks on connection teardown. Suggest adding `SF_FREE(sf->wif_audience);` next to it.

include/snowflake/client.h
```diff
@@ -388,6 +393,7 @@ typedef enum SF_ATTRIBUTE {
     SF_CON_WIF_PROVIDER,
     SF_CON_WIF_TOKEN,
     SF_CON_WIF_AZURE_RESOURCE,
+    SF_CON_WIF_AUDIENCE,
```
sfc-gh-makowalski: ABI nit: inserting `SF_CON_WIF_AUDIENCE` mid-enum renumbers every attribute after it (`SF_CON_WORKLOAD_IDENTITY_IMPERSONATION_PATH`, `SF_CON_APPLICATION_PATH`, …). A consumer compiled against the old header but linked against a newer lib would then pass mismatched attribute IDs. Appending new values at the end of `SF_ATTRIBUTE` avoids that. (Same consideration for the `wif_audience` field added mid-`SF_CONNECT`, though that struct is lib-allocated so lower risk.) Feel free to ignore if libsfc is always rebuilt from source with its consumers.

cpp/lib/Authenticator.cpp
```diff
@@ -205,37 +205,7 @@ extern "C" {
...
-              config.snowflakeEntraResource = std::string(conn->wif_azure_resource);
-              log_debug("Using Azure resource: %s", conn->wif_azure_resource);
-          }
-
-          // Pass workload identity impersonation path
-          if (conn->workload_identity_impersonation_path)
-          {
-              config.workloadIdentityImpersonationPath = conn->workload_identity_impersonation_path;
-          }
+          config.configureWIFAttestation(conn);
```
sfc-gh-pbulawa: Shouldn't we return if the configuration failed?
```c++
if (!config.configureWIFAttestation(conn)) return;
```

tests/test_create_wif_attestation.cpp
```diff
@@ -1032,11 +1094,56 @@ void test_unit_oidc_attestation_missing_token(void **) {
   assert_true(!attestationOpt);
 }
 
+void test_unit_wif_attestation_config(void**)
+{
+    AttestationConfig config;
+    SF_CONNECT* conn = snowflake_init();
```
sfc-gh-pbulawa, line 1100: Shouldn't we `snowflake_term(conn)` to avoid memory leaks

sfc-gh-pbulawa (changes requested):

cpp/lib/Authenticator.cpp
sfc-gh-ext-simba-jy: I added the validation. Since config.configureWIFAttestation(conn) returns void, your suggestion wasn't applicable. Instead, I changed the return type to SF_STATUS and added the validation.

tests/test_create_wif_attestation.cpp
sfc-gh-ext-simba-jy: I added snowflake_term(conn) at the end of the testing. Thank you!

lib/client.c
sfc-gh-ext-simba-jy: I missed that. Thank you for pointing it out @sfc-gh-makowalski. I've added SF_FREE(sf->wif_audience) in snowflake_term.

include/snowflake/client.h
sfc-gh-ext-simba-jy: I moved this at the end. Thank you!

sfc-gh-makowalski (approved):
LGTM but wait for @sfc-gh-pbulawa's approval too

sfc-gh-pbulawa (approved):

PR #1022, opened 2026-07-06
SNOW-3684636: FIPS support

sfc-gh-ext-simba-hx:
teamwork issue 1404


sfc-gh-pbulawa (changes requested):
Let's make the FIPS module to be provided by the user instead of building it ourselves.
Can we make a check with envvar `FIPS_ENABLED=1`, which will check if FIPS mode is enabled, and if not, shutdown the app with appropriate message.

sfc-gh-ext-simba-hx:
> Let's make the FIPS module to be provided by the user instead of building it ourselves. Can we make a check with envvar `FIPS_ENABLED=1`, which will check if FIPS mode is enabled, and if not, shutdown the app with appropriate message.

@sfc-gh-pbulawa will do. Should we do the same for ODBC as well?

sfc-gh-ext-simba-hx:
Hi @sfc-gh-pbulawa I've made change you requested PTAL.
One more question. For the environment variable name `FIPS_ENABLED` should we add prefix on it e.g. `SF_FIPS_ENABLED` ?

sfc-gh-pbulawa:
> Hi @sfc-gh-pbulawa I've made change you requested PTAL. One more question. For the environment variable name `FIPS_ENABLED` should we add prefix on it e.g. `SF_FIPS_ENABLED` ?

That's a good idea, let's use `SF_FIPS_ENABLED`.

lib/client.c
```diff
@@ -713,6 +715,23 @@ _snowflake_validate_url_components(SF_CONNECT *sf) {
  */
 SF_STATUS STDCALL
 _snowflake_check_connection_parameters(SF_CONNECT *sf) {
+    char envBuf[MAX_PATH + 1];
+    char* fipsEnv = sf_getenv_s(SF_FIPS_ENABLED_ENV_VAR, envBuf, sizeof(envBuf));
```
sfc-gh-pbulawa, line 719: Only a missing variable or a value whose first character is '0' skips the check. Empty string, "false", "off", and "true" all enforce FIPS.

On Unix, sf_getenv_s returns the buffer for an empty value, so SF_FIPS_ENABLED= enables enforcement

ci/test/test.sh
```diff
@@ -49,7 +49,19 @@ function test_component()
     fi
     pushd $CI_TEST_DIR/../..
         cd $cmake_dir
+        if [[ "$PLATFORM" == "darwin" ]]; then
+            libsuffix="dylib"
+        else
+            libsuffix="so"
+        fi
+        tests/openssl/bin/openssl fipsinstall -module tests/openssl/lib/ossl-modules/fips.$libsuffix -out tests/openssl/fipsmodule.cnf
```
sfc-gh-pbulawa, line 57: FIPS OpenSSL config is applied to every test, not test_fips.

ci/test_win.bat
```diff
@@ -75,7 +75,15 @@ exit /b 0
     pushd %cmake_dir%
         :: test cases would need this to find data files
         set APPVEYOR_BUILD_FOLDER=%scriptdir%..
+        :: setup FIPS
+        tests\openssl\bin\openssl fipsinstall -module tests\openssl\lib\fips.dll -out tests\openssl\fipsmodule.cnf
```
sfc-gh-pbulawa, line 79: fipsinstall failure is ignored

tests/test_fips.cpp
```diff
@@ -0,0 +1,97 @@
...
+  assert_int_equal(status, SF_STATUS_SUCCESS);
+
+  snowflake_stmt_term(sfstmt);
+  snowflake_term(sf);
+  sf_unsetenv(SF_FIPS_ENABLED_ENV_VAR);
+}
+
+void test_fips_disabled(void **unused) {
+  SF_UNUSED(unused);
+  EVP_default_properties_enable_fips(nullptr, 0);
```
sfc-gh-pbulawa, line 43: global OpenSSL FIPS state is turned off and never restored which mutates process-wide default properties.

sfc-gh-pbulawa (changes requested):
Please add tests:
- SF_FIPS_ENABLED unset → connect parameter check does not call FIPS APIs / does not fail with SF_STATUS_ERROR_FIPS_NOT_ENABLED.
- SF_FIPS_ENABLED=0 → same as unset.
- SF_FIPS_ENABLED= (empty) and SF_FIPS_ENABLED=false → should not enforce (once parsing is fixed).
- SF_FIPS_ENABLED=1 with FIPS off → snowflake_error()->error_code == SF_STATUS_ERROR_FIPS_NOT_ENABLED and sqlstate 08001, without a live account (stub/EVP_default_properties_enable_fips(NULL, 0)).
- snowflake_global_get_attribute(SF_GLOBAL_FIPS_ENABLED, ...) when provider missing vs loaded (unit, no network).

tests/test_fips.cpp
sfc-gh-ext-simba-hx: addressed

ci/test_win.bat
sfc-gh-ext-simba-hx: addressed

ci/test/test.sh
sfc-gh-ext-simba-hx: addressed, move env setting to CMakefileLists.txt for test_fips only.

lib/client.c
sfc-gh-ext-simba-hx: addressed

sfc-gh-ext-simba-hx:
Hi @sfc-gh-pbulawa  I've addressed all your review comments PTAL.

tests/test_fips.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  sf_unsetenv(SF_FIPS_ENABLED_ENV_VAR);
+  EVP_default_properties_enable_fips(nullptr, fipsDefault);
+}
+
+void test_fips_not_loaded(void **unused) {
+  SF_UNUSED(unused);
+  OSSL_PROVIDER *prov = OSSL_PROVIDER_load(NULL, "fips");
+  if (prov != NULL) {
+    // unload twice to remove the initial ref count from configuration
```
sfc-gh-pbulawa, line 128: Can you use `EVP_default_properties_enable_fips(nullptr, 0)` (as in `test_fips_disabled`), assert connect fails with `SF_STATUS_ERROR_FIPS_NOT_ENABLED`, restore in teardown.

tests/CMakeLists.txt
```diff
@@ -372,6 +373,19 @@ else()
         endif()
         target_link_libraries(${T} ${TESTLIB_OPTS_CXX})
         add_test(${T} ${T})
+        if (T STREQUAL "test_fips")
```
sfc-gh-pbulawa: Add one for `valgrind_test_fips` too.

sfc-gh-pbulawa: Or enable when `test_fips` is found as substr.

lib/client.c
```diff
@@ -713,6 +715,26 @@ _snowflake_validate_url_components(SF_CONNECT *sf) {
...
+
+    if ((fipsEnv && fipsEnv[0] != '\0') &&
```
sfc-gh-pbulawa, line 721: Please add `no` to the list.

sfc-gh-pbulawa (changes requested):

tests/test_fips.cpp
sfc-gh-ext-simba-hx: This test case covers different case from `test_fips_disabled` (not loaded VS disabled AKA not including `fipsmodule.cnf` VS not having `default_properties = fips=yes` in `openssl.cnf`).
Added asserting connection failure.

tests/CMakeLists.txt
sfc-gh-ext-simba-hx: addressed

lib/client.c
sfc-gh-ext-simba-hx: addressed

PR #1024, opened 2026-07-08
SNOW-3756831: fix ocsp cache corruption in inter-process race condition

sfc-gh-ext-simba-hx:
teamwork issue 1410
Fix the race condition issue with temp file + rename approach.
Also added hardening with NULL check on response base64 decoding.


patches/curl/lib/vtls/sf_ocsp.c
```diff
@@ -2207,104 +2208,156 @@
...
       {
-        infof(data, "Other process lock the file, ignored");
-        goto end;
+        return SF_OCSP_LOCKED;
       }
       else
       {
-        infof(data, "Remove the old lock file");
-        if (remove(cache_lock_file) != 0)
+        if (remove(lock_file) != 0)
```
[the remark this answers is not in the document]
sfc-gh-ext-simba-hx: Currently the change is for reproducing the issue only will make fix once confirming the test case can cover the issue.

patches/curl/lib/vtls/sf_ocsp.c
```diff
@@ -2207,6 +2215,78 @@ char* ensureCacheDir(char* cache_dir, struct Curl_easy *data)
...
+  }
+  fclose(fp);
+
+  if (rename(tmp_file, cache_file) != 0)
+  {
+    remove(tmp_file);
+    return SF_OCSP_TMP_RENAME_ERR;
+  }
+#endif
+}
```
sfc-gh-makowalski, line 2305: `sf_ocsp_write_cache` is declared `int`, but there's no `return SF_OCSP_SUCCESS;` on the success path — after a successful `rename`/`MoveFileEx`, both branches fall through to this closing brace, so the function returns an indeterminate value (UB). The caller's `switch (res)` then reads garbage, so even the success log (`"Write OCSP Response to cache file"`) is unreliable. CI doesn't catch this because the curl patch is built as a separate dependency, outside the `-Werror`/extra-warnings gate. Please add `return SF_OCSP_SUCCESS;` before the closing brace.

patches/curl/lib/vtls/sf_ocsp.c
```diff
@@ -2207,6 +2215,78 @@ char* ensureCacheDir(char* cache_dir, struct Curl_easy *data)
...
+  if (fp == NULL)
+  {
+    return SF_OCSP_TMP_OPEN_ERR;
+  }
+  if (fprintf(fp, "%s", content) < 0)
+  {
+    fclose(fp);
+    return SF_OCSP_TMP_WRITE_ERR;
+  }
+  fclose(fp);
```
sfc-gh-makowalski: `fclose()`'s return value isn't checked here (same on the Windows branch, the `fclose(fp)` before `MoveFileEx`). With buffered I/O, data can be flushed only at `fclose`, so a flush failure (e.g. `ENOSPC`) leaves a truncated temp file that then gets renamed over the good cache — reintroducing exactly the corruption this PR is meant to prevent. The previous code checked `fclose`. Suggest checking it and returning `SF_OCSP_TMP_WRITE_ERR` (skipping the rename) on failure.

patches/curl/lib/vtls/sf_ocsp.c
```diff
@@ -2207,6 +2215,78 @@ char* ensureCacheDir(char* cache_dir, struct Curl_easy *data)
...
+#else
+  snprintf(tmp_file, sizeof(tmp_file), "%s.tmp.XXXXXX", cache_file);
+  int tmp_fd = mkstemp(tmp_file);
+  if (tmp_fd == -1)
+  {
+    return SF_OCSP_TMP_NAME_ERR;
+  }
+
+  fp = fdopen(tmp_fd, "w");
+  if (fp == NULL)
```
sfc-gh-makowalski, line 2278: Error-path temp-file leak: if `fdopen` fails here, both the `mkstemp` fd and the on-disk temp file are leaked (no `close`/`remove`). Same for the `fprintf` failure just below (the `fclose` closes the fd but leaves the temp file behind), and on the Windows branch after `GetTempFileName` (which with `uUnique=0` creates the file, so the `fopen`/`fprintf` failure paths leak it too). Repeated failures would accumulate `*.tmp.XXXXXX` / `sfocsp*` files in the cache dir. Please `close`/`remove` the temp file on these error paths.

sfc-gh-makowalski (changes requested):
The atomic temp-file + same-directory `rename`/`MoveFileEx` approach is the right fix, and dropping the old `.lck` stale-lock logic is a nice simplification. Noting this is marked WIP — three concrete issues to address before it's final, flagged inline. (The base64-decode NULL check + bad-entry eviction looks good and is safe — the caller does a single lookup under the mutex, not an iteration.)

patches/curl/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: addressed.

patches/curl/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: addressed

patches/curl/lib/vtls/sf_ocsp.c
sfc-gh-ext-simba-hx: addressed.

patches/curl/lib/vtls/sf_ocsp.c
sfc-gh-makowalski: Thanks, the POSIX paths and the Windows fprintf/fclose paths are all cleaned up now. One sub-case is still open: the Windows fopen-null path (the `if (fp == NULL) return SF_OCSP_TMP_OPEN_ERR;` right after `GetTempFileName`) returns without `DeleteFile(tmp_file)`. Since `GetTempFileName` with uUnique=0 already created the file, that path leaks it. Could you add a `DeleteFile(tmp_file);` there too? Minor and non-blocking.

sfc-gh-ext-simba-hx: sorry my bad. Addressed.

tests/test_unit_ocsp_cache.cpp
```diff
@@ -0,0 +1,153 @@
...
+void test_corrupted_cache(void **unused)
+{
+  SF_UNUSED(unused);
+  char cache_file[4096];
+
+  get_ocsp_cache_file(cache_file);
+  if (!boost::filesystem::is_regular_file(cache_file))
+  {
+    printf("Skipping - cache file not available\n");
+  }
```
sfc-gh-makowalski, line 116: This branch prints "Skipping" but doesn't `return`, so it falls through: `loadFileInString` on a missing file yields `""`, `snowflake_cJSON_Parse("")` returns `NULL`, and `assert_non_null(ocsp_cache_root)` then **fails the test** instead of skipping it — with a misleading "Skipping…" line right before the failure.

This is reachable: within this binary, `test_concurrent_cache_writing` writes to `.` (cwd), not the real cache path, and the comment above forbids a connection before this case — so nothing here populates `~/.cache/.../ocsp_response_cache.json`. It exists only if an earlier test binary connected on the same runner, so a clean/isolated/sharded run (or a dev running just this binary) will hit the missing-file path and fail red.

Please `return;` in the skip branch:
```cpp
if (!boost::filesystem::is_regular_file(cache_file))
{
  printf("Skipping - cache file not available\n");
  return;
}
```
(Or have the test create/seed its own cache so it doesn't depend on prior-test state.)

sfc-gh-makowalski (changes requested):
Thanks for addressing the earlier `sf_ocsp_write_cache` feedback — the success return, `fclose` checks, and temp-file cleanup on all error paths (incl. the Windows fopen-null path) all look good now, and the atomic temp+rename fix is correct. One remaining issue in the new test: the "Skipping" branch doesn't actually skip (details inline).

tests/test_unit_ocsp_cache.cpp
sfc-gh-ext-simba-hx: addressed by adding return. Usually we run test cases together and in the case running this test individually, when seeing "Skipping" in the output we could generate cache file by making a simple connection before running the test.

sfc-gh-makowalski (approved):

PR #1025, opened 2026-07-09
SNOW-3684639: [Libsfc] Confirm host/server_url overrides the connection domain

sfc-gh-ext-simba-jy:
- [x] Add a connection parameter called "wif_host:, which overrides the default host in AWS and GCP attestation
- [x] Add test cases.


tests/test_create_wif_attestation.cpp
```diff
@@ -590,24 +599,27 @@ void test_unit_gcp_impersonation_single_account_success(void **) {
   assert_true(issuer == GCP_TEST_ISSUER);
 }
 
-void test_unit_gcp_impersonation_single_account_success_with_custom_audience(void**) {
+void test_unit_gcp_impersonation_single_account_success_with_custom_wif_cofig(void**) {
```
sfc-gh-makowalski: Typo: `wif_cofig` → `wif_config`. Same typo in the `cmocka_unit_test(...)` registration at line 1178.

include/snowflake/client.h
```diff
@@ -398,6 +398,7 @@ typedef enum SF_ATTRIBUTE {
     SF_CON_LOG_QUERY_TEXT,
     SF_CON_LOG_QUERY_PARAMETERS,
     SF_CON_WIF_AUDIENCE,
+    SF_CON_WIF_HOST,
```
sfc-gh-makowalski, line 412: `SF_CON_WIF_HOST` carries different semantics depending on the cloud provider, which isn't documented here:

- **AWS** (`AWSUtils.cpp:175`): expects a bare hostname — e.g. `sts.us-gov-east-1.amazonaws.com`. The code prepends `https://` and appends `/`.
- **GCP** (`GcpAttestation.cpp:90`): expects a full base URL including scheme and path prefix — e.g. `https://iamcredentials.privategoogleapis.com/v1`. It is concatenated directly with `/projects/-/serviceAccounts/...`.

A caller has no way to know which format to use without reading the provider-specific implementation. Please either (a) add a doc comment here explaining the per-provider format expectation, or (b) normalise the input internally so both providers accept the same format (e.g. always a bare hostname, or always a full base URL).

sfc-gh-makowalski (changes requested):

tests/test_create_wif_attestation.cpp
sfc-gh-ext-simba-jy: Fixed.

include/snowflake/client.h
sfc-gh-ext-simba-jy: I added the doc and add two different host apis in attestations. As a result, callers can use both a bare url or a full url.

sfc-gh-makowalski (approved):

PR #1029, opened 2026-07-16
SNOW-3781573: fix sf_getaddrinfo_mutex leak on getaddrinfo() error path

sfc-gh-boler:
## Summary
The SNOW-119090 patch in `patches/curl-8.20.0.patch` locks a process-global mutex (`sf_getaddrinfo_mutex`) around the system `getaddrinfo()` call but only unlocks it on the success path. **Any error return from `getaddrinfo()` leaks the mutex** — and since it's process-global and non-robust, one failed lookup deadlocks *all* subsequent DNS in the process on `pthread_mutex_lock()`. Deterministic hang, not a race.

The lock is gated by `sf_enable_getaddrinfo_lock` (default `0`); clients such as the ODBC driver enable it.

## Why it appears on RHEL 8 but not RHEL 7
The driver didn't change — glibc did (2.17 → 2.28), plus `ipv6.disable=1`. Slow/overloaded upstream DNS → `EAI_AGAIN`; NXDOMAIN → `EAI_NONAME`; `AF_UNSPEC` with IPv6 disabled → `EAI_ADDRFAMILY`/`EAI_NODATA`. All non-zero, all hit the leaking return. RHEL 8's resolver stack surfaces these where RHEL 7 masked them, and the global mutex serializes all DNS, so a slow resolver + the ~48-thread chunk downloader pile up and make an error return (hence the leak) far more likely.

This is **not** the glibc `__check_pf`/`pthread_cancel` theory: libcurl *detaches* resolver threads (`Curl_thread_destroy` = `pthread_detach`), never cancels them, so no glibc lock is orphaned.

## Fix
Narrow the critical section to just the `getaddrinfo()` call and unlock immediately after it, before any early return. Single unlock site that every exit passes through → no path can orphan the lock; also shrinks the serialized region. The following list construction is thread-local and doesn't need the lock. Unlock-failure handling is now report-only so a spurious unlock error can't fail a good resolve or leak `aihead`.

## Testing
- `git apply --check` clean against `curl-8.20.0`.
- Resulting `Curl_getaddrinfo_ex()`: exactly one lock + one unlock, `aihead` still freed, braces balanced.
- Standalone reproducer: lock enabled + one failed resolve → all later resolves hang (reproduced); same flow with the fix → no deadlock.

### DIFF
--- a/patches/curl-8.20.0.patch
+++ b/patches/curl-8.20.0.patch
@@ -400,7 +400,7 @@ index fd26c5f0bc..38c79131d7 100644
  int Curl_getaddrinfo_ex(const char *nodename,
                          const char *servname,
                          const struct addrinfo *hints,
-@@ -104,9 +129,22 @@ int Curl_getaddrinfo_ex(const char *nodename,
+@@ -104,9 +129,36 @@ int Curl_getaddrinfo_ex(const char *nodename,
    struct Curl_addrinfo *ca;
    size_t ss_size;
    int error;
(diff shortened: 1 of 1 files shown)


sfc-gh-makowalski (approved):
Approving — this is a real bug in the SNOW-119090 patch (mutex leak on getaddrinfo() error return when ForceLockGetaddrinfo=true).

However, this does not address SNOW-3781573 / the DTCC customer issue. The customer core traces (corepid_1132100, stack.zip) show threads blocked on glibc's __check_pf inside getaddrinfo(), with libcurl cleanup threads blocked in Curl_thread_join — not on sf_getaddrinfo_mutex. ForceLockGetaddrinfo is not set in the customer environment.

The evidenced customer root cause aligns with curl [#18532](https://github.com/curl/curl/issues/18532): pthread_cancel in curl 8.16.0's threaded resolver can orphan glibc locks mid-getaddrinfo(). DTCC runs ODBC 3.15.0 (curl 8.16.0.1). Fixed in curl 8.17.0+ (ODBC 3.17/3.18). Durable fix remains c-ares libcurl rebuild.

patches/curl-8.20.0.patch
```diff
@@ -421,25 +421,22 @@ index fd26c5f0bc..38c79131d7 100644
...
++  /* SNOW-119090: only the system getaddrinfo() call above needs to be
++     serialized (it may trigger a lazy libnss_*.so load). Release the lock
++     here, before any early return below, so it can never be orphaned on an
++     error path -- e.g. EAI_AGAIN from a slow/overloaded resolver, which
++     would otherwise deadlock all name resolution in the process. */
 +  if(sf_enable_getaddrinfo_lock == 1) {
 +    mutex_error = pthread_mutex_unlock(&sf_getaddrinfo_mutex);
-+    if(mutex_error) {
++    if(mutex_error)
 +      Curl_print_pthread_error(mutex_error);
```
sfc-gh-mhofman, line 434: function returns error which is no longer assigned here with a mutex_error value. Is this the intend? @sfc-gh-boler

sfc-gh-mhofman: We've checked with @sfc-gh-makowalski and this is actually fixing the problem here so closing this comment

sfc-gh-boler:
> The evidenced customer root cause aligns with curl [#18532](https://github.com/curl/curl/issues/18532): pthread_cancel in curl 8.16.0's threaded resolver can orphan glibc locks mid-getaddrinfo(). DTCC runs ODBC 3.15.0 (curl 8.16.0.1). Fixed in curl 8.17.0+ (ODBC 3.17/3.18). Durable fix remains c-ares libcurl rebuild.

Ok, so there is a buggy curl version at play too. We need them to upgrade. Moving resolution to /etc/hosts is a viable temporary workaround.

PR #1042, opened 2026-07-28
SNOW-3853133: fix delay on aws identity detector

sfc-gh-ext-simba-hx:
teamwork issue 1418

### DIFF
--- a/cpp/lib/PlatformDetection.cpp
+++ b/cpp/lib/PlatformDetection.cpp
@@ -4,6 +4,7 @@
 #include "snowflake/HttpClient.hpp"
 #include <boost/algorithm/string.hpp>
 #include <exception>
+#include <fstream>
 #include "../util/SnowflakeCommon.hpp"
 #include "../logger/SFLogger.hpp"
 
@@ -32,6 +33,9 @@ static std::string gcpMetadataBaseURL = GCP_METADATA_BASE_URL;
 static std::string gcpMetadataFlavorHeaderName = "Metadata-Flavor";
 static std::string gcpMetadataFlavor = "Google";
 
+// helper functions
+namespace
+{
 std::string getEnvironmentVariableValue(const std::string& envVarName)
 {
   char envbuf[MAX_ENV_VARIABLE_LENGTH];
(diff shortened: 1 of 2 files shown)


cpp/lib/PlatformDetection.cpp
```diff
@@ -201,19 +198,74 @@ PlatformDetectionStatus detectGcpIdentity(long timeout)
...
+  HttpRequest tokenReq{
+    HttpRequest::Method::PUT,
+    tokenUrl,
+    {{"X-aws-ec2-metadata-token-ttl-seconds", "21600"}},
+  };
+  HttpClientConfig cfg = { 0, timeout, 0, timeout };
+  std::unique_ptr<IHttpClient> httpClient(IHttpClient::createSimple(cfg));
+
+  auto tokenRespOpt = httpClient->run(tokenReq);
+  if (!tokenRespOpt || tokenRespOpt.get().code != 200)
```
sfc-gh-dszmolka: what would happen in a scenario when IMDS is enabled, but on IMDSv1 (without token) endpoint only?

in other drivers we seem to have a fallback if IMDSv2 is disabled: IMDS token PUT endpoint fails with a non-timeout response like 4xx, so it is reachable , which falls back to plain GET without the token. I checked only Python, JDBC, and Go, all 3 seem to have this fallback.

would it be possible to also consider "try v2, fall back to v1 if the token endpoint rejects us but is reachable" behaviour here too, before deciding IMDS is unreachable?

sfc-gh-ext-simba-hx: @sfc-gh-dszmolka Yes that's doable but might be unnecessary? For my understanding IMDSv2 is always enabled you can only choose whether to disable IMDSv1? https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/instance-metadata-transition-to-version-2.html

sfc-gh-dszmolka: checked right now on AWS Console, there's 2 options indeed:
* V1 and V2 (token optional) - "EC2 recommends using metadata version 2 unless you explicitly require metadata version 1."
* V2 only (token required) - "For V2 requests, you must include a session token in all instance metadata requests. Applications or agents that use V1 for instance metadata access will break."

This is only for newly launched instances. For existing instances, i did not find a reliable, recent source of statistics. 

From another side, security perspective, it would be straight the best practice to only support v2, following what AWS is doing. However, the main reason of us even calling the IMDS vX is to gather platform information, so worst case scenario we cannot gather it on v2. 

i'm fine with going only with IMDSv2 support after this validation, although i must admit i feel somewhat uncomfortable with "deviating from rest of the drivers" (just for the sheer fact of it) even though it would be possibly an undetectable problem or of less importance. Please, at least can we leave a comment why it was left  on v2 support only? :)

sfc-gh-ext-simba-hx: @sfc-gh-dszmolka Checked the implementation in [AWS SDK](https://github.com/aws/aws-sdk-cpp/blob/main/src/aws-cpp-sdk-core/source/internal/AWSHttpResourceClient.cpp#L330)
It fallback to IMDSv1 if getting token with v2 failed I've changed the implementation to do the same.

cpp/lib/PlatformDetection.cpp
```diff
@@ -201,19 +198,78 @@ PlatformDetectionStatus detectGcpIdentity(long timeout)
 
 PlatformDetectionStatus detectAwsIdentity(long timeout)
 {
-  SF_UNUSED(timeout);
-  auto awsSdkInit = AwsUtils::initAwsSdk();
-  Aws::Auth::DefaultAWSCredentialsProviderChain credentialsProvider;
```
sfc-gh-pbulawa, line 206: We are missing the shared credentials check now ~/.aws/credentials. Can you readd it?

tests/test_platform_detection/test_platform_detection.cpp
```diff
@@ -63,23 +63,22 @@ void test_detection_endpoint_core(const std::string& expectedPlatform, const std
...
   assert_true(detectedPlatforms.size() <= 2);
   assert_true(std::find(detectedPlatforms.begin(), detectedPlatforms.end(), expectedPlatform) != detectedPlatforms.end());
-  sf_unsetenv("AWS_ENDPOINT_URL_STS");
 }
 
 void test_ec2Instance(void**) {
   test_detection_endpoint_core("is_ec2_instance", "aws_ec2_instance_success.json");
 }
 
+void test_awsIdentity(void**) {
```
sfc-gh-pbulawa, line 78: Register the method in cmocka_run_group_tests tests[] array.

cpp/lib/PlatformDetection.cpp
```diff
@@ -270,21 +317,79 @@ void getDetectedPlatforms(std::vector<std::string>& detectedPlatforms, long time
...
+
+        task.worker = std::thread([detector, timeoutMs, state]() {
+          bool isDetected = (detector(timeoutMs) == PLATFORM_DETECTED);
+          {
+            std::lock_guard<std::mutex> lk(state->mtx);
+            state->detected = isDetected;
+            state->done = true;
+          }
+          state->cv.notify_one();
+        });
```
sfc-gh-pbulawa, line 550: Let's wrap it with `try/catch`, as right now if exception is thrown in a detector or HttpClient, it is not visible.

sfc-gh-pbulawa (changes requested):

tests/test_platform_detection/test_platform_detection.cpp
sfc-gh-ext-simba-hx: addressed

cpp/lib/PlatformDetection.cpp
sfc-gh-ext-simba-hx: addressed

cpp/lib/PlatformDetection.cpp
sfc-gh-ext-simba-hx: added.

sfc-gh-pbulawa (approved):

PR #1044, opened 2026-07-31
SNOW-3834837 migrate azure sdk to Azure SDK for C++

sfc-gh-ext-simba-hx:
migrate to the latest azure sdk for c++ (azure-stoarge-blobs 12.18.0) from the deprecated azure-storage-cpplite (will remove in a separated PR to keep this PR in a reviewable amount of changed files)
Also fixed reading overflow around sf_strncpy which caused segment fault in test_unit_logger after the migration, tighten the boundary check and null terminator.


cpp/SnowflakeAzureClient.cpp
```diff
@@ -79,31 +82,48 @@ SnowflakeAzureClient::SnowflakeAzureClient(StageInfo *stageInfo,
...
-                account, m_parallel, caBundleFile,
-                proxy->getHost(),
-                proxy->getPort(),
-                proxy->getUser(),
-                proxy->getPwd(),
-                proxy->getNoProxy());
+
+  try {
+    Azure::Storage::Blobs::BlobClientOptions options;
+    options.Retry.MaxRetries = m_maxRetries;
```
sfc-gh-pbulawa, line 97: `m_maxRetries` seems to be used uninitialized.

`FileTransferAgent` calls `setMaxRetries()` after `getClient()`, but Azure only reads retry options at construction. `setMaxRetries` only assigns the member and does not recreate/update the client.

cpp/SnowflakeAzureClient.hpp
```diff
@@ -64,6 +68,36 @@ struct MultiDownloadCtx_a
...
+    this->m_stream.read(reinterpret_cast<char*>(buffer), count);
+    return m_stream.gcount();
+  }
+
+  int64_t Length() const override
+  {
+    return m_length;
+  }
+
+  void Rewind() override
```
sfc-gh-pbulawa, line 92: Single-part upload uses this adapter around a non-seekable cipher stream. If MaxRetries > 0, Azure retry can re-read from EOF and upload empty/corrupt data or fail opaquely.

cpp/SnowflakeAzureClient.cpp
```diff
@@ -79,31 +82,48 @@ SnowflakeAzureClient::SnowflakeAzureClient(StageInfo *stageInfo,
   std::string account_name = m_stageInfo->storageAccount;
   std::string sas_key = m_stageInfo->credentials[azuresaskey];
   std::string endpoint = account_name + "." + m_stageInfo->endPoint;
```
sfc-gh-pbulawa: `cpplite` client prefixed a scheme automatically, but the new one does not. Let's make sure that when https is enabled, the scheme prefix is there.

lib/client.c
```diff
@@ -1151,26 +1157,30 @@ snowflake_global_get_attribute(SF_GLOBAL_ATTRIBUTE type, void *value, size_t siz
             break;
         case SF_GLOBAL_CLIENT_CONFIG_FILE:
             if (CLIENT_CONFIG_FILE) {
-              if (strlen(CLIENT_CONFIG_FILE) > size - 1) {
+              size_t copylen = strlen(CLIENT_CONFIG_FILE) + 1;
+              if (copylen > size) {
                 return SF_STATUS_ERROR_BUFFER_TOO_SMALL;
               }
               sf_strncpy(value, size, CLIENT_CONFIG_FILE, size);
```
sfc-gh-pbulawa:
```suggestion
              sf_strncpy(value, size, CLIENT_CONFIG_FILE, copylen);
```

cpp/SnowflakeAzureClient.cpp
```diff
@@ -345,61 +491,112 @@ RemoteStorageRequestOutcome SnowflakeAzureClient::doSingleDownload(
...
-  dataStream->flush();
-  if(errno == 0)
-    return RemoteStorageRequestOutcome::SUCCESS;
+  try {
+    auto containerClient = m_blobServiceClient->GetBlobContainerClient(cont);
+    auto blobClient = containerClient.GetBlockBlobClient(blob);
+    auto response = blobClient.Download();
+    if (!response.Value.BodyStream)
+    {
+      CXX_LOG_ERROR("%s file donwload failed:: BodyStream is NULL",
```
sfc-gh-pbulawa:
```suggestion
      CXX_LOG_ERROR("%s file download failed:: BodyStream is NULL",
```

patches/azure-sdk-cpp-12.18.0.patch
```diff
@@ -0,0 +1,94 @@
...
+diff --git a/sdk/core/azure-core/src/http/curl/curl_connection_private.hpp b/sdk/core/azure-core/src/http/curl/curl_connection_private.hpp
+index 60ccb558f..6a86cf8d0 100644
+--- a/sdk/core/azure-core/src/http/curl/curl_connection_private.hpp
++++ b/sdk/core/azure-core/src/http/curl/curl_connection_private.hpp
+@@ -131,6 +131,12 @@ namespace Azure { namespace Core {
+        */
+       virtual bool IsExpired() = 0;
+ 
++      /**
++       * @brief Checks whether this CURL connection is ued proxy.
```
sfc-gh-pbulawa:
```suggestion
+       * @brief Checks whether this CURL connection is used proxy.
```

sfc-gh-pbulawa (changes requested):

cpp/SnowflakeAzureClient.hpp
sfc-gh-ext-simba-hx: Created `m_blobServiceClientNoRetry` with `MaxRetries` set to 0 for single-part upload.

cpp/SnowflakeAzureClient.cpp
sfc-gh-ext-simba-hx: addressed by passing maxRetries to Azure client constructor.

cpp/SnowflakeAzureClient.cpp
sfc-gh-ext-simba-hx: Existing code with cpplite hardcoded `use_https` as true. Hardcoded to add https prefix here.

lib/client.c
sfc-gh-ext-simba-hx: addressed

cpp/SnowflakeAzureClient.cpp
sfc-gh-ext-simba-hx: addressed

patches/azure-sdk-cpp-12.18.0.patch
sfc-gh-ext-simba-hx: addressed

sfc-gh-pbulawa (approved):

PR #1051, opened 2026-08-14
Build vs2019 artifacts with vs2022 and v142 toolset

sfc-gh-ext-simba-hx:
teamwork issue 1420
Do we need to bump up build versions to enforce rebuild/update artifacts on S3? In theory it should be the same (VS16 artifacts from VS16 toolset and VS17 ones from VS17 toolset) so the artifacts built before stay valid.


scripts/build_curl.bat
sfc-gh-pbulawa: VS16 still uses 2019 generator below in cmake.

sfc-gh-pbulawa: Use -G "%cmake_generator%" -A %cmake_architecture% -T "%ToolSetVersion%" and delete the vc_version / VS15 mapping.

scripts/_init.bat
```diff
@@ -60,45 +60,29 @@ if "%target_name%"=="" (
...
-        ) else if exist "C:\Program Files (x86)\Microsoft Visual Studio\2022\BuildTools\VC" (
-            set VCINSTALLDIR=C:\Program Files ^(x86^)\Microsoft Visual Studio\2022\BuildTools\VC
-        ) else (
-            set "VCINSTALLDIR="
-            for /f "usebackq delims=" %%i in (`"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vswhere.exe" -latest -products * -requires Microsoft.VisualStudio.Component.VC.Tools.x86.x64 -property installationPath 2^>nul`) do set "VCINSTALLDIR=%%i\VC"
-            if not defined VCINSTALLDIR (
-                echo Set environment variable VCINSTALLDIR to sepecify Visual Studio 2022 install path.
-                goto :error
-            )
+::Always use Visual Studio 2022
```
sfc-gh-pbulawa, line 72: Previously an unknown/VS15 value left cmake_generator empty and failed. Now the generator is always VS2022 and vsdir=vs17 unless VS16 overrides it. VS15 therefore writes vs17 artifacts until setup_visual_studio fails with no VisualStudioVersion is set. VS15.

sfc-gh-pbulawa (changes requested):

scripts/_init.bat
sfc-gh-ext-simba-hx: addressed

scripts/build_curl.bat
sfc-gh-ext-simba-hx: addressed

scripts/_init.bat
```diff
@@ -60,44 +60,32 @@ if "%target_name%"=="" (
...
-if /I "%vs_version%"=="VS15" (
-    set cmake_generator=Visual Studio 15 2017
-    set vsdir=vs15
-)
-if "%cmake_generator%"=="" (
-    echo Specify the VS_VERSION to the Visual Studio Version [VS17, VS16, VS15]
-    goto :error
+
+::Always use Visual Studio 2022
+set cmake_generator=Visual Studio 17 2022
```
sfc-gh-pbulawa, line 73: **[ISSUE] `scripts/_init.bat:106-128` — dead `:find_vs16` left behind**

This hunk removes `call :find_vs16` and always points `VCINSTALLDIR` at VS2022, but `:find_vs16` (the VS2019 Enterprise/Professional/Community/BuildTools + vswhere `[16.0,17.0)` lookup) is still at the bottom of the file after `:error`.

Nothing calls it anymore. Please delete that label so we do not leave a second, unused VS2019 host path in the script.

sfc-gh-pbulawa (changes requested):
## Requested changes

The VS2022-hosted v142 model is right, but VS16 is still not actually built or installed that way in CI, and `_init.bat` still carries the old VS2019 lookup.

These two workflow files are not in this PR's diff, but they have to change with it or `VS16` remains unproven:

**[ISSUE] `.github/workflows/pdo-vendor.yml:239-263` — VS16 job still installs VS2019 Build Tools**

After this PR, `VCINSTALLDIR` is always VS2022 and `setup_visual_studio VS16` runs that `vcvarsall` with `-vcvars_ver=14.2`. A side-by-side VS2019 BT install does not add 14.2 under VS2022's `VC\Tools\MSVC`.

Install the v142 toolset **on VS2022** (`Microsoft.VisualStudio.Component.VC.v142.x86.x64` and the matching Spectre libs). Drop the VS2019 installer.

**[ISSUE] `.github/workflows/build-test.yml:96-104` — PR CI never runs VS16**

The Windows matrix is `vs_version: [ 'VS17' ]` only. The v142 path is the point of this PR and is never exercised here. Add at least one `VS16` Windows cell (x64 Release is enough) that uses VS2022 + v142, no VS2019.

**[ISSUE] `scripts/_init.bat:106-128` — dead `:find_vs16` left behind**

The `call :find_vs16` was removed, but the label and VS2019 path lookup remain after `:error`. Nothing reaches them. Delete `:find_vs16` so the next reader does not think standalone VS2019 is still a supported host.

PR #1054, opened 2026-08-19
SNOW-3977525: fix issue buffer not reset correctly with retry

sfc-gh-ext-simba-hx:
teamwork issue 1423
confirmed the test code can catch the issue with [commit](https://github.com/snowflakedb/libsnowflakeclient/pull/1054/commits/6e830599078e196ceb21739a0f054332499109b2) 
Although still waiting for customer test result the change should be good to go. https://github.com/snowflakedb/snowflake-sdks-drivers-issues-teamwork/issues/1423#issuecomment-5391517545


cpp/SnowflakeS3Client.cpp
```diff
@@ -262,9 +265,21 @@ void Snowflake::Client::SnowflakeS3Client::uploadParts(MultiUploadCtx * uploadCt
...
 
+  auto streamBuf = Aws::MakeShared<Aws::Utils::Stream::PreallocatedStreamBuf>(
+      "",
+      reinterpret_cast<unsigned char*>(uploadCtx->buf->getDataBuffer()),
+      uploadCtx->buf->getSize());
+
   uploadPartRequest.SetContentType(CONTENT_TYPE_OCTET_STREAM);
   uploadPartRequest.SetContentLength(uploadCtx->buf->getSize());
-  uploadPartRequest.SetBody(Aws::MakeShared<Aws::IOStream>("", uploadCtx->buf));
+#if defined(_DEBUG) || defined(TEST_DEBUG)
```
sfc-gh-pbulawa: Let's use wiremock instead of having a switch in the production code.

cpp/SnowflakeS3Client.cpp
```diff
@@ -262,9 +265,21 @@ void Snowflake::Client::SnowflakeS3Client::uploadParts(MultiUploadCtx * uploadCt
...
+  // for testing retry with debug build, fail the first attempt
+  if (!uploadCtx->m_isRetry)
+  {
+    uploadPartRequest.SetContentLength(uploadCtx->buf->getSize() + 10);
+  }
+#endif
+  uploadPartRequest.SetBody(Aws::MakeShared<Aws::IOStream>("", streamBuf.get()));
```
sfc-gh-pbulawa: SetBody takes `streamBuf.get()`; `iostream` does not own the streambuf. Locals are destroyed in reverse order, so `streamBuf` dies first and the request destructor then destroys an `IOStream` whose `rdbuf()` is already freed. Standard `~iostream` is not supposed to touch `rdbuf()`, but this is a dangling pointer. Declare `streamBuf` before `uploadPartRequest` (or otherwise keep the buf alive until after the request is destroyed). AWS TransferManager uses `Aws::New<PreallocatedStreamBuf>` and passes that pointer into `IOStream` for the same reason.

sfc-gh-pbulawa (changes requested):

cpp/SnowflakeS3Client.cpp
sfc-gh-ext-simba-hx: Declare streamBuf before uploadPartRequest.

cpp/SnowflakeS3Client.cpp
sfc-gh-ext-simba-hx: It might not easy to add test case with wiremock that would need to simulate the entire scenario of large file uploading. Remove the debug branch for now and we could add the test case in a separated PR later. @sfc-gh-pbulawa Please let me know if you think the test case is a blocker of this fix.

sfc-gh-pbulawa (approved):

cpp/SnowflakeS3Client.cpp
sfc-gh-pbulawa: I'm fine with merging it that way, but please followup with the test case later.

PR #1059, opened 2026-08-25
SNOW-2391156: Add CRL cache cleanup mechanism

sfc-gh-pbulawa:
## Summary

Before this change the CRL cache only ever grew. Entries were added on download
and read back on subsequent handshakes, but nothing ever removed them, so a
long-lived process accumulated in-memory CRLs indefinitely and the on-disk cache
directory accumulated files for CRLs that had long since expired.

This adds a cleanup pass over both tiers of the cache, plus a background thread
that runs it periodically.

**In-memory cache.** Each cached entry now records when it was obtained.
`download_time` is set to the current time when a CRL is downloaded, and to the
cache file's mtime when an entry is promoted from disk into memory. Cleanup drops
an entry when either the CRL's own `nextUpdate` has passed, or the entry has been
cached for longer than the configured validity window.

**On-disk cache.** Cleanup enumerates the cache directory (`FindFirstFile` on
Windows, `opendir`/`readdir` elsewhere) and deletes a file when it cannot be
parsed as a PEM CRL at all, when it has no `nextUpdate`, or when its `nextUpdate`
is older than the removal delay. The delay means recently expired CRLs are kept
for a while rather than deleted the moment they go stale, so a process that is
briefly offline can still fall back on them. If OpenSSL cannot compare the two
timestamps it reports a distinct failure value, and in that case the file is left
in place rather than being treated as expired.

**Scheduling.** `initCertCRL()` now starts a cleanup thread that runs one pass
immediately and then repeats on a fixed interval. It sleeps in one-second slices
and re-checks a stop flag, so shutdown does not have to wait out a full interval.
Thread creation and joining are wrapped in the same small platform abstraction
already used for the cache mutex (`CreateThread`/`WaitForSingleObject` on Windows,
pthreads elsewhere).

**Shutdown.** A new exported `termCertCRL()` is called from libcurl's
`ossl_cleanup`, so the cleanup thread is joined *before* OpenSSL tears itself
down; a thread touching `X509_CRL` objects after that point would be unsafe. The
existing `atexit` handler remains as a backstop and is now idempotent via an
initialization flag.

The memory pass runs under the existing cache mutex, but the disk pass runs
outside it, so directory scanning and file parsing never block a TLS handshake
that is trying to read the in-memory cache.

Also picked up along the way, since the cleanup code exercised these paths:
- `ucrl_unregister()` compared URI *pointers* rather than string contents, so it
  never actually found an entry. This meant re-registering a URI appended a
  duplicate instead of replacing it.
- The URI copied into a cache entry is now explicitly NUL-terminated, since
  `strncpy` does not guarantee it at exactly `PATH_MAX`.
- `get_cache_dir()` and `mkdir_if_not_exists()` now tolerate a `NULL`
  `struct Curl_easy *`, which the background thread has no handle to. Previously
  they would have passed it straight to `infof`/`failf`.

### Environment variables

All values are in seconds. A value that is empty, non-numeric, negative, or has
trailing garbage falls back to the default.

| Variable | Default | Effect |
| --- | --- | --- |
| `SF_CRL_CACHE_VALIDITY_TIME` | `86400` (24 hours) | How long a CRL may stay in the in-memory cache after it was downloaded before it is evicted, regardless of its own validity period. |
| `SF_CRL_ON_DISK_CACHE_REMOVAL_DELAY` | `604800` (7 days) | Grace period after a cached CRL file's `nextUpdate` before the file is deleted from disk. |
| `SF_CRL_CACHE_CLEANUP_INTERVAL` | `3600` (1 hour) | How often the background cleanup runs. Set to `0` to disable the thread entirely; cleanup then only happens if `cleanupCertCRLCache()` is called directly. |

`SF_CRL_RESPONSE_CACHE_DIR` is unchanged in behaviour and still selects the
on-disk cache directory; it is now referenced through a named constant instead of
a bare string literal, and it determines which directory the disk cleanup scans.

## Context

Part of the cross-driver CRL work. The other Snowflake drivers already expire and
prune their CRL caches; libsnowflakeclient was the remaining one that cached CRLs
without ever reclaiming them. The knobs and defaults above are chosen to line up
with those drivers so behaviour is consistent across the fleet.

Because this changes vendored libcurl source under `patches/curl`, the curl
dependency build version is bumped from `1` to `2` (in both `build_curl.sh` and
`build_curl.bat`). CI restores a dependency cache keyed on the component version,
so without the bump every job would keep linking a prebuilt libcurl that predates
this patch, and the new symbols would not exist. The aws, azure, and oob build
versions are bumped for the same reason, as their cached artifacts are built
against libcurl.


patches/curl/lib/vtls/sf_crl.c
```diff
@@ -73,6 +85,36 @@ int _mutex_term(SF_MUTEX_HANDLE *lock) {
 #endif
 }
 
+static int _thread_init(SF_THREAD_HANDLE *thread, void *(*proc)(void *), void *arg)
+{
+#ifdef _WIN32
+  *thread = CreateThread(NULL, 0, (LPTHREAD_START_ROUTINE)proc, arg, 0, NULL);
```
sfc-gh-makowalski: `CreateThread` expects `DWORD WINAPI ThreadProc(LPVOID)`, not `void *(*)(void *)`. Casting across that boundary is undefined on 32-bit Windows (`__stdcall` vs `__cdecl`, pointer vs `DWORD` return).

Please add a `WINAPI` wrapper that returns `DWORD`, and keep the pthreads signature on the Unix path only.

patches/curl/lib/vtls/sf_crl.c
```diff
@@ -792,16 +845,210 @@ static int error_handler(int ok, X509_STORE_CTX *ctx)
...
+SF_PUBLIC(void) cleanupCertCRLCache(void)
+{
+  if (!crl_cache_initialized)
+    return;
+
+  _mutex_lock(&crl_response_cache_mutex);
+  cleanup_memory_cache_locked();
+  _mutex_unlock(&crl_response_cache_mutex);
+
+  cleanup_disk_cache();
```
sfc-gh-makowalski, line 1013: Memory cleanup holds `crl_response_cache_mutex`, but disk cleanup runs after the unlock. `save_crl_to_cache` writes the PEM **while holding that same mutex**, so the sweeper can `unlink`/`DeleteFile` a file another thread is still writing.

Hold the mutex for `cleanup_disk_cache()` as well (or otherwise serialize all disk create/replace/delete on this cache).

patches/curl/lib/vtls/sf_crl.c
```diff
@@ -792,16 +845,210 @@ static int error_handler(int ok, X509_STORE_CTX *ctx)
...
+  int cmp;
+
+  fp = BIO_new_file(file_path, "r");
+  if (!fp)
+    return 1;
+
+  crl = PEM_read_bio_X509_CRL(fp, NULL, NULL, NULL);
+  BIO_free(fp);
+  if (!crl)
+    return 1;
```
sfc-gh-makowalski, line 902: Returning 1 here (and on `BIO_new_file` failure) means “delete this file.” Combined with walking every entry in the cache directory, that deletes anything that is not a parseable CRL — a truncated write, a permission error, or an unrelated file if `SF_CRL_RESPONSE_CACHE_DIR` points at a shared folder.

Treat open/parse failure as “keep” (and log), and only delete files this cache actually wrote. “Corrupt PEM → delete” can stay if you first restrict the walk to those filenames.

sfc-gh-makowalski (changes requested):
The cache-cleanup approach (background thread + join in `termCertCRL` before OpenSSL teardown) is the right shape, but a few issues should be fixed before merge.

Blocking:
1. Windows thread entry has the wrong calling convention / return type.
2. Disk cleanup is not serialized with disk writes — the sweeper can unlink a CRL another thread is still writing.
3. Any file that fails to parse as a CRL is deleted, including unrelated files if `SF_CRL_RESPONSE_CACHE_DIR` is a shared directory.

Please also drop the AWS / Azure / OOB build-number bumps unless those trees actually changed (curl `1 → 2` is the bump that belongs here).

sfc-gh-makowalski (approved):
The three requested changes are in place on this revision:

- Windows cleanup thread uses a DWORD WINAPI wrapper (crl_cleanup_thread_win).
- Disk cleanup runs under crl_response_cache_mutex, same as cache writes.
- Directory walk is limited to cache-style names (http___ / https___ / ldap___ / ldaps___); open failure keeps the file; unrelated-file test added.

Approving. Remaining nits (unrelated AWS/Azure/OOB version bumps, strncat PATH_MAX) are non-blocking.

